PR #57, opened 2015-04-25
Run system test using docker in docker

sachja:
All system and regression test can be run on single VM using docker in docker with this change. if the user sets an environment variable TESTBED=DIND before running the system/regression test suite, the testbed will be set using dockerhost scripts instead of vagrant. The default is still to use vagrant for now.

### DIFF
--- a/netd.go
+++ b/netd.go
@@ -17,9 +17,9 @@ package main
 
 import (
 	"flag"
-	"github.com/samalba/dockerclient"
 	"log"
 	"os"
+	"os/exec"
 	"strings"
 
 	"github.com/contiv/netplugin/core"
@@ -28,6 +28,7 @@ import (
 	"github.com/contiv/netplugin/crtclient/docker"
 	"github.com/contiv/netplugin/drivers"
 	"github.com/contiv/netplugin/plugin"
+	"github.com/samalba/dockerclient"
 )
 
 // a daemon based on etcd client's Watch interface to trigger plugin's
@@ -339,6 +340,25 @@ func handleContainerStart(netPlugin *plugin.NetPlugin, crt *crt.Crt,
 	return err
 }
 
+func handleContainerStop(netPlugin *plugin.NetPlugin, crt *crt.Crt,
+	contId string) error {
+	// If CONTIV_DIND_HOST_GOPATH env variable is set we can assume we are in docker in docker testbed
+	// Here we need to set the network namespace of the ports created by netplugin back to NS of the docker host
+	hostGoPath := os.Getenv("CONTIV_DIND_HOST_GOPATH")
+	if hostGoPath != "" {
+		osCmd := exec.Command("github.com/contiv/netplugin/scripts/dockerhost/setlocalns.sh")
+		osCmd.Dir = os.Getenv("GOSRC")
+		output, err := osCmd.Output()
+		if err != nil {
+			log.Printf("setlocalns failed. Error: %s Output: \n%s\n",
+				err, output)
+			return err
+		}
+		return err
+	}
+	return nil
+}
+
 func handleDockerEvents(event *dockerclient.Event, retErr chan error,
 	args ...interface{}) {
 	var err error
@@ -366,6 +386,10 @@ func handleDockerEvents(event *dockerclient.Event, retErr chan error,
 
 	case "die":
 		log.Printf("received die event for container \n")
+		err = handleContainerStop(netPlugin, crt, event.Id)
+		if err != nil {
+			log.Printf("error '%s' handling container %s \n", err, event.Id)
+		}
 		// decide if we should remove the container network policy or leave
 		// it until ep configuration is removed
 		// ep configuration as instantiated can be applied to another container
--- a/scripts/dockerhost/cleanup-dockerhosts
+++ b/scripts/dockerhost/cleanup-dockerhosts
@@ -1,4 +1,3 @@
-
 #!/bin/bash
 
 num_nodes=1
@@ -10,9 +9,9 @@ echo $num_nodes
 
 for i in `seq 1 $num_nodes`; 
 do
-    hostname="netplugin-node$i"
-    echo "Cleaning $hostname"
-    sudo docker exec $hostname service docker stop
-    sudo docker rm -f $hostname
+    host="netplugin-node$i"
+    echo "Cleaning $host"
+    sudo docker exec $host service docker stop
     sudo ip link delete $i-ext
+    sudo docker rm -f -v $host
 done
--- a/scripts/dockerhost/setlocalns.sh
+++ b/scripts/dockerhost/setlocalns.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+scriptdir=`dirname "$BASH_SOURCE"`
+HOST_SCRIPTDIR=$CONTIV_DIND_HOST_GOPATH/src/github.com/contiv/netplugin/scripts/dockerhost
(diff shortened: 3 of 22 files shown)


netd.go
```diff
@@ -17,17 +17,17 @@ package main
...
-	"strings"
-
 	"github.com/contiv/netplugin/core"
 	"github.com/contiv/netplugin/crt"
 	"github.com/contiv/netplugin/crtclient"
 	"github.com/contiv/netplugin/crtclient/docker"
 	"github.com/contiv/netplugin/drivers"
 	"github.com/contiv/netplugin/plugin"
+	"github.com/samalba/dockerclient"
+	"log"
```
mapuri: The usual convention I have seen is that the standar packages go at the top and then the other user packages with a empty line between the two. We should follow that here for consistency.

netd.go
```diff
@@ -339,6 +339,25 @@ func handleContainerStart(netPlugin *plugin.NetPlugin, crt *crt.Crt,
 	return err
 }
 
+func handleContainerStop(netPlugin *plugin.NetPlugin, crt *crt.Crt,
+	contId string) error {
+	// If HOST_GOPATH env variable is set we can assume we are in docker in docker testbed
+	// Here we need to set the network namespace of the ports created by netplugin back to NS of the docker host
+	hostGoPath := os.Getenv("HOST_GOPATH")
```
mapuri: May be we should introduce a more specific environment variable like CONTIV_DIND to avoid any future conflicting use of this variable elsewhere?

scripts/dockerhost/cleanup-dockerhosts
```diff
@@ -14,5 +13,4 @@ do
     echo "Cleaning $hostname"
     sudo docker exec $hostname service docker stop
     sudo docker rm -f $hostname
-    sudo ip link delete $i-ext
```
mapuri: is this not needed anymore? This corresponds to host container's veth pair, right?

scripts/dockerhost/setlocalns.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+scriptdir=`dirname "$BASH_SOURCE"`
+HOST_SCRIPTDIR=$HOST_GOPATH/src/github.com/contiv/netplugin/scripts/dockerhost
+pid=$($scriptdir/host-sh $HOST_SCRIPTDIR/docker-pid $(hostname))
+FILES=$($scriptdir/host-sh ls /sys/class/net/)
```
mapuri: is the path '/sys/class/net' os specific i.e. can it differ between say ubuntu and centos host oses? Just wondering if this puts an assumption on where the DIND environment is deployed. If so may be we should put it in a comment here.

systemtests/utils/utils.go
```diff
@@ -72,55 +72,60 @@ func startNetPlugin(t *testing.T, nodes []VagrantNode, nativeInteg bool) {
 
 	for i, node := range nodes {
 		//start the netplugin
-		cmdStr := fmt.Sprintf("sudo PATH=$PATH nohup netplugin %s -host-label host%d 0<&- &>/tmp/netplugin.log &", nativeIntegStr,
-			i+1)
-		output, err := node.RunCommandWithOutput(cmdStr)
```
mapuri: Having command output in case of errors helps, any reason to remove it?

systemtests/utils/vagrantcommand.go
```diff
@@ -45,6 +35,11 @@ func (c *VagrantCommand) getCmd(cmd string, args ...string) *exec.Cmd {
 	return osCmd
 }
 
+type VagrantCommand struct {
```
mapuri: looks like the structure definition got moved somewhere in between it's member functions, suggest to move it to top.

mapuri:
@sachja thanks for working on this. I think we should add system-test and regression targets for DinD environment as well since we might be maintaining the two infras for a bit and that will avoid possibility of breaking one that is not run often.

netd.go
sachja: gofmt did this change for me. Reverted back.

netd.go
sachja: done

scripts/dockerhost/cleanup-dockerhosts
sachja: It seems to be automatically deleted when docker container is removed. Calling this was returing an error as the link is not present.

scripts/dockerhost/setlocalns.sh
sachja: Seems to be there in all linux versions.

systemtests/utils/utils.go
sachja: For running node commands for dind Output() always seems to block forever. I spent a lot of time on trying different things but could not make it work other than converting that to Run(). We can try together tomorrow to see if you have any ideas.

systemtests/utils/vagrantcommand.go
sachja: done

sachja:
Handled your comments except use of Run() instead of Output() . 
Dont know where to update the jenkins job to run system test in dind environment .. Can you send me the details and also how to force rerun the jenkins job.

scripts/dockerhost/cleanup-dockerhosts
mapuri: I think devices don't get removed automatically on container delete. I think to remove veth pair you need to give primary device name. From `start-dockerhosts` looks like that is `$i-int ` , may be that is what is giving the error.

systemtests/utils/utils.go
mapuri: sure, let's see if we can figure something.

mapuri:
> Dont know where to update the jenkins job to run system test in dind environment .. Can you send me the details and also how to force rerun the jenkins job. 

I meant to add Makefile targets for dind based tests, something like below. This will automatically cause Jenkins to run the tests in DinD environment:

```
all: build unit-test system-test system-test-dind
...
...
system-test-dind: build
     TESTBED=DIND go test -v -run "sanity" github.com/contiv/netplugin/systemtests/singlehost
     TESTBED=DIND go test --timeout 20m -v -run "sanity" github.com/contiv/netplugin/systemtests/twohosts
regress-test-dind: build
     TESTBED=DIND go test -v -run "regress" github.com/contiv/netplugin/systemtests/singlehost
     TESTBED=DIND go test --timeout 60m -v -run "regress" github.com/contiv/netplugin/systemtests/twohosts

```

scripts/dockerhost/cleanup-dockerhosts
sachja: It is getting removed on container rm for me. Anyways moved the link delete to before docker rm

sachja:
Removed system-test-dind from make all for now. Can we look at making this run on jenkins in another change. Realized that we will also need to install nsenter, docker etc on the jenkins machine ot make this work besides the path issue.

mapuri:
@sachja 
Sorry for getting back late, have been busy with a few things at work. I would rather have this fixed as part of this PR as it adds am important piece to the test infra.

I have installed nsenter, docker and brigde-utils as described in `docs/Dockerhost.md` file on the jenkins host. You can try pushing the change back and trigger a sanity Or if there is a way for you to try this on a bare metal that might help as well.

The jenkins test just runs the scripts/CI.sh script which just triggers a `make`, so I doubt it has something to do with that script. I shall also try running these in my local environment once I get a chance.

sachja:
I fixed the issue with the path.

Another issue is that the script runs some commands as sudo. Would that
work for the jenkins user on this machine without asking for a password?

PR #65, opened 2015-05-10
Add prerequisite checks to bail out early in docker based demo setup

mapuri:
This shall save folks from running into unknown error conditions if they end up running the setup in a non tested/tried environment.

### DIFF
--- a/scripts/dockerhost/start-dockerhosts
+++ b/scripts/dockerhost/start-dockerhosts
@@ -1,5 +1,56 @@
 #!/bin/bash
 
+# check if all prequisites are installed before proceeding with setup.
+# Pre-requisites:
+# - Ubuntu 14+
+# - Docker 1.6+
+# - nsenter
+# - bridge utils
+# - ovs
+
+errors=""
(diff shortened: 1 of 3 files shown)


scripts/dockerhost/start-dockerhosts
```diff
@@ -1,5 +1,56 @@
 #!/bin/bash
 
+# check if all prequisites are installed before proceeding with setup.
+# Pre-requisites:
+# - Ubuntu 14+
+# - Docker 1.6+
+# - nsenter
+# - bridge utils
+# - ovs
+
```
jainvipin: We need to check for etcd (or some other state driver) as well.

mapuri: @jainvipin 
These pre-requisites are just for the host(server/vm) where docker based setup needs to be brought up. etcd is packaged and run as part of the docker image.

jainvipin: @mapuri - I see, this is not for the docker images themselves; clearly we don't need to run etcd in the host.

PR #66, opened 2015-05-10
scripts/dockerhost/start-dockerhosts: Various fixups and catchalls for missing binaries

erikh:
This adds several changes which make the script idempotent and communicates needs to the user more effectively.

The script in its current state does not validate many errors that could be caused by an improperly configured machine. Now, we tell the user what we need.

Additionally, the script would not stop in the face of errors, and would continue forward doing many things that will not work because of prior expectations. This has been (hopefully) corrected.

I did not verify that the tool indeed still works. I will be reviewing the code later today and can validate that then.

### DIFF
--- a/scripts/dockerhost/start-dockerhosts
+++ b/scripts/dockerhost/start-dockerhosts
@@ -51,12 +51,28 @@ if [ -n "${errors}" ]; then
     exit 1
 fi
 
+set -e
+
 scriptdir=`dirname "$BASH_SOURCE"`
 echo $scriptdir
 
+if ! sudo ip link | grep -q docker0
+then
+  echo "Docker bridge is dead; restart docker"
+  if [ "x$CONTIV_NOWARN" == "x" ]
+  then
+    echo "Please press enter to allow this. Set \$CONTIV_NOWARN to disable this warning."
+    read
+    echo "Restarting docker."
+  fi
+  sudo service docker restart 2>/dev/null || :
+fi
+
 # Create a linux bridge between containers
-sudo brctl addbr br-em1
-sudo ip link set br-em1 up
+if sudo brctl addbr br-em1 2>/dev/null
+then
+  sudo ip link set br-em1 up
+fi
 
 num_nodes=1
 if [ -n "$CONTIV_NODES" ];
@@ -66,21 +82,24 @@ fi
 echo "Num nodes = "$num_nodes
 
 # Create the docker image for hosts
-# sudo docker build -t ubuntu_netplugin $scriptdir
+sudo docker build -t ubuntu_netplugin $scriptdir
 
 # Pull ubuntu image for use within the docker hosts
-sudo docker pull ubuntu
+sudo docker pull ubuntu:latest
 
-mkdir /tmp/ubuntu_image
+mkdir -p /tmp/ubuntu_image
 
-sudo docker save -o /tmp/ubuntu_image/ubuntu-image.tar ubuntu
+if [ ! -f /tmp/ubuntu_image/ubuntu-image.tar ]
+then
+  sudo docker save -o /tmp/ubuntu_image/ubuntu-image.tar ubuntu:latest
+fi
 
 cluster=""
 first="true"
 for i in `seq 1 $num_nodes`; 
 do
     host="netplugin-node$i"
-    echo "Starting $host"
+    echo "Starting $host, removing any old ones if necessary"
     # godep modifies the host's GOPATH env variable, CONTIV_HOST_GOPATH
     # contains the unmodified path passed from the Makefile, use that
     # when it is defined.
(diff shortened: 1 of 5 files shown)


erikh:
Ha! I probably should have checked first, @mapuri already did this in #65. Maybe we can still use this anyway; leaving it open.

jainvipin:
@erikh - I went over both the changes, it seems we need to merge the changes but some checks/changes you added are needed on top of @mapuri's changes.

scripts/dockerhost/start-dockerhosts
```diff
@@ -1,11 +1,40 @@
...
 echo $scriptdir
 
+if ! which nsenter 2>&1 >/dev/null
+then
+  echo "Please install nsenter. It's a part of the util-linux package, which may not be current on your system."
+  exit 1
+fi
+
+if ! sudo ip link | grep -q docker0
+then
```
jainvipin: presence of docker0 may not be the best way to detect if docker daemon is running. perhaps we can use `service docker status` to confirm that. With netplugin some folks may disable docker0 bridge all together.

scripts/dockerhost/start-dockerhosts
```diff
@@ -1,11 +1,40 @@
...
+  then
+    echo "Please press enter to allow this. Set \$CONTIV_NOWARN to disable this warning."
+    read
+    echo "Restarting docker."
+  fi
+  sudo service docker restart 2>/dev/null || :
+  sudo systemctl restart docker 2>/dev/null || :
+fi
+
+if ! which brctl 2>/dev/null
```
mapuri: this check is done as part of the pre-req checks in #65 and can be removed

scripts/dockerhost/start-dockerhosts
mapuri: @jainvipin 
I think this condition arises today because of this script, if ovs is not installed on the host and the script is run then it ends up wiping the docker0 interface of the host. I think this check is trying to catch and help that condition. @erikh kindly confirm in case I missed something.

Just as a side note, one of the pre-req checks added in #65 checks for ovs being installed on the host and shall help prevent this condition from occurring as well.

mapuri:
@erikh thanks for working on this.

#65 adds checks for ensuring prerequisites for docker based setup are installed to error out gracefully. We should use the cleanup and traces you have added in this PR.

I have merged #65, can you kindly rebase these over that.

scripts/dockerhost/start-dockerhosts
erikh: That is correct; the docker0 bridge disappears after executing the make task. I don't know why though.

erikh:
I should have said something earlier, but should be done now.

erikh:
Rebased and fixed up a touch given the changes that have come in.

scripts/dockerhost/start-dockerhosts
```diff
@@ -80,25 +97,30 @@ first="true"
 for i in `seq 1 $num_nodes`; 
 do
     host="netplugin-node$i"
-    echo "Starting $host"
-    # godep modifies the host's GOPATH env variable, CONTIV_HOST_GOPATH
-    # contains the unmodified path passed from the Makefile, use that
-    # when it is defined.
-    if [ -n $CONTIV_HOST_GOPATH ]; then
```
mapuri: I think this change is needed to correctly mount the host's gopath.

The current sanity failure seems unrelated though.

erikh: Hmm. maybe that got pruned during the merge. sorry for the distraction! Will fix this weekend.

erikh:
I think this branch is ok now, seems to bootstrap appropriately. @mapuri can you PTAL?

scripts/dockerhost/start-dockerhosts
```diff
@@ -89,16 +106,30 @@ do
...
     fi
-    sudo docker run -d -i -t --name $host --privileged -e CONTIV_DIND_HOST_GOPATH=$gopath \
-        -e GOSRC=/gopath/src/ -v /var/lib/docker -v $gopath:/gopath -v /proc:/host/proc \
-        -v /tmp/ubuntu_image:/ubuntu_image sachja/ubuntu_netplugin bash \
-        -c "/gopath/src/github.com/contiv/netplugin/scripts/dockerhost/start-service.sh & bash"
+
+    sudo docker rm -f $host 2>/dev/null || :
+    sudo docker run -d -i -t --name $host \
+      --privileged \
+      -e CONTIV_DIND_HOST_GOPATH=$GOPATH \
```
mapuri: s/$GOPATH/$gopath

scripts/dockerhost/start-dockerhosts
```diff
@@ -89,16 +106,30 @@ do
...
+      -e GOSRC=/gopath/src/ -v /var/lib/docker \
+      -v $GOPATH:/gopath \
```
mapuri: s/$GOPATH/$gopath

erikh:
Hmm. Even after fixing `$GOPATH` in this makefile something is breaking. @mapuri sorry to ask, but can you take a peek at what's breaking?

mapuri:
sure Erik, let me fork your tree and try it out locally

mapuri:
@erikh 

I tried it locally and looks like the changes to start docker inside the host docker container might be causing the 'mountpoint for cgroup not found' error. I noticed that on starting the service using `service docker start` (i.e. the original way), the issue doesn't happens. 

I still need to dig further, but some differences I noticed are:
- with `service docker start`, docker is started as a process on the host-container while with the new change it is started as a child of the the bash process. I am not very sure but I think this might be making a difference wrt the mountpoints available to docker daemon.
- with `service docker start`, docker is started with an extra '-p' option but it didn't made a difference when I tried using that option in your diff

**With `service docker start`:**
```
# ps aux | grep docker
root       145  1.0  1.1 413392 11496 ?        Sl   22:25   0:04 /usr/bin/docker -d -p /var/run/docker.pid
```

**With the new change:**
```
# ps aux | grep docker
root        81  0.0  0.1  17956  1412 ?        Ss   22:36   0:00 /bin/bash -c docker -d -p /var/run/docker.pid -D 2>&1 >/tmp/docker.log
root        97  4.0  1.2 216784 12920 ?        Sl   22:36   0:04 docker -d -p /var/run/docker.pid -D
```

mapuri:
looking at `/etc/init.d/docker` (which is called as part of `service docker start`) it appears that it does more work before starting the docker daemon and one of those is correctly mounting cgroupfs :) , which can explain why the change to start docker daemon directly doesn't work.

```
case "$1" in
        start)
                fail_unless_root

                cgroupfs_mount

                touch "$DOCKER_LOGFILE"
                chgrp docker "$DOCKER_LOGFILE"

                ulimit -n 1048576
                if [ "$BASH" ]; then
                        ulimit -u 1048576
                else
                        ulimit -p 1048576
                fi

                log_begin_msg "Starting $DOCKER_DESC: $BASE"
                start-stop-daemon --start --background \
                        --no-close \
                        --exec "$DOCKER" \
                        --pidfile "$DOCKER_SSD_PIDFILE" \
                        --make-pidfile \
                        -- \
                                -d -p "$DOCKER_PIDFILE" \
                                $DOCKER_OPTS \
                                        >> "$DOCKER_LOGFILE" 2>&1
                log_end_msg $?
                ;;
```

Issue #48, opened 2015-04-09
feature: improve error reporting


jainvipin:
When errors are not aggregated, it doesn't give a clear trace of who triggered the error and eventually where it failed. Aggregation of stacked errors that gets displayed like a backtrace is very useful. An implementation that attempts to aggregate the errors can be found in kubernetes (https://github.com/GoogleCloudPlatform/kubernetes/blob/master/pkg/util/errors/errors.go)

gaurav-dalvi:
@erikh : Can we close this issue if this is fixed ? Just digging older issues !

erikh:
fwiw we did resolve this, it just never got integrated back into netplugin

PR #71, opened 2015-05-12
Enhance core.Error to store the file and line number info with error

mapuri:
Use core.Error interface throughout the code for consistency and to allow making use of the features provided by this interface now and in future.

Signed-off-by: Madhav Puri <madhav.puri@gmail.com>

### DIFF
--- a/core/error.go
+++ b/core/error.go
@@ -15,14 +15,28 @@ limitations under the License.
 
 package core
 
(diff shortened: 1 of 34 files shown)


mapuri:
Though improving errors will be a continuous effort, this shall address and give some direction to addressing #48

PR #72, opened 2015-05-20
Handle stateful restart of ovs network driver

### LINKED ISSUE #47: idempotency not handled right: error message spews out but things work ok
When applying one_host_two_nets.json I see following error on the netplugin, suggesting that add operation was done twice and thus ovs operation is failing on the duplicate row. Although things work okay, because this is an error on attempt to duplicate create, but it could be handled elegantly without incurring it being an error situation.
```
examples# netdcli -cfg one_host_multiple_nets.json 
```

netplugin output:
```
2015/04/09 03:15:02 Received event for key: /contiv/config/eps/host1-native-intf
2015/04/09 03:15:02 Endpoint operation create failed. Error: ovs operation failed. Error(s): [constraint violation(Transaction causes multiple rows in "Interface" table to have identical values ("eth1") for index on column "name".  First row, with UUID 60cb8422-931a-4364-9f08-35e44711ccf0, existed in the database before this transaction and was not modified by the transaction.  Second row, with UUID 0d682c79-8d69-4110-9531-b67411921afd, was inserted by this transaction.)]
```

mapuri:
- restore ovs driver's oper state on a stateful restart to preserve port-number information
- On getting endpoint create check if an oper state for the endpoint already exists. If it exists and doesn't match the received config then unprogram the stale info before proceeding with handling the create. The CreateEndpoint call becomes a noop if the configuration matches the oper state.
- This enables the stateful restart of the driver but following condition is not yet handled:
  - changing the host binding of the endpoint while the driver was down, which may leave stale state on the previous host.
- Added tests for this change.

Fixes #47 

Signed-off-by: Madhav Puri <madhav.puri@gmail.com>

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -58,6 +58,11 @@ type Plugin interface {
 	Endpoint
 }
 
+type InstanceInfo struct {
+	StateDriver StateDriver `json:"-"`
+	HostLabel   string      `json:"host-label"`
+}
+
 type Driver interface {
 	// A driver implements the programming logic
 }
@@ -65,7 +70,7 @@ type Driver interface {
 type NetworkDriver interface {
 	// A network driver implements the programming logic for network
 	Driver
-	Init(config *Config, stateDriver StateDriver) error
+	Init(config *Config, info *InstanceInfo) error
 	Deinit()
 	CreateNetwork(id string) error
 	DeleteNetwork(id string) error
@@ -74,7 +79,7 @@ type NetworkDriver interface {
 type EndpointDriver interface {
 	// An endpoint driver implements the programming logic for endpoints
 	Driver
-	Init(config *Config, stateDriver StateDriver) error
+	Init(config *Config, info *InstanceInfo) error
 	Deinit()
 	CreateEndpoint(id string) error
 	DeleteEndpoint(id string) error
--- a/drivers/ovsdriver.go
+++ b/drivers/ovsdriver.go
@@ -16,6 +16,7 @@ limitations under the License.
 package drivers
 
 import (
+	"encoding/json"
 	"fmt"
 	"log"
 	"reflect"
@@ -48,6 +49,9 @@ const (
 
 	GET_PORT_NAME = true
 	GET_INTF_NAME = false
+
+	OVS_OPER_PATH_PREFIX = OPER_PATH + "ovs-driver/"
+	OVS_OPER_PATH        = OVS_OPER_PATH_PREFIX + "%s"
 )
 
 type OvsDriverConfig struct {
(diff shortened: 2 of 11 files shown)


core/core.go
```diff
@@ -65,7 +69,7 @@ type Driver interface {
 type NetworkDriver interface {
 	// A network driver implements the programming logic for network
 	Driver
-	Init(config *Config, stateDriver StateDriver) error
+	Init(config *Config, info *InstanceInfo, stateDriver StateDriver) error
```
jainvipin: @mapuri - not sure if we should combine the Config and Instance info as init metadata structure; To avoid changing the core interfaces, if some other data needs to be passed during init, we may end up adding one more field to it.

drivers/ovsdriver.go
```diff
@@ -57,14 +61,38 @@ type OvsDriverConfig struct {
 	}
 }
 
+type OvsDriverOperState struct {
+	core.CommonState
+	// used to allocate port names. XXX: should it be user controlled?
+	CurrPortNum int `json:"currPortNum"`
```
jainvipin: moving CurrPortNum to Driver's global Oper State is ok. Is it more cosmetic? Because I am still not able to find out how an ep is given back the same port-number, etc. because of this change. EP's restore should come with various port numbers that were needed for those EPs from the ep-oper state.

core/core.go
mapuri: @jainvipin 

`Config` and `InstanceInfo` still are separate. InstanceInfo provides a way to pass plugin-instance specific info down to the drivers which are more runtime on nature (like host-label). 

I agree that we should define core API in a way that they change little when a new parameter shows up. Thinking more, I feel we can classify parameters passed to `Init()` into two categories viz. config and instance-specific. That way we can fold the StateDriver argument into `InstanceInfo` as well. Going forward the `InstanceInfo` structure can then provide a way to add mroe params (if any) that may need to be passed to drivers. WDYT? I can make the change if that sounds ok.

drivers/ovsdriver.go
mapuri: CurrPortNum is used for deriving the names of internal ports in ovs, so it needs to be restored to it's previous value in order to prevent drivers from reusing any existing ports. This is needed for driver to function correctly after a restart.

There is no explicit restore of Endpoint needed as part of `Init()` as the driver get's a replay of the configuration from the Plugin (as part of restore). Now on the replay, the checks in CreateEndpoint() ensure that the driver acts based on operational state, if any, it had created before which will ensure Driver continues to keep same port number etc for the endpoint.

core/core.go
jainvipin: @mapuri 

> That way we can fold the StateDriver argument into InstanceInfo as well. Going forward the InstanceInfo structure can then provide a way to add mroe params (if any) that may need to be passed to drivers. WDYT?

It would work too, because StateDriver is on specific InstanceInfo. On other note, should we call it `Config` and `Runtime` to keep it consistence with rest of the terminology.

mapuri: > should we call it Config and Runtime to keep it consistence with rest of the terminology

till now we have used `Runtime` for the oper state of something. Example, Config of a Driver, Runtime (or oper) of a Driver. But this info is more of the plugin-instance being passed to the Driver's Init(). So I think it is ok to call it InstanceInfo for now, until we have a need to pass non Instance specific things in future and then we may just call it `misc` :smile: .

mapuri:
ping @jainvipin 

I have incorporated the comment and rebased the change. PTAL

core/core.go
```diff
@@ -74,7 +79,7 @@ type NetworkDriver interface {
 type EndpointDriver interface {
 	// An endpoint driver implements the programming logic for endpoints
 	Driver
-	Init(config *Config, stateDriver StateDriver) error
+	Init(config *Config, info *InstanceInfo) error
```
jainvipin: yes this one looks good!

PR #73, opened 2015-05-21
Reorganize drivers

erikh:
This pushes most of the state drivers into `drivers/state`. There is also a small patch to ignore `.vagrant` directories that might live inside the repository. PTAL

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,6 @@ _testmain.go
 
 #ide files
 *.swp
+
+# vagrant
+.vagrant
--- a/gstate/gstate_test.go
+++ b/gstate/gstate_test.go
@@ -18,13 +18,13 @@ package gstate
 import (
 	"testing"
 
-	"github.com/contiv/netplugin/drivers"
 	"github.com/contiv/netplugin/resources"
+	"github.com/contiv/netplugin/state"
 )
 
 var (
 	gstateTestRA = &resources.EtcdResourceManager{Etcd: gstateSD}
-	gstateSD     = &drivers.FakeStateDriver{}
+	gstateSD     = &state.FakeStateDriver{}
 )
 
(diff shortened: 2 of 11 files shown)


jainvipin:
Technically these are state drivers. I am okay calling it state itself and not driver/state. @mapuri @shaleman - any opinions?

jainvipin:
@erikh - the changes looks good to me

mapuri:
I think organizing state-drivers under `drivers/state` is good.

One thing I am not very sure about is moving the Network and Endpoint config and oper states (viz OvsNetworkState and OvsEndpointState) there as well, since they are related to the NetworkDriver itself and might be better kept in the driver package itself? Example netmaster, gstate and resources also define certain states and it seems better to keep those states under their respective packages (close to the logic).

erikh:
Sure; I didn’t consider this when I made this patch yesterday.

I can stuff etcd/fakestate under `state` (if we want that) and put the ovs stuff back in `drivers`. I’m still thinking there’s a part of the picture where we’re still depending on OVS in an inflexible way and that’ll need to be revisited.

WDYT?

-Erik

mapuri:
> I can stuff etcd/fakestate under `state` (if we want that) and put the ovs stuff back in `drivers`.

SGTM

> I’m still thinking there’s a part of the picture where we’re still depending on OVS in an inflexible way and that’ll need to be revisited.

Yep, I agree ovs driver need to be revisited wrt programming and use of libovsdb and flows. I think @shaleman is exploring some of it as well. But my point in context of this PR was more general wrt keeping the state implementation close to the logic that depends on it.

erikh:
Fixed up to use `state` and puts the ovs stuff back in drivers. I squashed over the `.gitignore` commit; hope that's ok.

mapuri:
LGTM

PR #74, opened 2015-05-21
scripts/unittests: Quicker feedback for failed test builds.

erikh:
Somewhat Opinionated but... 

Noticed this while moving `drivers`. Now, any test failure, including build errors, will stop the VM immediately. I found that it was hard to quickly judge if the test had succeeded or not.

No need to merge it if it's not popular, I just thought this was a good idea.

### DIFF
--- a/scripts/unittests
+++ b/scripts/unittests
@@ -54,8 +54,12 @@ if ${run_in_vagrant}; then
     exit 0
 fi
 
-# running in the sand box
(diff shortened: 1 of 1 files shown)


jainvipin:
+1
I too find sometimes difficult to go up the screen and hunt for failures.

mapuri:
LGTM 

This makes it similar to how the singlehost and twohost tests fail as part `make systemtests`.

PR #75, opened 2015-05-21
go vet / go lint support

### LINKED ISSUE #68: integrate with tools like golint and govet to help improve and maintain overall code style
We can start with warning and as we cleanup code we can make the checks more strict

erikh:
This adds go vet and go lint support with downloading as necessary. I commented the sticky portions.

closes #68

### DIFF
--- a/scripts/checks
+++ b/scripts/checks
@@ -11,20 +11,51 @@ fi
 
 BUILD_PKGS=$1
 
-echo "Checking Go version..."
+echo "+++ Checking Go version..."
 ver=$(go version | awk '{print $3}')
 minVer="go1.4"
 if [[ ${ver} < ${minVer} ]]; then
-    echo "Go version check failed. Expected >=${minVer} but found ${ver}"
+    echo "!!! Go version check failed. Expected >=${minVer} but found ${ver}"
     exit 1
 fi
 
-echo "Checking gofmt..."
+echo "+++ Checking gofmt..."
 fmtRes=$(gofmt -l $BUILD_PKGS)
 if [ -n "${fmtRes}" ]; then
-    echo -e "gofmt checking failed:\n${fmtRes}"
+    echo -e "!!! gofmt checking failed:\n${fmtRes}"
     exit 1
 fi
 
-echo "All checks passed!!"
+if ! go get github.com/golang/lint/golint
+then
+  echo "!!! Could not install golint"
+  exit 1
+fi
+
+echo "+++ golint tree..."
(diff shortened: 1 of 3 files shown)


scripts/checks
```diff
@@ -11,20 +11,51 @@ fi
...
+echo "+++ Checking gofmt..."
 fmtRes=$(gofmt -l $BUILD_PKGS)
 if [ -n "${fmtRes}" ]; then
-    echo -e "gofmt checking failed:\n${fmtRes}"
+    echo -e "!!! gofmt checking failed:\n${fmtRes}"
     exit 1
 fi
 
-echo "All checks passed!!"
+if ! go get github.com/golang/lint/golint
```
mapuri: We can possibly move downloading golint to the 'deps' target in `Makefile`

scripts/checks
```diff
@@ -11,20 +11,51 @@ fi
...
+then
+  echo "!!! Could not install golint"
+  exit 1
+fi
+
+echo "+++ golint tree..."
+for i in $(find . -maxdepth 1 -type d -name '*' | grep -vE '/\..*$')
```
mapuri: The `checks` script takes `BUILD_PKGS` as argument that contains the list of our modules. We can alternatively iterate over those and then perhaps not need the `find`. It will keep this check consistent with rest of the checks as well.

```
lintRes=$(for dir in $BUILD_PKGS; do golint $dir; done)
if [ -n "${lintRes}" ]; then
    echo -e "!!! golint has suggestions:\n${lintRes}"
fi
```

scripts/checks
```diff
@@ -11,20 +11,51 @@ fi
...
+  then
+    echo "!!! Could not install go vet"
+    exit 1
+  fi
+fi
+
+echo "+++ go vet tree..."
+for i in ${BUILD_PKGS}
+do
+  go vet "${i}"
```
mapuri: shall we fail the script if 'go vet' fails as it catches code errors?

erikh: Sure; I thought the first rev was intentionally not breaking (so we can fix things), but I can do that in a few here.

mapuri: ah, ok...yeah that shall be fine, no need to fix those errors as part this PR.

scripts/checks
erikh: This is done; I moved everything to a `scripts/deps` since it was easier to.

-Erik

mapuri:
LGTM

PR #76, opened 2015-05-21
Move to logrus

erikh:
This just changes the package import to logrus and leaves the rest of the code alone, minus the removal of SetFlags calls (which are not relevant nor available in logrus).

We can exploit the new features in another patch.


crtclient/docker/docker.go
```diff
@@ -28,6 +27,8 @@ import (
 	"github.com/contiv/netplugin/crtclient"
 	"github.com/samalba/dockerclient"
 	"github.com/vishvananda/netlink"
+
+	log "github.com/Sirupsen/logrus"
```
jainvipin: @erikh - given that logrus is compatible with standard logger, a simple replacement works. Should we use other features from it too (log with fields, hooks, etc.)? Perhaps we'd become incompatible quickly with other things then.

Another question: I checked the package is licensed under as "MIT License", which will work for us w.r.t. royalties etc for using this software. Do you know any other significant open-source effort(s) using this?

erikh: Docker

jainvipin: LGTM

erikh: I should have clarified: docker uses it, so we should be comfortable both with its stability and its compatibility. I was hoping we could use additional features; f.e. json logging would fit well with tools like fluentd and logstash.

jainvipin: @erikh - makes sense

mgmtfn/pslibnet/main.go
```diff
@@ -67,7 +68,6 @@ var gcliOpts cliOpts
 func main() {
 	var flagSet *flag.FlagSet
 
-	log.SetFlags(log.LstdFlags | log.Lshortfile)
```
mapuri: I found `setFlag()` useful for easily modifying the log format (Example, having file name and line number comes in handy for debugging some failure without having to grep through code). 

I think logrus doesn't add this info by default and we could potentially define a custom formatter, if we want that info in logs later?

erikh: I can look into how that’s implemented in logrus (if at all), but I am preparing a log levels patch right now.

Is it possible to merge this and revisit early next week?

mapuri: yeah SGTM

PR #77, opened 2015-05-21
Fix go vet reported bad code

erikh:
There wasn't much. :)

Note that I removed the passing of `resErr` (chan error) to a few functions since it was only a component of the unreachable code found, making passing the arg superfluous.

### DIFF
--- a/netplugin/netd.go
+++ b/netplugin/netd.go
@@ -337,7 +337,7 @@ func handleContainerStart(netPlugin *plugin.NetPlugin, crt *crt.Crt,
 
 	err = attachContainer(netPlugin.StateDriver, crt, contName)
 	if err != nil {
-		log.Printf("error attaching container err \n", err)
+		log.Printf("error attaching container: %v\n", err)
 	}
 
 	return err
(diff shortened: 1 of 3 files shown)


netplugin/netd.go
```diff
@@ -337,7 +337,7 @@ func handleContainerStart(netPlugin *plugin.NetPlugin, crt *crt.Crt,
 
 	err = attachContainer(netPlugin.StateDriver, crt, contName)
 	if err != nil {
-		log.Printf("error attaching container err \n", err)
+		log.Print("error attaching container err \n", err)
```
jainvipin: Minor: Should change the above to print the error i.e. move the string to `"error %s attaching..."` and continue using Printf here. Most of the errors are printed that way.

erikh: Oh lol. Sorry, was moving too fast.

erikh:
Should be resolved now.

jainvipin:
LGTM

mapuri:
LGTM

Issue #50, opened 2015-04-09
feature: improve logging


jainvipin:
log messages do not need to be all on stdout:
- user option is to be provided to keep the logs in specific directory (logs directory) and a log file name picked in the location
- log messages to have ability to specify the severity fatal/critical, error, warning, info, verbose, extra-verbose
- standard formatting of the messages, per golang's log utilities

erikh:
https://github.com/Sirupsen/logrus might be a good thing here; already in use in docker so it'll be familiar territory for many.

I'm going to tackle this tomorrow, do you have any additional concerns?

erikh:
Sorry, I should have stated *why* logrus is a good idea. :)

I can use the `log` package but it's kind of fiddly, especially when it comes to syslog support (if we wanted to do that). glog can log to a socket but that's unfamiliar for many. logrus can do [numerous outputs](https://github.com/Sirupsen/logrus#hooks) which might prove useful to expose to our consumers.

jainvipin:
@erikh - I am not familiar with other packages. Let's go with your best pick.

gaurav-dalvi:
@jainvipin : Can we close this issue now ?

jainvipin:
sure - with multi-level logging I think we have basics!

PR #79, opened 2015-05-22
Implement multi-level logging

erikh:
This uses logrus's log levels features to implement multi-level logging.

Note that you cannot currently pick what levels to log; this'll be provided by another patch.

I tried to follow this convention:

* `Debugf` for all diagnostic messages
* `Errorf` for all errors or error messages
* `Fatalf` for all errors that cause a program to exit (I noticed there were a few of these that didn't exit; I added those exit calls, might be an issue)
* `Infof` for everything else

PTAL. Thanks!


erikh:
hmm. this breaks the system tests because of some code that greps the output of netdcli (which has now changed because of these logging changes). Will look into it sunday.

erikh:
So this is still broken due to #80 that I haven't investigated yet, but now has logging options to better exploit logrus's functionality. It can now log to syslog and output in json format if the user desires.

netdcli/netdcli.go
```diff
@@ -295,12 +295,13 @@ func validateOpts(opts *cliOpts) error {
...
 		if opts.oper.Get() == CLI_OPER_CREATE &&
 			opts.construct.Get() == CLI_CONSTRUCT_NW {
-			log.Printf("  auto allocating network subnet from global pool")
+			log.Infof("  auto allocating network subnet from global pool")
 		}
 	} else if opts.pktTag != "" {
 		_, err = strconv.Atoi(opts.pktTag)
 		if err != nil {
 			log.Fatalf("Error convertinng tag %s to integer \n", opts.pktTag)
+			os.Exit(1)
```
mapuri: We may not need explicit os.Exit(1) after call to log.Fatalf() as I think Fatalf() will call exit.

netdcli/netdcli.go
```diff
@@ -313,6 +314,7 @@ func validateOpts(opts *cliOpts) error {
 		opts.homingHost, err = os.Hostname()
 		if err != nil {
 			log.Fatalf("error obtaining the hostname, error %s \n", err)
+			os.Exit(1)
```
mapuri: same as above and rest of the places.

netdcli/netdcli.go
```diff
@@ -367,7 +373,7 @@ func executeOpts(opts *cliOpts) error {
 		return err
 	}
 
-	// log.Printf("parsed all valuees = %v \n", opts)
+	log.Debugf("parsed all valuees = %v \n", opts)
```
mapuri: s/valuees/values

netplugin/netd.go
```diff
@@ -33,6 +35,7 @@ import (
 	"github.com/samalba/dockerclient"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/Sirupsen/logrus/hooks/syslog"
```
mapuri: need to update godep for this.

systemtests/singlehost/setup_test.go
```diff
@@ -37,7 +37,7 @@ func TestMain(m *testing.M) {
 	err := testbed.Setup(os.Getenv("CONTIV_ENV"), 1)
 	log.Printf("Done with testbed setup...")
 	if err != nil {
-		log.Printf("Testbed setup failed. Error: %s", err)
+		log.Fatalf("Testbed setup failed. Error: %s", err)
```
mapuri: need to move the Fatal log after the call to Teardown() to ensure cleanup

systemtests/twohosts/setup_test.go
```diff
@@ -36,7 +36,7 @@ func TestMain(m *testing.M) {
 	err := testbed.Setup(os.Getenv("CONTIV_ENV"), 2)
 	log.Printf("Done with testbed setup...")
 	if err != nil {
-		log.Printf("Testbed setup failed. Error: %s", err)
+		log.Fatalf("Testbed setup failed. Error: %s", err)
```
mapuri: need to move the Fatal log after the call to Teardown() to ensure cleanup

netdcli/netdcli.go
erikh: It does in `log` but not in logrus. This is one of the ways it differs.

I will look at encapsulating this in a better way.

netplugin/netd.go
erikh: This should already be included in the logrus clone, is it necessary to import it explicitly?

erikh:
Hi,

I've added a new `utils` package with `LogExit` which behaves like the old `log.Fatal`. I didn't think there was a better place for it, but I might have overlooked something.

Patched this into the sections where log.Fatal would be expected to exit.

erikh:
Note that the tests will still fail until #83 is merged.

netplugin/netd.go
mapuri: ah, ok..sorry I missed that :disappointed:

mapuri:
@erikh 
ok, LGTM.

from code it appears that logrus also exits on call to Fatalf() [see this] (https://github.com/Sirupsen/logrus/blob/master/entry.go#L143), but looks there is a (misplaced?) loglevel check that may prevent it sometimes. So we may still need to move the Fatalf() call after Teardown() in setup_test.go

erikh:
@mapuri @jainvipin I think this is ready for review; I had to change a `Debugf` to an `Infof` in the first commit to ensure I didn't break IP address retrieval again (see #84 for a potential long-term solution). I also added flags to ensure debug logging would be echoed if requested, but it's kind of half done. I'd like to have selectable levels, but I don't think that needs to happen in this iteration.

PTAL

erikh:
Just for note taking sanity this partially resolves #50

utils/logutils.go
```diff
@@ -0,0 +1,12 @@
+package utils
+
+import (
+	"os"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+func LogExit(msg string, fmts ...interface{}) {
+	log.Fatalf(msg, fmts...)
```
mapuri: @erikh, I am still wondering if we need this utility function. Unless user set's log level to be `PanicLevel`, logrus will exit.

also just fyi I created a PR https://github.com/Sirupsen/logrus/pull/188 in logrus to fix the above behavior, once it is merged we can update our godeps to get the fix.

netplugin/netd.go
```diff
@@ -485,24 +494,68 @@ func handleEvents(netPlugin *plugin.NetPlugin, crt *crt.Crt, opts cliOpts) error
...
+	// formatter, because the tty detection is useless here.
+	if tf, ok := log.StandardLogger().Formatter.(*log.TextFormatter); ok {
+		tf.DisableColors = true
+	}
+
+	if syslogParam == "kernel" {
+		hook, err = logrus_syslog.NewSyslogHook("", "", syslog.LOG_INFO, "netplugin")
+		if err != nil {
+			log.Fatalf("Could not connect to kernel syslog")
+			os.Exit(1)
```
mapuri: can remove the Exit() call in this function.

utils/logutils.go
erikh: Ah ok; I didn’t realize you sent a patch upstream. Let’s hold off on this patch until that gets merged, then.

-Erik

Sirupsen: Both merged!

Sirupsen: Released as 0.8.2

erikh: Thanks Simon! @mapuri I’ll start fixing this up tonight, spare yourself some effort for now, we can discuss it tomorrow if necessary.

-Erik

mapuri: awesome thanks @Sirupsen !

crtclient/docker/docker.go
```diff
@@ -75,7 +75,7 @@ func (d *Docker) getContPid(ctx *crtclient.ContainerEpContext) (string, error) {
 	contInfo, err := d.Client.InspectContainer(contNameOrId)
 	if err != nil {
 
-		log.Printf("unable to get container info for '%s' \n",
+		log.Errorf("unable to get container info for '%s' \n",
```
jainvipin: Just pinging folks - is it a good idea to come up with a standard format of an error and warning messages? It wouldn't be part of this checkin regardless.

crtclient/docker/docker.go
```diff
@@ -284,11 +281,11 @@ func (d *Docker) configureIfAddress(ctx *crtclient.ContainerEpContext) error {
 	out, err = exec.Command("/sbin/ip", "netns", "exec", contPid, "ip",
 		"link", "set", ctx.InterfaceId, "up").Output()
 	if err != nil {
-		log.Printf("error bringing interface %s up 'out = %s', err = %s\n",
+		log.Errorf("error bringing interface %s up 'out = %s', err = %s\n",
 			ctx.InterfaceId, out, err)
 		return err
 	}
-	log.Printf("successfully configured ip and brought up the interface \n")
+	log.Errorf("successfully configured ip and brought up the interface \n")
```
jainvipin: This message should be Info, not an error.

mgmtfn/pslibnet/main.go
```diff
@@ -87,22 +87,23 @@ func main() {
 	err = flagSet.Parse(os.Args[1:])
 	if err != nil {
 		log.Fatalf("Failed to parse command. Error: %s", err)
+		os.Exit(1)
```
jainvipin: minor: should we use different exit codes on different errors?

netplugin/netd.go
jainvipin: I have the same question, doing explicit exit after Fatalf would seem like Fatalf may not panic/exit.

resources/vlanresource_test.go
```diff
@@ -39,25 +39,25 @@ type vlanRsrcValidator struct {
 func (vt *vlanRsrcValidator) nextCfgState() {
 	vt.expCfg = vt.expCfg[1:]
 	if len(vt.expCfg) > 0 {
-		log.Printf("after pop cfg is: %+v\n", vt.expCfg[0])
+		log.Debugf("after pop cfg is: %+v\n", vt.expCfg[0])
```
jainvipin: @mapuri - are these excessive debugs? Perhaps the dev level debug should be decprecated after some code solidification and add others that might give events and gives a flow of an operation

jainvipin:
@erikh - except for some comments, the changes LGTM

crtclient/docker/docker.go
erikh: +1

mapuri: yeah, that's a good idea. logrus provides a WithFields() logging method that can give structure to our log messages as a starter, https://github.com/Sirupsen/logrus#fields

mgmtfn/pslibnet/main.go
mapuri: Just found this: http://www.tldp.org/LDP/abs/html/exitcodes.html

Most of our errors might just fall in exit code 1. We can explore adding our own codes, do you see the exit code being interpreted by the parent scripts/processes?

resources/vlanresource_test.go
mapuri: yeah, we can remove these if they become excessive. I had added them to debug the tests in these files but left them there for future tests.

erikh:
What would we like to do about the exit codes and debug messages in this patch?

erikh:
rebased

mapuri:
@erikh 
I don't see comments around using logrus.Fatalf() and removing os.Exit() not taken care of. Were you planning to take care of those separately?

> What would we like to do about the exit codes and debug messages in this patch?

I am ok taking care of these separately. ping @jainvipin

erikh:
gah, sorry! Forgot. will resolve tomorrow.

crtclient/docker/docker.go
erikh: I'll take a stab at this today.

erikh:
@mapuri @jainvipin PTAL, I've made an attempt to clean up this patch. I hope to have one around error logging and stack traces later today.

mapuri:
LGTM

@erikh  I think we need to update the godeps for logrus to pull in Fatalf() related fixes, otherwise this is good to merge.

jainvipin:
LGTM!

erikh:
Logrus is updated, waiting for tests to pass now.

PR #82, opened 2015-05-26
Fix many golint errors

### LINKED ISSUE #81: fix 'go lint' reported isssues
When doing make I see following spewing out...

```
+++ golint tree...
state/etcdstatedriver.go:31:2: exported const RECURSIVE should have comment (or a comment on this block) or be unexported
state/etcdstatedriver.go:34:6: exported type EtcdStateDriverConfig should have comment or be unexported
state/etcdstatedriver.go:40:6: exported type EtcdStateDriver should have comment or be unexported
state/etcdstatedriver.go:44:1: exported method EtcdStateDriver.Init should have comment or be unexported
state/etcdstatedriver.go:60:1: exported method EtcdStateDriver.Deinit should have comment or be unexported
. . .
```
Opening the issue to track fixing them.

erikh:
Partially resolves #81. I will be building another PR to finish it off tomorrow.

### DIFF
--- a/crtclient/crtclient.go
+++ b/crtclient/crtclient.go
@@ -15,29 +15,31 @@ limitations under the License.
 
 package crtclient
 
-// container runtime client interface implementations
-
-type Config struct {
-	// Config object parsed from a json styled config
-	V interface{}
-}
-
-type ContainerEpContext struct {
+// ContainerEPContext is used to manage the parameters for all
+// container+endpoint management operations.
+type ContainerEPContext struct {
 	NewContName    string
 	NewAttachUUID  string
 	CurrContName   string
 	CurrAttachUUID string
-	InterfaceId    string
-	IpAddress      string
+	InterfaceID    string
+	IPAddress      string
 	SubnetLen      uint
 	DefaultGw      string
 }
 
+// Config should be replaced by core.Config FIXME
+type Config struct {
+	V interface{}
+}
+
+// ContainerIf implementations are used to configure and manage container
+// interfaces.
 type ContainerIf interface {
 	Init(config *Config) error
 	Deinit()
-	AttachEndpoint(ctx *ContainerEpContext) error
-	DetachEndpoint(ctx *ContainerEpContext) error
-	GetContainerId(contName string) string
+	AttachEndpoint(ctx *ContainerEPContext) error
+	DetachEndpoint(ctx *ContainerEPContext) error
+	GetContainerID(contName string) string
 	GetContainerName(contName string) (string, error)
 }
--- a/drivers/ovsconstants.go
+++ b/drivers/ovsconstants.go
@@ -1,0 +1,43 @@
+package drivers
+
+const (
+	operCreateBridge oper = iota
+	operDeleteBridge
+	operCreatePort
+	operDeletePort
+)
+
+const (
+	// StateBasePath is the base path for all state operations.
+	StateBasePath = "/contiv/"
+	// StateConfigPath is the path to the root of the configuration state
+	StateConfigPath = StateBasePath + "config/"
+	// StateOperPath is the path to the operations stored in state.
+	StateOperPath = StateBasePath + "oper/"
+
+	networkConfigPathPrefix  = StateConfigPath + "nets/"
+	networkConfigPath        = networkConfigPathPrefix + "%s"
+	endpointConfigPathPrefix = StateConfigPath + "eps/"
+	endpointConfigPath       = endpointConfigPathPrefix + "%s"
+	networkOperPathPrefix    = StateOperPath + "nets/"
+	networkOperPath          = networkOperPathPrefix + "%s"
+	endpointOperPathPrefix   = StateOperPath + "eps/"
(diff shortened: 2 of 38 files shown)


crtclient/crtclient.go
```diff
@@ -15,29 +15,26 @@ limitations under the License.
...
+	IPAddress      string
 	SubnetLen      uint
 	DefaultGw      string
 }
 
+// ContainerIf implementations are used to configure and manage container
+// interfaces.
 type ContainerIf interface {
-	Init(config *Config) error
+	Init(config interface{}) error
```
mapuri: Just curious, was this reported by go-lint? `Config` was defined as a struct just so as to make it more explicit that it contains an interface consumable by json marshal/unmarshal routines. Albeit we could have used `core.Config` here itself than defining a new struct.

drivers/ovsconstants.go
```diff
@@ -0,0 +1,43 @@
+package drivers
+
+const (
+	operCreateBridge oper = iota
+	operDeleteBridge
+	operCreatePort
+	operDeletePort
+)
+
+const (
```
mapuri: May be we should merge all these consts into one block? I am not sure if there is any golang specific convention here though.

drivers/ovsdriver.go
```diff
@@ -28,62 +28,44 @@ import (
...
-	DELETE_PORT
-
-	GET_PORT_NAME = true
-	GET_INTF_NAME = false
-
-	OVS_OPER_PATH_PREFIX = OPER_PATH + "ovs-driver/"
-	OVS_OPER_PATH        = OVS_OPER_PATH_PREFIX + "%s"
-)
-
+// OvsDriverConfig implements the NetworkDriver and EndpointDriver interface
```
mapuri: This shall be:

```
OvsDriverConfig defines the configuration required to initialize the OvsDriver.
```

drivers/ovsdriver.go
```diff
@@ -28,62 +28,44 @@ import (
...
+// for an vlan based openvSwitch deployment
 type OvsDriverConfig struct {
 	Ovs struct {
-		DbIp   string
+		DbIP   string
 		DbPort int
 	}
 }
 
+// OvsDriverOperState carries state about the operations.
```
mapuri:
```
OvsDriverOperState carries operational state of the OvsDriver.
```

netmaster/netmaster.go
```diff
@@ -971,7 +982,8 @@ func validateEpBindings(epBindings *[]ConfigEp) error {
 	return nil
 }
 
-func CreateEpBindings(stateDriver core.StateDriver, epBindings *[]ConfigEp) error {
+// CreateEpBindings creates bridges between host and containers.
```
mapuri:
```
CreateEpBindings binds an endpoint to a host by updating host-label info in driver's endpoint configuration.
```

mapuri:
@erikh thanks for taking care of this. The changes LGTM except for a few minor comments. Also the sanity seems to be failing may be there is some problem with the trunk itself, I will try running it locally and see if I can find anything as well.

crtclient/crtclient.go
erikh: This was not, I should have made note of it or just flat-out not included it in this patch, sorry. I will probably remove it, but let’s sort this out.

I presume you want to ensure it’s a JSON object? Is there some reason you cannot just `type Config interface{}`?

drivers/ovsconstants.go
erikh: I like the iota’s to be in a separate block so it’s obvious they all belong to the same group, otherwise I have no preferences.

-Erik

erikh:
This is probably #83, I will rebase this after merge + fixing the interface issue, then we can revisit this patch.

-Erik

crtclient/crtclient.go
mapuri: Yep, you are right. `type Config interface{}` shall also work.

Initially when I had defined `core.Config` struct my understanding was that it will be more explicit to pass a struct containing json config as argument to the Init() routines of `drivers`. But thinking more having a type-def like `type Config interface{}` achieves the same by keeping the type of the config argument more explicit. Still I think we should use the same `core.Config` type here as well than defining a new type, may be we can make the change as part of separate PR or whatever works for you.

drivers/ovsconstants.go
mapuri: yeah sounds good, let's keep two blocks then.

Also `ovsOperPathPrefix` and `ovsOperPath` can be moved right after 'endpointOperPath', just to keep similar kind of consts together

crtclient/crtclient.go
erikh: I concur. I will address this in a second patch.

-Erik

erikh: I've reverted this in this patch and intend to introduce it more globally in a future one.

erikh:
This should be ready for another pass after the tests pass.

mapuri:
LGTM, can merge once the test pass!

PR #89, opened 2015-05-29
Final pass of linting errors

erikh:
This also makes lint passing mandatory. I've also globally capitalized JSON as it prescribes.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -13,7 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// The package 'core' provides definition for a generic interface that helps
+// Package core provides definition for a generic interface that helps
 // provision networking for an endpoint (like a container,
 // a vm or a bare-metal host). The interface is invoked (north-bound) by the
 // 'daemon' or the extension-plugin (TBD) part of docker. The interface in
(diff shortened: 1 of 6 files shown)


core/core.go
```diff
@@ -57,6 +57,7 @@ type Plugin interface {
 	Endpoint
 }
 
+// InstanceInfo encapsulates data about the host.
```
mapuri:
```
InstanceInfo encapsulates data that is specific to a running instance of netplugin like label of host on which it is started. 
```

core/state.go
```diff
@@ -1,5 +1,6 @@
 package core
 
+// State interface encapsulates all state within the netplugin system.
```
mapuri:
```
State identifies data uniquely identifiable by 'id' and stored in a (distributed) key-value store implemented by core.StateDriver. 
```

scripts/checks
```diff
@@ -29,7 +29,13 @@ fi
 echo "+++ golint tree..."
 for i in $(find . -maxdepth 1 -type d -name '*' | grep -vE '/\..*$')
 do
-  golint "${i}"
+  output=$(golint "${i}") # golint has no exit codes
+  if [ -n "${output}" ]
+  then
+    echo -e "!!! golint checking failed: "
+    echo -n "${output}"
+    exit 1
```
mapuri: Just a minor nit. This will cause the script to error out on first module it hits errors, would it instead be better to collect all go-lint errors and print them together?

erikh: Yeah, that is a better idea. I will retool in a bit.

erikh:
This is ready for merge, I think I covered all of your comments.

mapuri:
LGTM, thanks Erik!

PR #90, opened 2015-06-02
Allow docker to be injected from the GOPATH if it exists.

erikh:
Small patch to import docker from $GOPATH if it is placed there, the idea being that we can run bleeding edge versions of docker this way. Quick n' Dirty. Let me know if you would like this done another way.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 
 .PHONY: all build clean default system-test unit-test
 
-TO_BUILD := ./netplugin/ ./netdcli/ ./mgmtfn/k8contivnet/ ./mgmtfn/pslibnet/
+TO_BUILD := ./netplugin/ ./netdcli/ ./mgmtfn/k8contivnet/ ./mgmtfn/pslibnet/ ./mgmtfn/docknet
 HOST_GOBIN := `which go | xargs dirname`
 HOST_GOROOT := `go env GOROOT`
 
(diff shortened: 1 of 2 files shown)


Vagrantfile
```diff
@@ -132,6 +132,13 @@ provision_node = <<SCRIPT
  -listen-peer-urls http://#{node_addr}:2380 \
  -initial-cluster #{node_peers} \
  -initial-cluster-state new 0<&- &>/tmp/etcd.log &) || exit 1
+
+if [ -f "#{netplugin_synced_gopath}/bin/docker" ]
```
mapuri: Will it be better if we pass the path to docker binary as a separate environment variable say CONTIV_DOCKER_BIN? 

It will give more flexibility to try any docker binary from the host and also help in cases where docker binary might not be there in gopath, for instance for me somehow the docker binary just gets generated in the `bundles/<docker-release>` directory in my docker workspace and doesn't get's placed in my `$GOPATH/bin` directory.

Also slightly related, there may be a use case to have this facility for other relevant binaries as well (like rkt, k8s etc), not sure if there is an easy generic way to address all. But I am ok to take care of that separately.

Makefile
```diff
@@ -1,7 +1,7 @@
 
 .PHONY: all build clean default system-test unit-test
 
-TO_BUILD := ./netplugin/ ./netdcli/ ./mgmtfn/k8contivnet/ ./mgmtfn/pslibnet/
+TO_BUILD := ./netplugin/ ./netdcli/ ./mgmtfn/k8contivnet/ ./mgmtfn/pslibnet/ ./mgmtfn/docknet
```
mapuri: This shall not go in this patch I believe without associated code.

erikh:
yeah this is not a good idea. I will close this ticket.

PR #91, opened 2015-06-06
Implement the StateDriver for a Consul datastore

mapuri:
This PR brings in support for consul based datastore to netplugin. 

I have added unittests for the consul state-driver. I shall complete the integration with netmaster and system-tests as a subsequent PR.

### DIFF
--- a/state/etcdstatedriver.go
+++ b/state/etcdstatedriver.go
@@ -170,10 +170,10 @@ func (d *EtcdStateDriver) ReadState(key string, value core.State,
 	return nil
 }
 
-// ReadAllStateCommon reads and unmarshals (given a function) all state into a
+// readAllStateCommon reads and unmarshals (given a function) all state into a
 // list of core.State objects.
 // XXX: move this to some common file
-func ReadAllStateCommon(d core.StateDriver, baseKey string, sType core.State,
+func readAllStateCommon(d core.StateDriver, baseKey string, sType core.State,
 	unmarshal func([]byte, interface{}) error) ([]core.State, error) {
 	stateType := reflect.TypeOf(sType)
 	sliceType := reflect.SliceOf(stateType)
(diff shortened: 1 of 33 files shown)


state/etcdstatedriver.go
```diff
@@ -211,7 +215,11 @@ func (d *EtcdStateDriver) ReadAllState(baseKey string, sType core.State,
 	return ReadAllStateCommon(d, baseKey, sType, unmarshal)
 }
 
-func (d *EtcdStateDriver) channelStateEvents(sType core.State,
+// ChannelStateEvents watches for updates(created, modify, delete) to a state of
+// specified type and unmarshals (given a function) all changes and puts then on
+// channel of core.WatchState objects.
+// XXX: move this to some common file
+func ChannelStateEvents(d core.StateDriver, sType core.State,
```
jainvipin: Is this a new public function exposed for the state driver now, which was embedded into WatchAll earlier, but I saw your comment about why Watchall can miss some events due to timing. Please clarify.

jainvipin:
overall LGTM; I am curious to know if we need to run both state drivers simultaneously and use one for state-sync and other for different reasons, like ddns services with consul, and etcd. for state distribution. How do we manage specifying them and hopefully not run into conflicts.

state/etcdstatedriver.go
mapuri: > Is this a new public function exposed for the state driver now, which was embedded into WatchAll earlier

No, this is an existing function. It is now shared by etcd and consul state-driver. Actually, thinking more it need not be public as both the drivers are in same `state` package. I will make this function private.

> but I saw your comment about why Watchall can miss some events due to timing. Please clarify.

This function itself is not related to WatchAll missing events. Regarding the comment itself, I realized that condition during my implementation of consul statedriver. Today the way we use watch (which is only when the netplugin instance starts), we first restore all the state (by calling ReadAll()) and then start the vent loop (by calling WatchAll()). This introduces a small time window (between call to ReadAll() and WatchAll()) where any changes in state might get missed.

This is not a major issue today because we use watch to monitor configuration state, which usually is not changing that quickly. Never the less I wanted to note it down so that we are aware of it and can address it if needed.

mapuri:
> I am curious to know if we need to run both state drivers simultaneously. 

No, we don't need to have two data-stores for netplugin to work. An instance of netplugin shall use only one statedriver and that is specified as part of it's configuration. This is important as in an ideal world netplugin shall use the same datastore as the one used by the management-function, so that user doesn't need to configure and run multiple datastores. Eg. netplugin shall use etcd in kubernetes based environment while it shall be able to use consul in libnetwork based environment assuming libnetwork shall put it's state in consul.

erikh:
code LGTM

PR #92, opened 2015-06-08
Fix error handling of docker events so that netplugin does not terminate with the docker daemon.

erikh:
This fixes an issue where terminating docker, or docker's socket not being available, would terminate the netplugin. This change allows the netplugin to be started at any time and docker to be restarted over and over again without interrupting netplugin's operation.

Note that there is an interesting situation that I don't have a good answer for which is the conflated error handling over a single channel for the event handling routines. I don't think this is a good approach.

### DIFF
--- a/netplugin/netd.go
+++ b/netplugin/netd.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"time"
 
 	"github.com/contiv/netplugin/core"
 	"github.com/contiv/netplugin/crt"
@@ -465,19 +466,26 @@ func handleStateEvents(netPlugin *plugin.NetPlugin, crt *crt.CRT, opts cliOpts,
 }
 
 func handleEvents(netPlugin *plugin.NetPlugin, crt *crt.CRT, opts cliOpts) error {
-
 	recvErr := make(chan error, 1)
+	recvEventErr := make(chan error)
 
 	//monitor and process state change events
 	go handleStateEvents(netPlugin, crt, opts, recvErr)
+	startDockerEventPoll(netPlugin, crt, recvEventErr, opts)
 
-	if !opts.nativeInteg {
-		// start docker client and handle docker events
-		// wait on error chan for problems handling the docker events
-		dockerCRT := crt.ContainerIf.(*docker.Docker)
-		dockerCRT.Client.StartMonitorEvents(handleDockerEvents, recvErr,
-			netPlugin, crt)
-	}
+	go func() {
+		for {
+			err := <-recvEventErr
+			if err != nil {
+				log.Warnf("Failure occured talking to docker events. Sleeping for a second and retrying. Error: %s", err)
+			}
+
+			time.Sleep(1 * time.Second)
+			// FIXME note that we are assuming this goroutine crashed, which if not
+			// might generate double events. Synchronize this with a channel.
+			startDockerEventPoll(netPlugin, crt, recvEventErr, opts)
+		}
+	}()
 
 	err := <-recvErr
 	if err != nil {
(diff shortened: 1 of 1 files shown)


erikh:
I'm also having trouble running the system tests in my env so please be patient while I get jenkins to pass.

netplugin/netd.go
```diff
@@ -464,28 +465,36 @@ func handleStateEvents(netPlugin *plugin.NetPlugin, crt *crt.CRT, opts cliOpts,
...
+func handleEvents(netPlugin *plugin.NetPlugin, crt *crt.CRT, opts cliOpts) {
+	recvErr := make(chan error)
 
 	//monitor and process state change events
 	go handleStateEvents(netPlugin, crt, opts, recvErr)
+	startDockerEventPoll(netPlugin, crt, recvErr, opts)
+
+	go func() {
+		for {
+			err := <-recvErr
```
mapuri: I think this will cause `startDockerEventPoll()` to be called even if the error was set by `handleStateEvents()` in which case we should probably exit. May be we can create a separate error channel for `startDockerEventPoll()` to signal errors?

Also even when `startDockerEventPoll()` errors out may be we shall check for a specific error (like `broken pipe` or something more specific that can indicate a `docker` daemon restart) and continue to exit in other scenarios.

erikh: Yeah I wasn’t quite sure how you wanted this to be handled. I’ll try and refresh the PR this evening, if not, tomorrow.

-Erik

erikh: Yeah I wasn’t quite sure how you wanted this to be handled. I’ll try and refresh the PR this evening, if not, tomorrow.

-Erik

erikh:
@mapuri I decided to leave error conditions alone (for now, as we have no diverse actions to handle), but I have segregated the error channels. PTAL.

netplugin/netd.go
```diff
@@ -464,28 +465,37 @@ func handleStateEvents(netPlugin *plugin.NetPlugin, crt *crt.CRT, opts cliOpts,
...
+		for {
+			err := <-recvEventErr
+			if err != nil {
+				log.Warnf("Failure occured talking to docker events. Sleeping for a second and retrying. Error: %s", err)
+			}
+
+			time.Sleep(1 * time.Second)
+			// FIXME note that we are assuming this goroutine crashed, which if not
+			// might generate double events. Synchronize this with a channel.
+			startDockerEventPoll(netPlugin, crt, recvErr, opts)
```
mapuri: I think `recvErr` => `recvEventErr`

netplugin/netd.go
```diff
@@ -464,28 +465,37 @@ func handleStateEvents(netPlugin *plugin.NetPlugin, crt *crt.CRT, opts cliOpts,
...
+func startDockerEventPoll(netPlugin *plugin.NetPlugin, crt *crt.CRT, recvErr chan error, opts cliOpts) {
 	if !opts.nativeInteg {
 		// start docker client and handle docker events
 		// wait on error chan for problems handling the docker events
 		dockerCRT := crt.ContainerIf.(*docker.Docker)
-		dockerCRT.Client.StartMonitorEvents(handleDockerEvents, recvErr,
-			netPlugin, crt)
-	}
-
-	err := <-recvErr
```
mapuri: now with error-channels separated out, we can bring back this code to block on `recvErr` channel inside `handleEvents()` to correctly signal errors to the calling function.

netplugin/netd.go
```diff
@@ -584,10 +594,7 @@ func main() {
 	//logger := log.New(os.Stdout, "go-etcd: ", log.LstdFlags)
 	//etcd.SetLogger(logger)
 
-	err = handleEvents(netPlugin, crt, opts)
-	if err != nil {
```
mapuri: This code can be reverted back to exit on getting state-event errors.

contiv:
Some tests failed!!

contiv:
All tests passed!!

netplugin/netd.go
```diff
@@ -584,10 +601,9 @@ func main() {
...
+	if err := handleEvents(netPlugin, crt, opts); err != nil {
 		os.Exit(1)
-	} else {
-		os.Exit(0)
 	}
+
+	select {} // loop
```
mapuri: This `select{}` loop might not be needed now as `handleEvents()` will block until error.

mapuri:
LGTM, other than one minor comment it is good to merge.

erikh:
OK I can look into it, but the channel is buffered, so it may not have that behavior. Working on native-integ stuff atm so will investigate later.

erikh:
I was wrong about that, feel free to merge after the tests pass.

mapuri:
@erikh  thanks for taking care of the comments

PR #94, opened 2015-06-10
Fix go vet and golint error

mapuri:
And revert back to treating go vet and golint messages as errors


contiv:
All tests passed!!

PR #96, opened 2015-06-12
Vagrant and Docker script changes for consul cluster integration

mapuri:
This set of changes brings in necessary scripting and configuration to bring up a consul cluster in our test/demo environments.

In subsequent PRs I shall bring in consul integration with the netplugin.

**Note:** As part of docker-in-docker script changes, I have moved to using the customized ubuntu image (viz. contiv/ubuntu) that is published and pulled from docker-hub. That image get's updated everytime a change is pushed to our Dockerfile in the repo (by using the automated-build feature of docker hub). This relieves us of managing the image every time Dockerfile changes. But this requires us to push Dockerfile to github first, which implies it requires some manual testing of image on Dockerfile editor's end. I need to see if I can eliminate this requirement while keeping goodness of automated-build.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -86,6 +86,15 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
         node_peers = ""
         node_ips.length.times { |i| node_peers += "#{node_names[i]}=http://#{node_ips[i]}:2380 "}
         node_peers = node_peers.strip().gsub(' ', ',')
+        consul_join_flag = if n > 0 then "-join #{node_ips[0]}" else "" end
+        consul_bootstrap_flag = "-bootstrap-expect=3"
+        if num_nodes < 3 then
+            if n == 0 then
+                consul_bootstrap_flag = "-bootstrap"
+            else
+                consul_bootstrap_flag = ""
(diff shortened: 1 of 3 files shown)


jainvipin:
@mapuri - I understand you already merged the changes; one small question - do we bring up both clusters (etcd and consul) with these changes and per the choice of runtime use one or the other?

mapuri:
> do we bring up both clusters (etcd and consul) with these changes and per the choice of runtime use one or the other?

Yes, that is correct. For the test/demo environment I bring-up both the data stores. The idea is to be able to run systemtests for both the data-stores in one spin-up of the test environment. The data store will be chosen based on the configuration settings (for netplugin, netdcli etc). A little bit refactoring of code is needed there, which I am planning to push in my next PR.

erikh:
I’m seeing some heisenbugs in the unit tests surrounding consul testing. Is it related to this patch?

mapuri:
@erikh do you have failed test logs? I can check. There were some timing related conditions around updating state and starting watch back to back in the tests for which I had added a workaround by adding a sleep. Are you seeing failures in Watch related tests?

PR #97, opened 2015-06-12
Update core.Errorf to calculate the stack trace at the point of construction.

erikh:
This allows us to show a trace, which will be provided on any
`(*core.Error).Error()` (used in logging and printing functions) call if
`CONTIV_TRACE` is set.

Not exactly required but I think will prove handy. The original error messages are enhanced to contain the function name (including the namespace). 

PTAL and let me know what you think.

### DIFF
--- a/core/error_test.go
+++ b/core/error_test.go
@@ -2,6 +2,8 @@ package core
 
 import (
 	"fmt"
+	"os"
+	"strings"
 	"testing"
 )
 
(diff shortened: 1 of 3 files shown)


core/error_test.go
```diff
@@ -10,12 +12,47 @@ func TestErrorStringFormat(t *testing.T) {
...
+	lineNum := 27 // line number where error was formed
+	funcName := "github.com/contiv/netplugin/core.getError"
+
+	expectedStr := fmt.Sprintf("%s [%s %s %d]", msg, funcName, fileName, lineNum)
+
+	if e.Error() != expectedStr {
+		t.Fatalf("Error message yielded an incorrect result with CONTIV_TRACE unset: %s", e.Error())
+	}
+
+	os.Setenv("CONTIV_TRACE", "1")
```
mapuri: *minor comment*: may be we should reset this environment variable at the end of the test. Just so in future if we add more tests for this package it doesn't interfere with them.

mapuri:
thanks @erikh 

LGTM. 

One thought, with your other recent patch for multi-level logging which introduces --debug level flag, may be we should also implicitly set the CONTIV_TRACE environment (when logging level is set to debug) so that we can integrate this function better through a common flag. We can tackle that in this or a separate PR but just wanted to check if that will be useful.

jainvipin:
@LGTM
A stack of error-stacks when an error occurs - cool. Looking forward to see a sample output upon a real error.

erikh:
@jainvipin you can accomplish this with `CONTIV_TRACE=1 netdcli -oper delete -construct vlan-rsrc anything`. It’s not the prettiest thing, but we can massage that over time.

erikh:
I think that should happen, but i’d like to do that in an isolated patch which I will start work on now.

I’ve rebased against the latest master.

erikh:
I’ve applied the -debug patch; PTAL and merge if you’re happy.

PR #98, opened 2015-06-12
netdcli now returns JSON output when queried with `-oper get`

erikh:
I'd like this to eliminate the bespoke parsers we have in a lot of places to scrape netdcli's output.

Note that this breaks pslibnet and k8contivnet; I was hoping since you guys have testbeds, etc for these you could make the appropriate corrections. As you can see in the system tests it's not very complicated to change.

the `drivers` reference while unmarshalling is something we need to shove in a struct in netmaster, I think, to abstract it from OVS. I talked to @mapuri yesterday about a network driver I want to write, and that will alleviate this problem then.

### DIFF
--- a/netdcli/netdcli.go
+++ b/netdcli/netdcli.go
@@ -16,12 +16,12 @@ limitations under the License.
 package main
 
 import (
+	"encoding/json"
 	"flag"
 	"fmt"
 	"net"
 	"os"
 	"strconv"
-	"strings"
 
 	"github.com/contiv/netplugin/core"
 	"github.com/contiv/netplugin/drivers"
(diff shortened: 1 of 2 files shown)


netdcli/netdcli.go
```diff
@@ -502,8 +502,11 @@ func executeOpts(opts *cliOpts) error {
 		} else {
 			//XXX: poor man's pretty print for struct to keep output greppable
```
mapuri: This comment can go now we have a better way :blush:

mapuri:
LGTM, this is much better than scraping cli output

netdcli/netdcli.go
erikh: Oops! Fixing now.

mapuri:
merging this.

Regarding changes to k8contivnet and pslibnet. I think k8contivnet doesn't use the get operations yet so it needn't change right now. There is a new PR in kubernetes to add a `status` cli that will need a read operation, once we add that we can make ise of the new output.

As for pslibnet I was thinking of removing that code once our version of docknet is available so we shall be ok leaving it there for now.

PR #99, opened 2015-06-13
Add utility functions for driver instantiation

mapuri:
This aids sharing code for instantiating drivers (particularly state-drivers) between different binaries (viz. netplugin, netdcli) that use state-drivers directly.

### DIFF
--- a/netdcli/netdcli.go
+++ b/netdcli/netdcli.go
@@ -29,6 +29,8 @@ import (
 	"github.com/contiv/netplugin/netutils"
 	"github.com/contiv/netplugin/resources"
 	"github.com/contiv/netplugin/state"
+	"github.com/contiv/netplugin/utils"
+	"github.com/hashicorp/consul/api"
 
 	log "github.com/Sirupsen/logrus"
 )
@@ -129,7 +131,8 @@ type cliOpts struct {
 	cfgHostBindings bool
 	oper            operation
 	construct       construct
-	etcdURL         string
+	stateStore      string
+	storeURL        string
 	tenant          string
 	netID           string
 	pktTag          string
@@ -182,10 +185,14 @@ func init() {
 		"cfg",
 		false,
 		"JSON file describing the global and network intent. Use '-' to read configuration from stdin")
-	flagSet.StringVar(&opts.etcdURL,
-		"etcd-url",
-		"http://127.0.0.1:4001",
-		"Etcd cluster url")
+	flagSet.StringVar(&opts.stateStore,
+		"state-store",
+		utils.EtcdNameStr,
+		"State store to use")
+	flagSet.StringVar(&opts.storeURL,
+		"store-url",
+		"",
+		"Etcd or Consul cluster url. Empty string resolves to respective state-store's default URL.")
 	flagSet.StringVar(&opts.tenant,
 		"tenant",
 		"default",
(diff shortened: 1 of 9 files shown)


jainvipin:
@mapuri - looks much cleaner - thanks for making this change.
overall changes LGTM.

One small comment: similar to netdcli, do we want to allow specifying the statedriver as an arg to the netplugin (binary) as well?

mapuri:
@jainvipin 
> do we want to allow specifying the statedriver as an arg to the netplugin (binary) as well?

netplugin already takes the driver configuration (for state as well network drivers) as a json style configuration parameter, so we don't need a separate argument. On a related note though, in future when netdcli/netmaster requires taking more configuration (say, loading network-driver specific state-translation) then we might actually want to make netdcli/netmaster take all configuration as json formatted string as well (as it is more extensible unlike defining new cli flags). Right now passing state-driver configuration as cli args there seems sufficient.

jainvipin:
Ah... yes the entire config json struct can be given as config - that's good.

netdcli/netdcli.go
```diff
@@ -182,10 +186,18 @@ func init() {
...
+		"State store to use")
 	flagSet.StringVar(&opts.etcdURL,
 		"etcd-url",
 		"http://127.0.0.1:4001",
 		"Etcd cluster url")
+	flagSet.StringVar(&opts.consulURL,
+		"consul-url",
+		"http://127.0.0.1:8500",
+		"Consul cluster url")
 	flagSet.StringVar(&opts.tenant,
```
erikh: since we know these both take urls, how about `-state-store -connect` instead of the multiple parameters, and then we can sort out the logic for handling connect at `executeOpts` time.

WDYT?

erikh: I meant to comment on the green side of this diff! Sorry.

netdcli/netdcli.go
```diff
@@ -373,46 +385,43 @@ func executeOpts(opts *cliOpts) error {
...
-		log.Fatalf("error %v validating opts\n", err)
+		return err
 	}
 
-	etcdDriver := &state.EtcdStateDriver{}
-	driverConfig := &state.EtcdStateDriverConfig{}
-	driverConfig.Etcd.Machines = []string{opts.etcdURL}
-	config := &core.Config{V: driverConfig}
-	err = etcdDriver.Init(config)
+	stateDriver, err := utils.GetStateDriver()
```
erikh: maybe here we should pass the connection options?

netdcli/netdcli.go
```diff
@@ -498,26 +508,53 @@ func executeOpts(opts *cliOpts) error {
...
-		} else {
-			content, err := json.MarshalIndent(coreState, "", "  ")
-			if err != nil {
-				log.Fatalf("Failed to marshal corestate %+v", coreState)
-			}
-			fmt.Println(string(content))
+			log.Errorf("Failed to read %s. Error: %s", opts.construct.Get(), err)
+			return err
 		}
+		content, err := json.MarshalIndent(coreState, "", "  ")
```
erikh: If we're going to use this MarshalIndent (with no leading spaces and a 2-space indent) regularly, we should make it a function out of it.

netplugin/netd.go
```diff
@@ -602,14 +603,14 @@ func main() {
 
 	defConfigStr := fmt.Sprintf(`{
                     "drivers" : {
-                       "network": "ovs",
-                       "endpoint": "ovs",
+                       "network": "%s",
```
erikh: Note that you can use `%q` here to avoid having to quote yourself.

netdcli/netdcli.go
mapuri: I kept these as separate flags since I wasn't sure if same url convention will apply to all data-stores that we integrate with. Also having per state-store configuration flag allowed me to have different **defaults** for the urls based on the `-state-store` that user specifies, so user can just run netdcli as `netdcli -state-store etcd -cfg <>` to use a etcd state-store with it's default url.

netdcli/netdcli.go
mapuri: `GetStateDriver()` is just a utility to return a previously instantiated singleton state-driver (through a call to `NewStateDriver()`) so it doesn't need to take any arguments. I process the connection flags and init the state-driver instance in netdcli in the `initStateDriver()` function that is called just before `executeOpts()` and `executeJson()`

netdcli/netdcli.go
mapuri: I agree. However, right now we are just using it for netdcli to print the info on screen. And since we have now moved away from screen-scraping with #98 we don't have a need yet to make a common function for this.

Actually I realized, I should have been more clear about some minor changes in this diff. This change is actually related to code refactor to make go-lint happy after some of my changes to return error to the calling function instead of calling Fatalf() .

netplugin/netd.go
mapuri: good point, let me make the change.

netdcli/netdcli.go
erikh: Ah ok. Sorry, I meant to say, -connect could take a free-form string and then the constructor could process that string. I do understand the argument about the defaults, but they could be passed if no arguments are relative to driver.

mapuri: Got it, yes that will work. Let me make that change, for now I will expect the string to be url and process it to a json string as expected by the constructor.

I think the current change is more transient as netdcli/netmaster are expected to quickly evolve to work across different state and network-drivers and have more complex configurations. Once we have that we can start taking a json configuration as argument itself like we do for netplugin.

mapuri:
@erikh I have updated the diffs to use a common flag for accepting the datastore's url. PTAL

erikh:
LGTM

PR #103, opened 2015-06-16
Move OvsDriver to drivers/ovs, adjust constants/packages accordingly

erikh:
This change contains no logic differences, just moves files around. PTAL.


erikh:
Note that I changed `drivers.OvsThing` to `ovs.Thing`. `golint` complained less when I did this, and it made sense now that the OVS driver is in its own package.

mapuri:
@erikh, thanks changes LGTM. This provides a lot cleaner code separation for network drivers.

One additional minor change that will be needed is to add the `github.com/contiv/netplugin/drivers/ovs` package to `scripts/unittests` file inorder to run the tests for the ovs driver.

jainvipin:
@erikh - changes LGTM.

@mapuri, @shaleman, @erikh - Would we have have multiple flavors of OVS drivers e.g. vlan-bridge, vxlan-overlay, etc. In which case do we keep the `drivers/ovs` hierarchy or change it to `drivers/ovsbridge`, and `drivers/ovsoverlay`, etc. Alternatively we keep all under `driver/ovs` and look at the modes underneath. Thoughts?
A side note: I think we might have flavors of ovs drivers

mapuri:
> Would we have have multiple flavors of OVS drivers e.g. vlan-bridge, vxlan-overlay, etc.

Good point. Yes, we might have multiple (independent) flavors of networking functions such that all are based on ovs. I would prefer these implementation to go under their own separate hierachies as suggested in second option i.e. `drivers/ovsvlanbridge`, `drivers/ovsofrouter`... 

right now we just have one single implementation which tries to do both vlan and vxlan in same implementation. Given we will most likely rewrite (and may be move out) the vxlan part, shall we call the current one as `drivers/ovsvlanbridge`?

Just a side note (and slightly off topic), deciding on directory names might become painful as number of driver implementations grow, may be we should also think about avoiding this problem by having a way for driver implementations from other developers to be vendored in netplugin repo instead. I understand we don't have this problem yet but would like to have it :blush:

erikh:
I think for now, keeping it simple is probably best; since these are compiled things (As opposed to API or configuration), we can change it pretty easily later.

Running the tests now, and will push the fixes to the tests in a minute.

shaleman:
I was collecting my thoughts around this. Some of the things I’d like to do as part of bringing in Ofnet is:


  1.  Eliminate host configuration section. We can’t require user to configure all hosts. We should auto-detect all hosts based on where Netplugin is running. Let each netplugin register with etcd/consul and let others discover this. There is an open issue here around how to detect uplink interface in VLAN mode(I.e, which uplink interface to add to OVS)
  2.  Expand the concept of peer hosts. This exists in net master today. But, extend it to netplugin as well and derive VTEP creation based on peer host state.
  3.  I don’t see the need for separate network driver and endpoint driver. We can probably merge these into one. This network driver would have (a) network add/delete (b) endpoint add/delete (c) peer host add/delete APIs. This is not a high priority though.
  4.  I’m not sure separate drivers for vxlan/vlan/vrouter mode is required. Mainly because large part of the code that deals with OVS provisioning will remain the same. Only the part that deals with uplinks/VTEPs and forwarding will be different. I’d prefer it to be different modes in the same network driver. I guess, I’m arguing for one network driver each for  (I)OVS, (ii)linux bridge, (iii)VPP etc. I.e, one network driver for each data path type.
  5.  I have some discomfort around container/endpoint binding. Mainly, I think we should have two level of binding I.e: Container<->network binding and separate container<->host binding. Since these two will come from two different sources. I’m still looking thru the code. I’ll send a follow up email on this.

-sukhesh

mapuri:
@shaleman thanks for sharing your thoughts, below are some replies to your points. 

Sorry for a long response. Also I might be repeating some of the things I have said before and this might not be the PR to have this discussion but I though it might be worth to capture things at one place. We can meet and close this separately if it makes it easier.

>  Eliminate host configuration section. We can’t require user to configure all hosts. We should auto-detect all hosts based on where Netplugin is running. Let each netplugin register with etcd/consul and let others discover this. There is an open issue here around how to detect uplink interface in VLAN mode(I.e, which uplink interface to add to OVS)

The host configuration kind of solves the issue of auto detection of the interface to use for data-path for the driver, as you bring up at the end of your comment. But I agree this configuration need not be pushed by the user, if it can be *auto-generated* or *detected*.

I think there is a subtle point that we need to ponder on here though. Whether the logic of auto-detection should be in **netplugin** or the **network-function**. For netplugin being an infra piece that provides integration between varied management-functions and network-functions, I think this logic is better suited to be in the **network-driver** i.e. if network-driver can auto-detect data path interface this or similar configuration is not needed.

The current ovs-vlan-driver expects this to be passed as configuration as the idea back then was that this info would be easily determined at the higher layer that takes care of bringing in new servers into the cluster and starts services on the server, like say the docker-driver (one way is here in @erikh docknet implementation, https://github.com/erikh/netplugin/blob/docknet/mgmtfn/dockcontivnet/main.go#L109). Again this is initial implementation and I think we have room to change once we come up with new drivers.

>  Expand the concept of peer hosts. This exists in net master today. But, extend it to netplugin as well and derive VTEP creation based on peer host state.

This is actually a conscious design choice to extract out the **state-translation** and logically-centralized **resource allocation** from netplugin and move to netmaster. Accordingly following is the role assignment in current design:
- netplugin instances run per host and get triggered by changes (to all the interesting) states and then in turn trigger the network-driver (in form of events on Network and Endpoint constructs). They don't need to know anything about other netplugin instances as the available state shall be enough to program the local network-driver.
- Whereas netmaster is cluster aware (if needed); consumes the logical-configuration; performs requisite state-translations (from logical-configuration to concrete-configuration) and resource-allocations; and publish the translated state which triggers the netplugin instances.

Now regarding generating vtep configuration, I think the argument here is similar to the host-configuration point above i.e. whether it should be implicit in an infra piece or in the network-driver. If we agree that this is network-driver specific then in the model that I describe above the driver specific **state-translation** nicely captures the place where we should be generating the vtep configuration in netmaster and push trigger netplugin instances. This is what exists today, albeit **state-translation** is hardcoded for ovs today.

> I don’t see the need for separate network driver and endpoint driver. We can probably merge these into one. This network driver would have (a) network add/delete (b) endpoint add/delete (c) peer host add/delete APIs. This is not a high priority though.

`core.EndpointDriver` and `core.NetworkDriver` are just two separate interfaces but can be implemented by single implementation (as is the case with our ovs implementation today). The rationale behind having them as separate interfaces is the following:
- Network and Endpoints are two core constructs that we want the network-drivers to implement. Having these as separate constructs and not just APIs in a single driver construct makes it more explicit and saves us from interface pollution, like should we now have Add/DelAcl() as a method in this interface as well. Having a **minimal** interface keeps the complexity in the driver implementation while ensuring netplugin suitably delivers the integration function it wants to provide.

> I’m not sure separate drivers for vxlan/vlan/vrouter mode is required....I’d prefer it to be different modes in the same network driver. 

While I see the concern with code reuse (which can be solved by having library modules as well). However, there are some gotchas of overloading single implementation with different internal modes that we may need to deal with. These are actually based from what we have today with our mixed vlan-vxlan over ovs implementation. Again just listing them out as things to look out for:
- It bloats up the driver configuration and operational state irrespective of  actual deployment usecase. For instance, even in a pure vlan deployment OvsConfigEndpointState will contain VtepIP field (which is kind of redundant for that deployment).
- It sometimes complicates the state modelling and state-translation logic. A lot of assumptions can be simplified if we are able to design drivers with a certain deployment in mind (like vlan only; vxlan only; L3 only and so on), for instance if we didn't had a mixed vlan and vxlan implementation (where vlan is just not used for external connectivity), we probably didn't had to deal with global reservation and sharing of the vlan tags space (between vlan networks and vxlan networks) as is the case today.

erikh:
I'm going to close this as it's long since lapsed and been dealt wtih in different ways.

PR #104, opened 2015-06-16
Minor refactoring to accommodate state-based resource-managers

mapuri:
Renamed the current ResourceManager implementation to imply that it is based of a distributed statestore. This enables integration with different underlying statestores.

Note that this also now requires the StateStore interfaces to be added/defined for consistent/atomic
read/writes, which shall be handled as separate commit.

### DIFF
--- a/resources/stateresourcemanager.go
+++ b/resources/stateresourcemanager.go
@@ -27,24 +27,57 @@ var resourceRegistry = map[string]reflect.Type{
 	AutoSubnetResource: reflect.TypeOf(AutoSubnetCfgResource{}),
 }
 
-// EtcdResourceManager implements the core.ResourceManager interface.
-// It manages the resource in a logically centralized manner using serialized
-// writes to a etcd based datastore.
-type EtcdResourceManager struct {
-	//XXX: should be '*drivers.EtcdStateDriver', but leaving is
-	//core.StateDriver to get tests going and until the netmaster
-	//is changed to pickup the resource-manager from config
-	Etcd core.StateDriver
+// StateResourceManager implements the core.ResourceManager interface.
+// It manages the resources in a logically centralized manner using serialized
+// writes to underlying state store.
+type StateResourceManager struct {
+	stateDriver core.StateDriver
+}
+
+var gStateResourceManager *StateResourceManager
+
+// NewStateResourceManager instantiates a state based resource manager
+func NewStateResourceManager(sd core.StateDriver) (*StateResourceManager, error) {
+	if gStateResourceManager != nil {
+		return nil, core.Errorf("state-based resource manager instance already exists.")
+	}
+
+	gStateResourceManager = &StateResourceManager{stateDriver: sd}
+	err := gStateResourceManager.Init()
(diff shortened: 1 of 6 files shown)


resources/stateresourcemanager.go
```diff
@@ -27,24 +27,21 @@ var resourceRegistry = map[string]reflect.Type{
...
-type EtcdResourceManager struct {
-	//XXX: should be '*drivers.EtcdStateDriver', but leaving is
-	//core.StateDriver to get tests going and until the netmaster
-	//is changed to pickup the resource-manager from config
-	Etcd core.StateDriver
+// StateResourceManager implements the core.ResourceManager interface.
+// It manages the resources in a logically centralized manner using serialized
+// writes to underlying state store.
+type StateResourceManager struct {
+	StateDriver core.StateDriver
```
erikh: A style nitpick but perhaps `NewStateResourceManager(stateDriver)` would allow us to set this up as an interface, which should make it easier to extend.

mapuri: That's a good suggestion. I think having this and a getter (similar to state-driver utilities) will also help clean some of existing code. Will push a change in a bit.

erikh: Awesome. Thanks!

mapuri:
@erikh and all

I have updated the patch to have resource manager instantiation and access available through appropriate functions. PTAL when you get a chance.

resources/stateresourcemanager.go
```diff
@@ -27,24 +27,57 @@ var resourceRegistry = map[string]reflect.Type{
...
+
+var gStateResourceManager *StateResourceManager
+
+// NewStateResourceManager instantiates a state based resource manager
+func NewStateResourceManager(sd core.StateDriver) (*StateResourceManager, error) {
+	if gStateResourceManager != nil {
+		return nil, core.Errorf("state-based resource manager instance already exists.")
+	}
+
+	gStateResourceManager = &StateResourceManager{stateDriver: sd}
```
erikh: I'm wondering if this should have a mutex wrapped around it?

resources/stateresourcemanager.go
```diff
@@ -27,24 +27,57 @@ var resourceRegistry = map[string]reflect.Type{
...
+	return gStateResourceManager, nil
+}
+
+// ReleaseStateResourceManager releases the singleton instance of the state
+// based resource manager
+func ReleaseStateResourceManager() {
+	if gStateResourceManager != nil {
+		gStateResourceManager.Deinit()
+	}
+	gStateResourceManager = nil
```
erikh: this too

erikh:
LGTM other than the mutex suggestion, which may be ignored if not an issue.

Thanks for doing this!

resources/stateresourcemanager.go
mapuri: mutex is not needed right now as the assumption is that this is a singleton instance that is created just once by the main routine at startup and is never released/modified by other go routines. We can revisit if this assumption changes.

PR #105, opened 2015-06-18
Add system-tests for consul state-store integration

mapuri:
Changes are mostly around modifying the test-infra to take netplugin configuration with consul.

In addition there is a:
- fix in netd.go to correctly setup host-label when netplugin configuration is passed from command-line
- fix in consul driver to trim a leading / in the keys as consul driver rejects the keys otherwise.

### DIFF
--- a/systemtests/singlehost/regression_test.go
+++ b/systemtests/singlehost/regression_test.go
@@ -20,6 +20,7 @@ import (
 	"testing"
 
 	"github.com/contiv/netplugin/systemtests/utils"
+	u "github.com/contiv/netplugin/utils"
 )
 
 func TestOneHostMultipleNets_regress(t *testing.T) {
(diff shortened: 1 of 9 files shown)


systemtests/singlehost/regression_test.go
```diff
@@ -51,7 +51,7 @@ func TestOneHostMultipleNets_regress(t *testing.T) {
 	defer func() {
 		utils.DockerCleanup(t, node1, "myContainer4")
 	}()
-	ipAddress = utils.GetIPAddress(t, node1, "purple-myContainer4")
+	ipAddress = utils.GetIPAddress(t, node1, "purple-myContainer4", "ectd")
```
erikh: you misspelled "etcd" here.

mapuri: ah, good catch.

I should be using constants as this is error prone, let me take care of this.

mapuri:
@erikh updated the hard-coded strings with constants. PTAL

erikh:
LGTM

PR #111, opened 2015-06-25
Make netmaster a daemon servicing REST endpoints for configuration

mapuri:
This PR brings in an initial set of changes required to make netmaster a daemon and expose a REST interface to post logical configurations and read endpoint/network state from netmaster.

As part of this I have refactored some of netmaster code (no logic changes) to separate out the public part of the netmaster i.e. the Configuration itself (into `intent` package under netmaster) and the more private parts that ties together the Configuration processing (aka `state-translation`) and resource management (into `master` package under netmaster).

I created a `client` package to provide some public functions to directly interact with the netmaster instead of having to write a http client from scratch. And I changed netdcli to use the same for posting configuration.

Following are some of the remaining work-items that I am aware of:
- Add version support to the netmaster's interface. I think we still need to think if we need to version the `intent` package or the REST interface itself.
- Changes to support multiple instances of netmaster in a cluster. Right now my understanding is we should be able to run multiple instances behind a load-balancer or a HA-proxy with out any changes but we need to try it out.
- Make netdcli use the 'get' interfaces to query state. Right now I have just changed the parts that do 'post' configuration.
- Add necessary documentation around the `intent` package describing each configuration element and it's attributes.
- Now that we have netmaster as a daemon, we can possibly have it create a default configuration on startup (as desired in #106)

Most of these build on this PR and are pretty are pretty independent. I shall file issues for each and work on them next.

Fixes #108

PR #113, opened 2015-06-30
 Changes to netdcli to read state using the GET interface of netmaster

mapuri:
Other changes:
- a fix in netmaster's get logic to return error if the read fails or endpoint doesn't exist.
- changes in system test to invoke netdcli on the first node where netmaster's single instance is started and binds to localhost address right now.

This depends on #111


erikh:
LGTM

PR #114, opened 2015-07-01
Integrate ofnet with netplugin

shaleman:
This pull request brings in ofnet to netplugin.
Summary of changes:
 - Combine network driver and endpoint driver into one.
- Support multiple OVS instances by introducing the concept of OVS switch instance.
- Use ofnet library for Vxlan flow programming.
- Autodiscover hosts and create VTEPs automatically by creating peer host state.
- add step-up and vlan-intf configuration options.


mapuri:
@shaleman thanks for sharing the diffs.

On a quick look, the changes look good. I will be taking a more closer look later today as well. 

The only outstanding point from my perspective is the combination of `core.NetworkDriver` and `core.EndpointDriver` interfaces and inventing a peer-host construct in the network driver. 

This goes back to my comment for the reason behind current design here https://github.com/contiv/netplugin/pull/103#issuecomment-112590078 and how it forces network-drivers to model their logic around these two constructs and makes it hard to invent new ones. IIRC we decided to go with current design in the last in person meeting as well if it works for the ofnet based driver.

However, I realize we have debated this a lot. I am completely fine if folks are finding the current design limiting and decide to change it wrt the Interfaces exposed to network-drivers, especially since we will now have working code with the new design. ping @jainvipin.

shaleman:
@mapuri,

Only reason I had to merge these two was, util/driverfactory was creating two instances of ovsdriver(one for NetworkDriver and one for EndpointDriver). I think we discussed this briefly over phone. This was making me create multiple instances of OvsSwitch and so on. I couldn’t find any other way to avoid creating two instances of ovsdriver other than merging the two driver interfaces.

Btw, in core/core.go there was a “Network” interface and separate “NetworkDriver” interface. Similarly there was “Endpoint" interface and “EndpointDriver" interface.  Along with driver interface, I ended up merging “Network” and “Endpoint” interfaces too. This is probably not correct. OTOH, I’m not sure if core.Network interface is still being used. May be we can remove it if its dead code.

-sukhesh

mapuri:
> Only reason I had to merge these two was, util/driverfactory was creating two instances of ovsdriver(one for NetworkDriver and one for EndpointDriver). I think we discussed this briefly over phone.

Yeah, we did discuss. The solution I proposed then (also being used by current ovs driver implementation) was to skip creating the bridge again if the it already exists. This code incidentally also helps with driver restart and reloads as well. Not sure if we could have done that here though.

>  in core/core.go there was a “Network” interface and separate “NetworkDriver” interface. 
> ...
> OTOH, I’m not sure if core.Network interface is still being used. 

Actually they are being used (very much implicitly though, so it might not be very obvious). These are the Interfaces that are implemented by `core.Plugin` Interface (`plugin.Netplugin` in our case) to convert north-bound events (from netmaster) and relay it to the network-driver. The idea was to have a place to hook in logic/guarantees that netplugin might need to provide to the drivers (like endpoints creates be delivered after network create; and reverse for deletes etc) on getting triggers from managemen-function. Right now the netplugin implementation of these interfaces is a pass-through as this guarantee is provided by management-functions (like libnetwork). You can check how this is used in `processNetEvent()` here (https://github.com/contiv/netplugin/blob/master/netplugin/netd.go#L90)

shaleman:
Btw, I messed up branching in my fork of netplugin.
I’ll close this pull request and open new one with correct branching..

-sukhesh

shaleman:
closing this pull request and opening new one due to branching problems..

PR #115, opened 2015-07-01
Bring in Openflow controller for vxlan

shaleman:
This pull request adds openflow controller to netplugin.
Summary of changes:

- Bring in openflow controller with vxlan and vrouter support from github.com/contiv/ofnet
- Combine network driver and endpoint driver into one.
- Support multiple OVS bridges by introducing the concept of OVS switch instance.
- Use ofnet library for Vxlan flow programming.
- Autodiscover hosts and create VTEPs automatically by creating peer host state.
- add vtep-ip and vlan-intf configuration options.


Vagrantfile
```diff
@@ -59,6 +59,12 @@ fi
 
 ## install openvswitch and enable ovsdb-server to listen for incoming requests
 #(apt-get install -y openvswitch-switch > /dev/null) || exit 1
+(wget -nv -O ovs-common.deb https://cisco.box.com/shared/static/v1dvgoboo5zgqrtn6tu27vxeqtdo2bdl.deb &&
```
mapuri: We may want to pull and add this to the base image to avoid a pull every time. Also this will need http_proxy settings to be passed to `vagrant up` which will not work in automated jenkins build environment.

Vagrantfile
```diff
@@ -67,6 +73,10 @@ fi
 # wget https://dl.bintray.com/mitchellh/consul/0.5.2_linux_amd64.zip && \
 # unzip 0.5.2_linux_amd64.zip && \
 # mv /tmp/consul /usr/bin) || exit 1
+
+# add vagrant user to docker group
+(usermod -a -G docker vagrant)
```
mapuri: do we need to add `|| exit 1` to exit if this command fails?

drivers/ovsdriver.go
```diff
@@ -441,9 +142,14 @@ func (d *OvsDriver) Init(config *core.Config, info *core.InstanceInfo) error {
...
-	if d.ovs != nil {
-		d.createDeleteBridge(defaultBridgeName, operDeleteBridge)
-		(*d.ovs).Disconnect()
+	log.Infof("Cleaning up ovsdriver")
+
+	// cleanup both vlan and vxlan OVS instances
+	if d.switchDb["vlan"] != nil {
+		d.switchDb["vlan"].Delete()
+	}
+	if d.switchDb["vlan"] != nil {
```
mapuri: `vlan` => `vxlan`

examples/three_host_vxlan.json
```diff
@@ -0,0 +1,50 @@
+
+{
```
mapuri: We should add a system test for this config

mapuri:
other than the conversation from #114 and few minor comment LGTM

Vagrantfile
shaleman: I'm creating a vagrant box with this change right now. Without it Jenkins tests are failing :(

Vagrantfile
shaleman: This is just for our convenience so that we don't have to do 'sudo docker' for every docker command. We can ignore the errors.

drivers/ovsdriver.go
shaleman: Thanks. good catch!

examples/three_host_vxlan.json
shaleman: Good point. I'll look into it.

PR #117, opened 2015-07-07
First revision of dockcontivnet

erikh:
@jainvipin 

There is no readme at this point because the supporting code to trigger these network events does not yet exist in docker. Please see issue https://github.com/docker/docker/pull/14143 for more information.

There are still a lot of things to do here and I can start creating tasks for that.

### DIFF
--- a/mgmtfn/dockcontivnet/main.go
+++ b/mgmtfn/dockcontivnet/main.go
@@ -1,0 +1,501 @@
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"os/exec"
+	"time"
+
+	"github.com/contiv/netplugin/drivers"
+	"github.com/contiv/netplugin/netmaster/client"
+	"github.com/contiv/netplugin/netmaster/intent"
+
+	"github.com/docker/docker/pkg/plugins"
+	"github.com/docker/libnetwork/drivers/remote/api"
+	"github.com/gorilla/mux"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+var (
+	optListen = flag.String("listen", "localhost:4545", "host:port to listen on")
+	optDebug  = flag.Bool("debug", false, "Get debug logging")
+)
+
+func init() {
+	flag.Usage = usage
+}
+
+func usage() {
+	fmt.Printf("Usage: " + os.Args[0] + " <interface-name> <tenant> <network>\n")
+	flag.PrintDefaults()
+	os.Exit(1)
+}
+
+func main() {
+	flag.Parse()
+	if len(flag.Args()) != 3 {
+		flag.Usage()
+	}
+
+	if *optDebug {
+		log.SetLevel(log.DebugLevel)
+		log.Debug("Debug level logging enabled")
+	}
+
+	var (
+		interfaceName = flag.Arg(0)
+		tenantName    = flag.Arg(1)
+		networkName   = flag.Arg(2)
+	)
+
+	hostname, err := os.Hostname()
+	if err != nil {
+		log.Fatalf("Could not retrieve hostname: %v", err)
+	}
+
+	log.Debugf("Configuring router")
+
+	router := mux.NewRouter()
+	s := router.Headers("Accept", "application/vnd.docker.plugins.v1+json").
+		Methods("POST").Subrouter()
+	s.HandleFunc("/Plugin.Activate", activate(hostname, interfaceName))
+	s.HandleFunc("/Plugin.Deactivate", deactivate(hostname, interfaceName))
+	s.HandleFunc("/NetworkDriver.CreateNetwork", createNetwork(tenantName, networkName))
+	s.HandleFunc("/NetworkDriver.DeleteNetwork", deleteNetwork(tenantName, networkName))
+	s.HandleFunc("/NetworkDriver.CreateEndpoint", createEndpoint(tenantName, networkName, hostname))
+	s.HandleFunc("/NetworkDriver.DeleteEndpoint", deleteEndpoint(tenantName, networkName, hostname))
+	s.HandleFunc("/NetworkDriver.EndpointOperInfo", endpointInfo)
+	s.HandleFunc("/NetworkDriver.Join", join(networkName))
+	s.HandleFunc("/NetworkDriver.Leave", leave(networkName))
+
+	if *optDebug {
+		log.Debug("Enabling catchall networkdriver interface")
+		s.HandleFunc("/NetworkDriver.{action:.*}", action)
+	}
+
+	http.ListenAndServe(*optListen, router)
+}
+
+// Catchall for additional driver functions.
+func action(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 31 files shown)


jainvipin:
@erikh: thanks for putting this together - a good start indeed!
It would be good to combine 'dockcontivnet' with another daemon (likely netplugin) to avoid running multiple binaries. On the other hand, k8contivnet is different from this because it is one time invocation and not listening to the REST APIs.

erikh:
OK. I can do this work if you’d like me to. I don’t think it’ll be too hard, just spin up a goroutine which encapsulates the current program’s main() for each tenant/network combination available to the host. it can then push a socket out to the docker plugin system (once it is stable) that the libnetwork can use (hopefully) by way of discovery.

I’ll put in some issues in the morning.

mapuri:
Thanks @erikh, from a quick look this looks good. I will review in more detail later in day as well. 

Regarding docker features still in development, does it make sense to vendor the experimental docker branch and pull outstanding PRs and start adding tests with our plugin? I understand it runs the risk if those PRs don't make it to docker but it might also give us a chance to push for those PRs and suggest enhancements based on our testing. WDYT?

@jainvipin 
> It would be good to combine 'dockcontivnet' with another daemon (likely netplugin) to avoid running multiple binaries. On the other hand, k8contivnet is different from this because it is one time invocation and not listening to the REST APIs.

I am not sure if I agree. dockcontivnet like k8contivnet is a mgmt-function facing logic. And combining one with netplugin and leaving the other out just doesn't fit the netmaster/netplugin design of mgmt-functions interacting with netmaster (and **not** netplugin) that we have right now. Having this separation gives us the ability to run right mix of binaries based on the deployments  (container runtimes, orchestartors etc) without requiring to overload a single binary with multiple functions.

jainvipin:
I understand what you are saying - I am going through the same thoughts. Exposing multiple binaries to make up for our design is something we can do away with internally IMO. How conveniently docker is one binary, interfacing with multiple entities...

>>    It would be good to combine 'dockcontivnet' with another daemon (likely netplugin) to avoid running multiple binaries. On the other hand, k8contivnet is different from this because it is one time invocation and not listening to the REST APIs.

> I am not sure if I agree. dockcontivnet like k8contivnet is a mgmt-function facing logic. And combining one with netplugin and leaving the other out just doesn't fit the netmaster/netplugin design of mgmt-functions interacting with netmaster (and not netplugin) that we have right now. Having this separation gives us the ability to run right mix of binaries based on the deployments (container runtimes, orchestartors etc) without requiring to overload a single binary with multiple functions.

mapuri:
> How conveniently docker is one binary, interfacing with multiple entities...

I agree and that is my point, our design is not same as docker. Do we want to take docker's route? I am not sure because we want to integrate with multiple systems (each having their own subtle differences) which imo are best hidden/abstracted by building a separate binary with well defined interfacing with netmaster/netplugin itself, leaving sussyetm specific complexity in respective binaries. But I am open if we want to revisit our design choice and reshuffle things.

mgmtfn/dockcontivnet/main.go
```diff
@@ -0,0 +1,474 @@
...
+}
+
+func leave(networkName string) func(http.ResponseWriter, *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		logEvent("leave")
+		w.WriteHeader(200)
+	}
+}
+
+func netdcli(payload string, add bool) error {
```
mapuri: the `netdcli` related function can be implemented by using the methods provided by `netmaster/client.Client`. This shall eliminate the need for spawning the netdcli binary and screen scrapping.

mgmtfn/dockcontivnet/main.go
```diff
@@ -0,0 +1,474 @@
...
+	optListen = flag.String("listen", "localhost:4545", "host:port to listen on")
+	optDebug  = flag.Bool("debug", false, "Get debug logging")
+)
+
+func init() {
+	flag.Usage = usage
+}
+
+func usage() {
+	fmt.Printf("Usage: " + os.Args[0] + " <interface-name> <tenant> <network>\n")
```
mapuri: I think this is still following one plugin instance per tenant per network model, which I think can change assuming https://github.com/docker/libnetwork/pull/222 PR will make it to docker, so we can start testing with labels.

This goes back to my earlier comment on us vendoring the experimental docker and start identifying/pushing for the interesting PRs. Looks like the labels PR has merge conflicts so it might need more work before we can start using it but we do need it, right?

erikh:
To answer several questions at once:

1) This is an alpha. It’s not ready for use, yet. I am just committing the first alpha, for use when docker’s networking support arrives. I don’t think it’s a good time to spend a lot of time critiquing it.
2) For single binary, I don’t have a strong opinion, but from a process supervision point, having one binary (instead of several) is easier to run safely when the components depend on each other.

mapuri:
> This is an alpha. It’s not ready for use, yet. I am just committing the first alpha, for use when docker’s networking support arrives. I don’t think it’s a good time to spend a lot of time critiquing it.

sure, SGTM. I provided comments just so we don't loose track of items that are open and/or need discussion. We can track these as separate issues.

> For single binary, I don’t have a strong opinion, but from a process supervision point, having one binary (instead of several) is easier to run safely when the components depend on each other.

We will be having multiple binaries (we already have two viz. netmaster and netplugin). My concern is more about identifying independent functions and *their* interfacing. With single/monolith binaries it's easier to loose track of how the broad functions fit and interface. And it can soon become a code maintenance nightmare as our code base grows.

erikh:
Ok. the only reason this will still not build, is because docker/libnetwork#249 is not merged into docker yet.

It might be a while before we can do that.

erikh:
PTAL and merge :)

mgmtfn/dockcontivnet/main.go
mapuri: This function is not getting used, right?

netplugin/netd.go
```diff
@@ -654,32 +656,39 @@ func handleStateEvents(netPlugin *plugin.NetPlugin, crt *crt.CRT, opts cliOpts,
...
-	startDockerEventPoll(netPlugin, crt, recvEventErr, opts)
 
-	go func() {
+	//monitor and process state change events
+	if opts.nativeInteg {
 		for {
-			err := <-recvEventErr
-			if err != nil {
-				log.Warnf("Failure occured talking to docker events. Sleeping for a second and retrying. Error: %s", err)
+			<-recvErr
```
mapuri: I think we discussed this on slack a few days ago. It's ok to error out if an error is received on `recvErr` channel. This will make error handling consistent with the check at line 685 below.

mgmtfn/dockcontivnet/main.go
erikh: Yeah you're right. Nice catch, thanks.

erikh:
I just pushed two changes, sorry. The first removes that method that you asked about, the second adds better logging (taken from volplugin) that can wrap the requests and output them.

mapuri:
LGTM Eric

I looked at the sanity failure. I think I will need to reproduce it locally to debug, let me try it. If the sanity has passed locally for you I think we can merge this now.

PR #118, opened 2015-07-16
Re-enable dind system tests

shaleman:
I had disabled dind system tests to build a new docker image with ovs2.3.1 and fix some timing issues.
This change will:

- Enable dind system tests
- Pull new contiv/ubuntu docker image with ovs2.3.1 and docker 1.7
- Fix some timing issues seen in dind tests.
- changes dind docker network interfaces to be same as vagrant setup

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -31,7 +31,7 @@
 		},
 		{
 			"ImportPath": "github.com/contiv/ofnet",
-			"Rev": "71287c33566a0b4e552280eddbc66655ac2ac5e1"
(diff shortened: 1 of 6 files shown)


mapuri:
The sleep after every command will slow the entire dind test suite. Not sure if there is a better way to deal with timing issues than a blanket sleep like this though :cry: 

otherwise LGTM

shaleman:
I was thinking about that too.. Actually, I just need little bit delay between starting netplugin and running tests so that RPC connections can be setup  before tests start to run. Let me try that..

shaleman:
I removed all the extra Sleeps and sanity is running fine. So, i'll go ahead and merge this change

PR #120, opened 2015-07-17
Replace inline build with ansible build

erikh:
This replaces the in-line shell build + custom box with an ansible build from an official vagrant box. No make targets or testing has been affected, but I did find a pretty serious issue with our godeps testing it out, which I will fix in a different PR.

My hope is to eventually extend this to provide a little more flexibility as to how the testing environment is configured, e.g., for different OS releases. We can partition this out further into roles as we add new environments or scenarios that we can leverage individual parts for.

### DIFF
--- a/ansible/site.yml
+++ b/ansible/site.yml
@@ -1,0 +1,45 @@
+- hosts: all
+  sudo: true
+  environment: proxy_env
+  tasks:
+    - hostname: "name={{ hostname }}"
+    - template: src=templates/envvar.sh.j2 dest=/etc/profile.d/envvar.sh owner=root group=root mode=0755
+    - apt: "update_cache=yes name={{ item }}"
+      with_items:
+        - unzip
+        - vim-nox
+        - curl
+        - python-software-properties
+        - git 
+    - shell: creates=/usr/local/go/bin/go curl -L https://storage.googleapis.com/golang/go1.4.2.linux-amd64.tar.gz | tar -xvz -C /usr/local
+    - get_url: 
+        url: https://github.com/coreos/etcd/releases/download/v2.0.0/etcd-v2.0.0-linux-amd64.tar.gz
+        dest: /tmp/etcd.tar.gz
+    - shell: creates=/usr/bin/etcd tar vxzf /tmp/etcd.tar.gz && mv etcd-v2.0.0-linux-amd64/etcd* /usr/bin
+    - shell: creates=/usr/bin/docker curl https://get.docker.com | bash
+    - lineinfile:
+        dest: /etc/default/docker
+        line: ". /etc/profile.d/envvar.sh"
+    - user: name=vagrant groups=docker append=yes
+    - get_url:
+        dest: "{{ item.dest }}"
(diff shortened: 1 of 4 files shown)


erikh:
ah; well, ansible will need to be installed on the jenkins host before we can further test this. :)

ansible/site.yml
```diff
@@ -0,0 +1,45 @@
...
+    - hostname: "name={{ hostname }}"
+    - template: src=templates/envvar.sh.j2 dest=/etc/profile.d/envvar.sh owner=root group=root mode=0755
+    - apt: "update_cache=yes name={{ item }}"
+      with_items:
+        - unzip
+        - vim-nox
+        - curl
+        - python-software-properties
+        - git 
+    - shell: creates=/usr/local/go/bin/go curl -L https://storage.googleapis.com/golang/go1.4.2.linux-amd64.tar.gz | tar -xvz -C /usr/local
```
mapuri: It will be nice to not have to download the packages, if they are already installed, as downloading packages adds significantly to the setup time.

Just wondering if ansible allows for it in some way. I was reading about ansible modules being idempotent (http://docs.ansible.com/playbooks_intro.html#tasks-list), but not sure if that is applicable here. Basically I am thinking we can start with a base box, and have ansible provision it once. Then package the box and upload it and use it as our box (without touching the ansible script itself).

mapuri:
thanks @erikh this looks good. May be next step (as a separate PR) could be use this same script for provisioning our docker-in-docker environment as well so that we don;t need to maintain two separate scripts as we do today.

ansible/site.yml
erikh: It does, note the `creates` attributes on several lines. This is one mechanism to determine if something is already installed, packages and the get_url stuff will manage the rest of the idempotent concerns.

I hope to separate this into ansible "build time" and "run time" concerns so that we can build our own images with ease, but I was hoping this could make it in first and I could span out from there.

erikh:
I can do that in this PR. Can you put ansible on the jenkins host, or should I set it up to install via `make` like the other repos?

ansible/site.yml
mapuri: ah, sounds good!

I wasn't sure as I saw the diffs that replace the vagrant box image with `ubuntu/vivid64` above. Is that a pre-configured box or a standard box? If not, we need to replace it with our pre-configured box, right?

mapuri:
+1 

Installing dependencies through `make` is definitely a good idea from ease of getting the user ready. May be we can prompt the user if they are missing dependencies and then install upon getting user's acknowledgement. I am not sure if it a lot of work to address in this PR, let me know and I can install ansible on jenkins host.

mapuri:
@erikh just FYI I went ahead and installed ansible on our jenkins host.

erikh:
Thanks!

-Erik

erikh:
hmm… can you check the version of ansible on the host?

Let’s chat about this on monday, no rush.

-Erik

erikh:
closing for the new vms pr.

PR #121, opened 2015-07-18
CentOS support

jainvipin:
- Add centos targets in Makefile (right now they are not part of sanity)
- Add support to add/delete eps dynamically upon container up/down
- Add system utils to know various system attributes, unit tests
- Added CentOS box in atlas cloud, referred by Vagrantfile for centos tests
- Pull desired ofnet changes and update Godeps


utils/sysutils.go
```diff
@@ -0,0 +1,64 @@
...
+		return err
+	}
+
+	strOutput := string(output)
+	if strings.Contains(strOutput, "CentOS") {
+		SysAttrs.OsType = "centos"
+	} else if strings.Contains(strOutput, "Ubuntu") {
+		SysAttrs.OsType = "ubuntu"
+	} else {
+		SysAttrs.OsType = "unsuppored"
```
erikh: a nit, but "unsupported"

jainvipin: Thanks! For now noted; will fix along with other changes.

PR #126, opened 2015-08-04
Replace 'vagrant ssh' with golang ssh package in system tests

mapuri:
folks, I tried out golang ssh package based on a discussion we had a few days back around improving test times.

This improved the system test run times significantly (about 70%!). PTAL


systemtests/utils/vagrantnode.go
```diff
@@ -15,34 +15,114 @@ limitations under the License.
...
+func (n *VagrantNode) RunCommand(cmd string) error {
+	var (
+		s   *ssh.Session
+		err error
+	)
+
+	if s, err = n.client.NewSession(); err != nil {
+		return err
+	}
+	defer s.Close()
```
jainvipin: Can we keep the session open to avoid the overhead of closing the session all together?

mapuri: Yeah that would have been nice and I tried doing that first but it doesn't work. The ssh package supports only one command execution per session (https://godoc.org/golang.org/x/crypto/ssh#Session.CombinedOutput)

However, from the code the `session.Close()` itself didn't seem too expensive, they just seem to close a channel interface. Most of the work is done as part of `client.Dial()`

mapuri:
The test job now completes in about 21 min than taking over an hour before :smiley: 

http://1f53f76f.ngrok.com/job/Pull-Request%20Build%20Master/buildTimeTrend

systemtests/utils/vagrant.go
```diff
@@ -75,11 +76,52 @@ func (v *Vagrant) Setup(env string, numNodes int) error {
...
+			if string(nodeInfosBytes[j]) == fmt.Sprintf("Host %s", nodeName) {
+				nodeInfoPos = j
+				break
+			}
+		}
+		if nodeInfoPos == -1 {
+			return core.Errorf("Failed to find %q info in vagrant ssh-config output: \n%s\n", nodeName, output)
+		}
+		port := ""
+		if n, err := fmt.Sscanf(string(nodeInfosBytes[nodeInfoPos+1]), "Port %s", &port); n == 0 || err != nil {
```
jainvipin: Minor: Can we avoid sscanf? Can regex not give you all the values natively? I remember doing this in python was very easy... need to dig more about Go.

systemtests/utils/vagrant.go
```diff
@@ -75,11 +76,52 @@ func (v *Vagrant) Setup(env string, numNodes int) error {
 		return err
 	}
 
+	// some more work to figure the ssh port and private key details
+	// XXX: vagrant ssh-config --host <> seems to be broken as-in it doesn't
+	// correctly filter the output based on passed host-name. So filtering
+	// the output ourselves below.
+	if output, err = vCmd.RunWithOutput("ssh-config"); err != nil {
+		return core.Errorf("Error runningg vagrant ssh-config. Error: %s. Output: \n%s\n", err, output)
```
erikh: `runningg`

systemtests/utils/vagrant.go
```diff
@@ -75,11 +76,52 @@ func (v *Vagrant) Setup(env string, numNodes int) error {
 		return err
 	}
 
+	// some more work to figure the ssh port and private key details
```
erikh: we could throw away all of this if we used the vagrant insecure keys, and then just bundled that private key in our repository or on the vagrant image.

erikh: Actually we still need the port number, so disregard. Sorry.

systemtests/utils/vagrantnode.go
```diff
@@ -15,34 +15,114 @@ limitations under the License.
...
 // VagrantNode implements a node in vagrant testbed
 type VagrantNode struct {
-	Name    string
-	NodeNum int
+	Name   string
+	client *ssh.Client
+}
+
+//NewVagrantNode intializes a node in vagrant testbed
+func NewVagrantNode(name string, nodeNum int, port, privKeyFile string) (*VagrantNode, error) {
```
erikh: `nodeNum` is never used inside this function

systemtests/utils/vagrantnode.go
```diff
@@ -15,34 +15,114 @@ limitations under the License.
...
+	)
+
+	if s, err = n.client.NewSession(); err != nil {
+		return "", err
+	}
+	defer s.Close()
+
+	// start and forget about the command as user asked to run in background.
+	// The limitation is we/ won't know if it fails though. Not a worry right
+	// now as the test will fail anyways, but might be good to find a better way.
```
erikh: I wonder if you don't call `s.Wait()`, even in the background, if it leaves zombies lying around?

systemtests/utils/vagrant.go
mapuri: I do use regex to parse the info about host, port, file etc from cli output (multiline etc). But sscanf worked just well without having to form yet another regex for the fields, especially parsing filename, host-name string from the main string.

Any reason to not use sscanf()? I just chose sscanf() for ease of use here.

systemtests/utils/vagrantnode.go
mapuri: good point, it was there in earlier version of VagrantNode but later I decided to clean it up. Will fix this.

systemtests/utils/vagrantnode.go
mapuri: yeah, that is the trick I was able to use for running background processes :)

For `Run()` and `RunWithOutput()` the session internally does a `s.Wait()`

mapuri:
folks, I have updated the diffs with comments incorporated. 

@jainvipin I have kept the sscanf() for now based on our discussion.

PTAL when you get a chance.

PR #127, opened 2015-08-06
Makefile, Vagrantfile: use GOBIN properly as a complement to $GOROOT/bin, not a replacement.

erikh:
@mapuri 
This fixes `$GOBIN` support. `$GOBIN` is the target for `go install -v` and may differ from `$GOROOT/bin` which is the path to the go toolchain.

If you do this right now, you will notice that it can't find godep; this is because on my system GOBIN and GOROOT (both unified) and GOPATH (per-project) are all different.

Signed-off-by: Erik Hollensbe <erik+github@hollensbe.org>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,6 @@ PKGS := `find . -mindepth 1 -maxdepth 1 -type d -name '*' | grep -vE '/\..*$\|Go
 PKGS += `find . -mindepth 2 -maxdepth 2 -type d -name '*'| grep -vE '/\..*$\|Godeps|examples|docs|scripts'`
 TO_BUILD := ./netplugin/ ./netmaster/ ./netdcli/ ./mgmtfn/k8contivnet/ ./mgmtfn/pslibnet/
 HOST_GOBIN := `if [ -n "$$(go env GOBIN)" ]; then go env GOBIN; else dirname $$(which go); fi`
-HOST_GO_BINARY := `dirname $$(which go)`
 HOST_GOROOT := `go env GOROOT`
 
 all: build unit-test system-test system-test-dind centos-tests
@@ -26,10 +25,10 @@ checks:
 	./scripts/checks "$(PKGS)"
 
 build: deps checks
-	rm -rf Godeps/_workspace/pkg
 	godep go install -v $(TO_BUILD)
 
 clean: deps
+	rm -rf Godeps/_workspace/pkg
 	godep go clean -i -v ./...
 
 # setting CONTIV_NODES=<number> while calling 'make demo' can be used to bring
@@ -45,6 +44,9 @@ start-dockerdemo:
 
 clean-dockerdemo:
 	scripts/dockerhost/cleanup-dockerhosts
+
+ssh:
+	@vagrant ssh netplugin-node1 || echo 'Please run "make demo"'
 
 unit-test: build
 	CONTIV_HOST_GOPATH=$(GOPATH) CONTIV_HOST_GOBIN=$(HOST_GOBIN) \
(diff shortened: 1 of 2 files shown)


erikh:
I also added some additional makefile enhancements in the last two commits.

Makefile
```diff
@@ -44,6 +46,9 @@ start-dockerdemo:
 clean-dockerdemo:
 	scripts/dockerhost/cleanup-dockerhosts
 
+ssh:
```
mapuri: do we need to do `vagrant up` or `make demo` first? Otherwise this target will fail right?

Makefile
```diff
@@ -6,7 +6,8 @@
 PKGS := `find . -mindepth 1 -maxdepth 1 -type d -name '*' | grep -vE '/\..*$\|Godeps|examples|docs|scripts|mgmtfn|systemtests'`
 PKGS += `find . -mindepth 2 -maxdepth 2 -type d -name '*'| grep -vE '/\..*$\|Godeps|examples|docs|scripts'`
 TO_BUILD := ./netplugin/ ./netmaster/ ./netdcli/ ./mgmtfn/k8contivnet/ ./mgmtfn/pslibnet/
-HOST_GOBIN := `which go | xargs dirname`
+HOST_GOBIN := `if [ -n "$$(go env GOBIN)" ]; then go env GOBIN; else dirname $$(which go); fi`
+HOST_GO_BINARY := `dirname $$(which go)`
```
mapuri: Is `HOST_GO_BINARY` getting used?

mapuri:
changes look good Erik other than few minor comments.

Makefile
erikh: ah crap.

Makefile
erikh: Yes; I looked into that, but the only good way to do that without running a bunch of other stuff is to eliminate the dependency. This is mostly help for dev.

erikh:
Removed that var. PTAL, merge if you're comfortable.

Makefile
mapuri: Ah, I didn't mean to setup anything extra than what exists. I was wondering if something like below will ensure that vagrant environment is up before trying to ssh:

```
ssh:
   make demo || true
   vagrant ssh netplugin-node1
```
But this works as well I guess sin ssh will fail if environment is not up already.

Makefile
```diff
@@ -25,6 +25,7 @@ checks:
 	./scripts/checks "$(PKGS)"
 
 build: deps checks
+	rm -rf Godeps/_workspace/pkg
```
mapuri: one minor nit should this be part of `clean` target instead?

erikh: I guess I could call clean before build, but the important thing is that it happens before build.

Makefile
erikh: `make demo` always exits 0. the problem is that make demo calls build, which calls several other tasks. I'd like to avoid having those run just to ssh in.

Would this work?

`vagrant ssh netplugin-node1 || echo 'Please run "make demo"'`

Makefile
mapuri: Yeah, I understand. And I was hoping we could avoid clean before every build. 

Basically if we clean every time we are adding to the build time (as this will force a rebuild of vendored packages). 

And if I understand it correctly this cleanup is mostly needed when we are updating to a new version of vendored packages, right? So I was wondering if we move the `rm -rf ...` command to `clean` target then we leave it to the user to call `make clean` once every time they update godeps etc. Would that work?

Makefile
mapuri: yes, looks good.

erikh:
Ok, all should be good now, PTAL.

mapuri:
thanks Erik, LGTM

ok to merge once sanity passes.

PR #128, opened 2015-08-06
Integrate contivModel to netmaster

shaleman:
This adds basic support for contivModel to netmaster.
 - Bring in contivModel from objmodel.
- Integrate tenant and network creation with contivModel.
- Does not remove old REST endpoints from net master and netdcli still goes thru old REST interfaces


netmaster/master/netmaster.go
```diff
@@ -448,108 +447,125 @@ func CreateNetworks(stateDriver core.StateDriver, tenant *intent.ConfigTenant) e
...
 	rm := core.ResourceManager(tempRm)
 
-	err = validateNetworkConfig(tenant)
-	if err != nil {
-		log.Errorf("error validating network config. Error: %s", err)
-		return err
+	// Create network state
+	nwCfg := &drivers.OvsCfgNetworkState{}
+	nwCfg.StateDriver = stateDriver
+	if nwCfg.Read(network.Name) == nil {
```
mapuri: should this check be `!= nil` ?

netmaster/objApi/apiController.go
```diff
@@ -0,0 +1,639 @@
...
+		if err != nil {
+			log.Fatalf("Error creating default tenant. Err: %v", err)
+		}
+	}
+
+	return ctrler
+}
+
+// AppCreate creates app state
+func (ac *APIController) AppCreate(app *contivModel.App) error {
```
mapuri: Just curious about app related handling in netmaster here. Is it a temporary till we find it's best resting place?

netmaster/objApi/apiController.go
```diff
@@ -0,0 +1,639 @@
...
+// AppUpdate updates app
+func (ac *APIController) AppUpdate(app, params *contivModel.App) error {
+	log.Infof("Received AppUpdate: %+v, params: %+v", app, params)
+
+	return nil
+}
+
+// AppDelete delete the app
+func (ac *APIController) AppDelete(app *contivModel.App) error {
+	log.Infof("Received AppDelete: %+v", app)
```
mapuri: I think we need to cleanup the link state we created before here?

netmaster/objApi/apiController.go
```diff
@@ -0,0 +1,639 @@
...
+	log.Infof("Received NetworkUpdate: %+v, params: %+v", network, params)
+	return errors.New("Cant change network parameters after its created")
+}
+
+// NetworkDelete deletes network
+func (ac *APIController) NetworkDelete(network *contivModel.Network) error {
+	log.Infof("Received NetworkDelete: %+v", network)
+
+	// Find the tenant
+	tenant := contivModel.FindTenant(network.TenantName)
```
mapuri: This seems to assume that network would be deleted before tenant is. Is this assured by the orchestrator? If not may be we can let network delete go through anyways?

netmaster/objApi/apiController.go
```diff
@@ -0,0 +1,639 @@
...
+	if tenant == nil {
+		return errors.New("Tenant not found")
+	}
+
+	// Remove link
+	modeldb.RemoveLinkSet(&tenant.LinkSets.Networks, network)
```
mapuri: We need to unlink Network from tenant as well right

netmaster/objApi/apiController.go
```diff
@@ -0,0 +1,639 @@
...
+
+	// Save the tenant too since we added the links
```
mapuri: s/added/remove

netmaster/objApi/apiController.go
```diff
@@ -0,0 +1,639 @@
...
+}
+
+// PolicyDelete deletes policy
+func (ac *APIController) PolicyDelete(policy *contivModel.Policy) error {
+	log.Infof("Received PolicyDelete: %+v", policy)
+	return nil
+}
+
+// ServiceCreate creates service
+func (ac *APIController) ServiceCreate(service *contivModel.Service) error {
```
mapuri: similar to comment for app handling, even service handling will move out of netmaster eventually, right?

netmaster/objApi/apiController.go
```diff
@@ -0,0 +1,639 @@
...
+
+		// save the network
+		err := network.Write()
+		if err != nil {
+			return err
+		}
+	}
+
+	// Check if user specified any endpoint group for the service
+	if len(service.EndpointGroups) == 0 {
```
mapuri: Is it possible for a service to have fewer endpoint-groups than networks? I am wondering if we need to handle that by creating default-endpoint groups in that case as well.

netmaster/objApi/apiController.go
```diff
@@ -0,0 +1,639 @@
...
+
+			// Create default endpoint group for the service
+			err = contivModel.CreateEndpointGroup(&endpointGroup)
+			if err != nil {
+				log.Errorf("Error creating endpoint group: %+v, Err: %v", endpointGroup, err)
+				return err
+			}
+
+			// Add the endpoint group to the list
+			service.EndpointGroups = append(service.EndpointGroups, dfltEpgName)
```
mapuri: Just curious, what is the difference between appending to `service.EndpointGroups` here v/s linking a service to endpoint-groups through `AddLinkSet()` looks like both are relating a service to endpoint-group.

netmaster/objApi/apiController.go
```diff
@@ -0,0 +1,639 @@
...
+		return errors.New("VolumeProfile not found")
+	}
+
+	// fixup default values
+	if service.Scale == 0 {
+		service.Scale = 1
+	}
+
+	// Create service instances
+	for idx := int64(0); idx < service.Scale; idx++ {
```
mapuri: We may need to take care of allocating network resource per instance as well like an IP per instance for a service behind a HAProxy?

netmaster/objApi/apiController.go
```diff
@@ -0,0 +1,639 @@
...
+
+	volProfKey := service.TenantName + ":" + service.VolumeProfile
+	volProfile := contivModel.FindVolumeProfile(volProfKey)
+	if volProfile == nil {
+		log.Errorf("Could not find the volume profile: %s", service.VolumeProfile)
+		return errors.New("VolumeProfile not found")
+	}
+
+	// fixup default values
+	if service.Scale == 0 {
```
mapuri: I wonder if default can come from model itself i.e. if a user creates an object and doesn't set some fields then those fields automatically assume good defaults than just 0?

netmaster/objApi/apiController.go
```diff
@@ -0,0 +1,639 @@
...
+		return err
+	}
+
+	// Create private network for the tenant
+	err = contivModel.CreateNetwork(&contivModel.Network{
+		Key:         tenant.TenantName + ":" + "privateNet",
+		IsPublic:    false,
+		IsPrivate:   true,
+		Encap:       "vxlan",
+		Subnet:      "10.1.0.0/16",
```
mapuri: I am a bit wary about us always choosing these values by default inside netmaster as we might not always choose a good value. Should we have a way for user to provide these instead?

netmaster/objApi/apiController.go
```diff
@@ -0,0 +1,639 @@
...
+func (ac *APIController) TenantDelete(tenant *contivModel.Tenant) error {
+	log.Infof("Received TenantDelete: %+v", tenant)
+
+	// Get the state driver
+	stateDriver, err := utils.GetStateDriver()
+	if err != nil {
+		return err
+	}
+
+	// FIXME: Should we walk all objects under the tenant and delete it?
```
mapuri: I think we need to unlink the (network, service, app etc) objects here, so that they can go to unresolved state and trigger appropriate cleanups.

`DeleteTenantId()` used to delete the networks and endpoints under the tenant so I think in it's current form it will take care of cleaning up state from netmaster and netplugins. Volumes, epgs etc might need to be handled similar way.

netmaster/master/netmaster.go
shaleman: This is taken as is from CreateNetworks() code in existing netnaster. Here is what the original code looks like:
for _, network := range tenant.Networks {
		nwCfg := &drivers.OvsCfgNetworkState{}
		nwCfg.StateDriver = stateDriver
		if nwCfg.Read(network.Name) == nil {
			// TODO: check if parameters changed and apply an update if needed
			continue
		}

I think the plan was to create the network if it doesn't exist, but don't update it. Effectively, we don't handle network parameter updates.

netmaster/objApi/apiController.go
shaleman: Ignore all objects except tenant and network for now. I'll remove all the objects that are not needed in net master in a subsequent commit.

netmaster/objApi/apiController.go
shaleman: This is just a stub code. please ignore it for now. It'll be cleaned up soon.

netmaster/objApi/apiController.go
shaleman: I'm not assuming any orchestrator. Network object always assumes there is a tenant object its referring to. But, i see your point. I'll change it to delete network even it can't find the tenant.

netmaster/objApi/apiController.go
shaleman: Since network object itself is being deleted, we don't need to clean it up. I'm only removing the link from tenant ->network.

netmaster/objApi/apiController.go
shaleman: Thanks

netmaster/objApi/apiController.go
shaleman: Correct. This will move out of netmaster.

netmaster/objApi/apiController.go
shaleman: Good point. Let me think about how to make sure service->network and service->endpointGroup->network bindings are consistent. I'll take this up when i'm adding endpoint group support.

shaleman:
Will re-submit this pull request with most recent changes..

PR #130, opened 2015-08-10
Makefile, Vagrantfile: build on the VM instead of the host.

erikh:
This builds netplugin on the VM and drops it in `$PWD/bin`. unit tests and builds are working both on OS X and Linux. system tests still require linux, but I think we can make this work with a third vm :imp: -- not today, though.

Note that almost all the mounts were removed in place of two static mounts that make the build pipeline much simpler. I do not know if the tests depend on these variables in some way, but they did not seem to manipulate testing at all when I removed them from the Vagrantfile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,8 @@
 
 # find all verifiable packages.
 # XXX: explore a better way that doesn't need multiple 'find'
-PKGS := `find . -mindepth 1 -maxdepth 1 -type d -name '*' | grep -vE '/\..*$\|Godeps|examples|docs|scripts|mgmtfn|systemtests'`
-PKGS += `find . -mindepth 2 -maxdepth 2 -type d -name '*'| grep -vE '/\..*$\|Godeps|examples|docs|scripts'`
+PKGS := `find . -mindepth 1 -maxdepth 1 -type d -name '*' | grep -vE '/\..*$\|Godeps|examples|docs|scripts|mgmtfn|systemtests|bin'`
+PKGS += `find . -mindepth 2 -maxdepth 2 -type d -name '*'| grep -vE '/\..*$\|Godeps|examples|docs|scripts|bin'`
 TO_BUILD := ./netplugin/ ./netmaster/ ./netdcli/ ./mgmtfn/k8contivnet/ ./mgmtfn/pslibnet/ ./mgmtfn/dockcontivnet/
 HOST_GOBIN := `if [ -n "$$(go env GOBIN)" ]; then go env GOBIN; else dirname $$(which go); fi`
 HOST_GOROOT := `go env GOROOT`
@@ -24,8 +24,11 @@ deps:
 checks:
 	./scripts/checks "$(PKGS)"
 
-build: deps checks
+run-build: deps checks
 	godep go install -v $(TO_BUILD)
+
+build: start
+	vagrant ssh netplugin-node1 -c 'sudo -i bash -lc "source /etc/profile.d/envvar.sh && cd /opt/golang/src/github.com/contiv/netplugin && make run-build"'
 
 clean: deps
 	rm -rf Godeps/_workspace/pkg
@@ -33,8 +36,13 @@ clean: deps
 
 # setting CONTIV_NODES=<number> while calling 'make demo' can be used to bring
 # up a cluster of <number> nodes. By default <number> = 1
-demo: build
-	CONTIV_HOST_GOBIN=$(HOST_GOBIN) CONTIV_HOST_GOROOT=$(HOST_GOROOT) vagrant up
+start: 
+	vagrant up
+
+stop:
+	vagrant destroy -f
+
+demo: start build
 
 clean-demo:
 	vagrant destroy -f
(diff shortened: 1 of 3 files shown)


erikh:
can jenkins allocate a pty? that's what's breaking the build, I think.

mapuri:
@erikh,  From build failures it looks like it is trying to download the godep and golint packages and failing as our jenkins build happens behind a proxy and the http_proxy information is not set/passed to the vagrant vm.

To pass that info we can set the CONTIV_ENV variable before doing `vagrant up`

I will review rest of the changes and post my comments in a bit.

Makefile
```diff
@@ -24,17 +24,25 @@ deps:
 checks:
 	./scripts/checks "$(PKGS)"
 
-build: deps checks
```
mapuri: @erikh IIRC from out last discussion we wanted to do host native build inside a container and then have a build iniside vm (just for mac users).

Just wondering if you want to address that as part of this PR? as otherwise the build might fail for folks trying out netplugin inside a vm.

Makefile
```diff
@@ -24,17 +24,25 @@ deps:
...
+run-build: deps checks
 	godep go install -v $(TO_BUILD)
 
+build: start
```
mapuri: I think we can add a direct dependency for `deps` on the `build` target.

This will remove the need for downloading godep and golint inside the vagrant vm. See next comment on keeping the `CONTIV_HOST_GOPATH` env

Makefile
```diff
@@ -48,27 +56,25 @@ clean-dockerdemo:
 ssh:
 	@vagrant ssh netplugin-node1 || echo 'Please run "make demo"'
 
-unit-test: build
-	CONTIV_HOST_GOPATH=$(GOPATH) CONTIV_HOST_GOBIN=$(HOST_GOBIN) \
```
mapuri: On removing these env vars.

`CONTIV_HOST_GOBIN` and `CONTIV_HOST_GOROOT` are used to mount and use host's go environment inside the vagrant vm. This was done to address unit-test failures (which are built inside the vm, see #31 for some historical context) when their is a version mismatch between host's go environment and the one inside vagrant vm. I think we can get rid of these (and also `HOST_GOBIN` and `HOST_GOROOT`), since now we are always able to depend on the sandbox go environment (either container or vm).

`CONTIV_HOST_GOPATH` is used to mount the host's go path inside the vm. Since we used to build natively this would ensure that go tool dependencies like godep and golint are downloaded just once in the host and reused in the vm. I think we can keep this intact as an alternative to having to pass http_proxy information to the vagrant vm using `CONTIV_ENV`

Makefile
```diff
@@ -48,27 +56,25 @@ clean-dockerdemo:
...
-					   CONTIV_HOST_GOROOT=$(HOST_GOROOT) ./scripts/unittests -vagrant
+unit-test: build clean
```
mapuri: just curious why do we need a `clean` after `build`?

Makefile
erikh: This is solved with a static mount now; check the Vagrantfile.

Makefile
erikh: Because this situation happens with the system tests on linux:

* make build fires, builds on a vm, generates pkg dirs
* systemtests fire, which compile on the host, which may have an incompatible set of objects in pkg, and then it fails.

erikh: Happy to accept ideas for how to change it, this is just how it worked out.

Makefile
mapuri: I may be I am missing something. I just see two mounts there, one for the `./` directory (i.e. the dir where Vagrantfile resides)  and other for `./bin` directory.

However, `CONTIV_HOST_GOPATH` used to mount the $GOPATH of the host itself which will automatically mount the `$GOPATH/src` and `$GOPATH/bin` dirs and hence make godep and golint available in the vm as well.

Makefile
mapuri: I see, so essentially the `clean` forces a rebuild on host? It might then fail for MAC, right?

Previously this was solved by having `CONTIV_HOST_GOBIN` and `CONTIV_HOST_GOROOT` vars, which would mount host's go environment inside the vm and avoid the mismatch of the obj files no matter where the build is triggered.

I didn't realize we still had this issue since system tests are being built natively on the host, may be we can keep the instrumentation of `CONTIV_HOST_GOBIN` and `CONTIV_HOST_GOROOT` intact.

erikh: Let me pour over this today and see if I can rectify things.

-Erik



On August 10, 2015 at 3:06:12 PM, Madhav Puri (notifications@github.com) wrote:

In Makefile:

> @@ -48,27 +56,25 @@ clean-dockerdemo:
>  ssh:
>    @vagrant ssh netplugin-node1 || echo 'Please run "make demo"'
>   
> -unit-test: build
> -  CONTIV_HOST_GOPATH=$(GOPATH) CONTIV_HOST_GOBIN=$(HOST_GOBIN) \
> -                                     CONTIV_HOST_GOROOT=$(HOST_GOROOT) ./scripts/unittests -vagrant
> +unit-test: build clean
I see, so essentially the clean forces a rebuild on host? It might then fail for MAC, right?

Previously this was solved by having CONTIV_HOST_GOBIN and CONTIV_HOST_GOROOT vars, which would mount host's go environment inside the vm and avoid the mismatch of the obj files no matter where the build is triggered.

I didn't realize we still had this issue since system tests are being built natively on the host, may be we can keep the instrumentation of CONTIV_HOST_GOBIN and CONTIV_HOST_GOROOT intact.

erikh:
new vms pr replacing this

PR #131, opened 2015-08-14
Misc fixes for DIND environment

mapuri:
- pass force-delete-ep flag for docker version >1.6
- fixed the handling of force-delete-ep in DinD environment
- fixed the command string for starting netplugin and netmaster to get the daemon logs

### DIFF
--- a/netplugin/netd.go
+++ b/netplugin/netd.go
@@ -505,7 +505,6 @@ func handleContainerStop(netPlugin *plugin.NetPlugin, crt *crt.CRT, opts *cliOpt
 				err, output)
 			return err
 		}
-		return err
 	}
 
 	if opts.forceDeleteEp {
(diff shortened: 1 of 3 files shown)


erikh:
Looks good, but I wonder if the stdin issues could be resolved by `docker exec -i`

mapuri:
ah, ok...I need to look into that. But I think `-i` keeps stdin open.

However we were actually trying to close stdin explicitly inside the command, which somehow is not working with commands run using `docker exec -d`, so I removed the extra `0<&-` as it seemed not needed with `docker exec -d` anyways.

What I also noticed was that with the netmaster and netplugin related background commands I could only get the output redirection to work (inside the container) by redirecting stdout and stderr separately to the files i.e. `&> /tmp/file` didin't work but `1>/tmp/file 2>&1` worked.

erikh:
yeah, the `&>` thing is because that's a bashism, not a posix shell thing.

mapuri:
ah that explains it! we use 'sh -c' to run the `docker exec` commands in shell and not bash...

PR #119, opened 2015-07-16
First revision of Setup and Build document

rschmied:
Documented the steps needed to bring up a working development environment from scratch. Uses a VM on top of a VMware hypervisor (for nested virtualization support).

PR #133, opened 2015-08-20
Makefile, Vagrantfile: refactor of build system

erikh:
This implements several features for the build system:

* Builds happen on the VM in a consistent way, so they can happen on non linux
  platforms.
* New builder images include:
  * First class support of 4.0 (Ubuntu) and 4.1 (CentOS) mainline kernels
  * Docker 1.8.1
  * systemd (with proxy support)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,11 +34,15 @@ deps:
 checks:
 	./scripts/checks "$(PKGS)"
 
+host-build:
+	sudo /bin/bash -c 'source /etc/profile.d/envvar.sh; make run-build'
+
 run-build: deps checks
 	godep go install -v $(TO_BUILD)
 
 build: start
 	vagrant ssh netplugin-node1 -c 'sudo -i bash -lc "source /etc/profile.d/envvar.sh && cd /opt/golang/src/github.com/contiv/netplugin && make run-build"'
+	make stop
 
 clean: deps
 	rm -rf Godeps/_workspace/pkg
@@ -55,7 +59,7 @@ demo-centos:
 stop:
 	CONTIV_NODES=$${CONTIV_NODES:-2} vagrant destroy -f
 
-demo: stop start build
+demo: stop start
 
 start-dockerdemo:
 	scripts/dockerhost/start-dockerhosts
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ checks:
 host-build:
 	sudo /bin/bash -c 'source /etc/profile.d/envvar.sh; make run-build'
 
-run-build: deps checks
+run-build: deps checks clean
 	godep go install -v $(TO_BUILD)
 
 build: start
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,8 @@
 
 # find all verifiable packages.
 # XXX: explore a better way that doesn't need multiple 'find'
-PKGS := `find . -mindepth 1 -maxdepth 1 -type d -name '*' | grep -vE '/\..*$\|Godeps|examples|docs|scripts|mgmtfn|systemtests'`
-PKGS += `find . -mindepth 2 -maxdepth 2 -type d -name '*'| grep -vE '/\..*$\|Godeps|examples|docs|scripts'`
+PKGS := `find . -mindepth 1 -maxdepth 1 -type d -name '*' | grep -vE '/\..*$\|Godeps|examples|docs|scripts|mgmtfn|systemtests|bin'`
+PKGS += `find . -mindepth 2 -maxdepth 2 -type d -name '*'| grep -vE '/\..*$\|Godeps|examples|docs|scripts|bin'`
 TO_BUILD := ./netplugin/ ./netmaster/ ./netdcli/ ./mgmtfn/k8contivnet/ ./mgmtfn/dockcontivnet/
 HOST_GOBIN := `if [ -n "$$(go env GOBIN)" ]; then go env GOBIN; else dirname $$(which go); fi`
 HOST_GOROOT := `go env GOROOT`
@@ -34,8 +34,11 @@ deps:
 checks:
 	./scripts/checks "$(PKGS)"
 
-build: deps checks
+run-build: deps checks
 	godep go install -v $(TO_BUILD)
+
+build: start
+	vagrant ssh netplugin-node1 -c 'sudo -i bash -lc "source /etc/profile.d/envvar.sh && cd /opt/golang/src/github.com/contiv/netplugin && make run-build"'
 
 clean: deps
 	rm -rf Godeps/_workspace/pkg
@@ -43,11 +46,16 @@ clean: deps
 
 # setting CONTIV_NODES=<number> while calling 'make demo' can be used to bring
 # up a cluster of <number> nodes. By default <number> = 1
-demo: build
-	CONTIV_HOST_GOBIN=$(HOST_GOBIN) CONTIV_HOST_GOROOT=$(HOST_GOROOT) vagrant up
+start:
+	CONTIV_NODE_OS=${CONTIV_NODE_OS} vagrant up
 
-clean-demo:
-	vagrant destroy -f
+demo-centos:
+	CONTIV_NODE_OS=centos make demo
+
+stop:
+	CONTIV_NODES=2 vagrant destroy -f
+
+demo: stop start build
 
 start-dockerdemo:
 	scripts/dockerhost/start-dockerhosts
@@ -58,28 +66,31 @@ clean-dockerdemo:
 ssh:
 	@vagrant ssh netplugin-node1 || echo 'Please run "make demo"'
 
-unit-test: build
-	CONTIV_HOST_GOPATH=$(GOPATH) CONTIV_HOST_GOBIN=$(HOST_GOBIN) \
-					   CONTIV_HOST_GOROOT=$(HOST_GOROOT) ./scripts/unittests -vagrant
+unit-test: stop clean build
+	./scripts/unittests -vagrant
 
-unit-test-centos: build
-	CONTIV_NODE_OS=centos CONTIV_HOST_GOPATH=$(GOPATH) CONTIV_HOST_GOBIN=$(HOST_GOBIN) \
-					   CONTIV_HOST_GOROOT=$(HOST_GOROOT) ./scripts/unittests -vagrant
+unit-test-centos: stop clean
+	CONTIV_NODE_OS=centos make build
+	CONTIV_NODE_OS=centos ./scripts/unittests -vagrant
 
 # setting CONTIV_SOE=1 while calling 'make system-test' will stop the test
 # on first failure and leave setup in that state. This can be useful for debugging
 # as part of development.
-system-test: build
-	CONTIV_HOST_GOROOT=$(HOST_GOROOT) CONTIV_HOST_GOBIN=$(HOST_GOBIN) CONTIV_HOST_GOPATH=$(GOPATH) godep go test --timeout 30m -run "sanity" \
+system-test: system-test-singlehost system-test-multihost
+
+# the `make stop` here and below are necessary because build leaves around a VM (intentionally)
+system-test-singlehost: stop clean build
+	make stop
+	godep go test -v --timeout 30m -run "sanity" \
 					   github.com/contiv/netplugin/systemtests/singlehost 
-	CONTIV_HOST_GOROOT=$(HOST_GOROOT) CONTIV_HOST_GOBIN=$(HOST_GOBIN) CONTIV_HOST_GOPATH=$(GOPATH) godep go test --timeout 80m -run "sanity" \
+
+system-test-multihost: stop clean
+	make build stop
+	godep go test -v --timeout 80m -run "sanity" \
 					   github.com/contiv/netplugin/systemtests/twohosts
 
-system-test-centos: build
-	CONTIV_NODE_OS=centos CONTIV_HOST_GOPATH=$(GOPATH) godep go test --timeout 30m -run "sanity" \
-					   github.com/contiv/netplugin/systemtests/singlehost
-	CONTIV_NODE_OS=centos CONTIV_HOST_GOPATH=$(GOPATH) godep go test --timeout 90m -run "sanity" \
-					   github.com/contiv/netplugin/systemtests/twohosts
+system-test-centos: stop clean
+	CONTIV_NODE_OS=centos make build stop system-test-singlehost system-test-multihost
 
 centos-tests: unit-test-centos system-test-centos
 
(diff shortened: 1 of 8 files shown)


erikh:
is CI down?

contiv:
build status: https://25b17de9.ngrok.com/job/Pull-Request%20Build%20Master/43/console

Makefile
```diff
@@ -24,20 +24,28 @@ deps:
 checks:
 	./scripts/checks "$(PKGS)"
 
-build: deps checks
+run-build: deps checks
 	godep go install -v $(TO_BUILD)
 
+build: start
+	vagrant ssh netplugin-node1 -c 'sudo -i bash -lc "source /etc/profile.d/envvar.sh && cd /opt/golang/src/github.com/contiv/netplugin && make run-build"'
```
mapuri: I think we should stop the vm once build is done, otherwise it will stay running.

Makefile
```diff
@@ -24,20 +24,28 @@ deps:
...
+start: 
+	CONTIV_NODE_OS=${CONTIV_NODE_OS} vagrant up
+
+demo-centos:
+	CONTIV_NODE_OS=centos make demo
 
-clean-demo:
-	vagrant destroy -f
+stop:
+	CONTIV_NODES=2 vagrant destroy -f
```
mapuri: We should let `CONTIV_NODES` come from user's environment

Makefile
```diff
@@ -24,20 +24,28 @@ deps:
...
+
+demo: stop start build
```
mapuri: I am not sure if adding `stop` here and at multiple places below is the right thing to do as we might accidentally cleanup something that user might have setup. We should probably error out or just use the existing environment?

Makefile
```diff
@@ -48,38 +56,41 @@ clean-dockerdemo:
 ssh:
 	@vagrant ssh netplugin-node1 || echo 'Please run "make demo"'
 
-unit-test: build
-	CONTIV_HOST_GOPATH=$(GOPATH) CONTIV_HOST_GOBIN=$(HOST_GOBIN) \
-					   CONTIV_HOST_GOROOT=$(HOST_GOROOT) ./scripts/unittests -vagrant
+unit-test: stop clean build
```
mapuri: IIRC from discussion in #130 (here: https://github.com/contiv/netplugin/pull/130#discussion_r36670085) the need for `clean` arises due to mismatching go object files when go environment differs between host and vm, right?

doing a `clean` increases build+test time and also need to add it for almost every test target. I am thinking if we should consider keeping the `CONTIV_HOST_GOBIN` and `CONTIV_HOST_GOROOT` vars?

README.md
```diff
@@ -20,6 +20,8 @@ Please do not use this code in production, until code goes through more testing
 
 ###Building and Testing
 
+**Note:** Vagrant 1.7.4 and VirtualBox 5.0+ are required to build and test netplugin.
```
mapuri: if we had #119 merged then most likely this would go as an update to the setup document there. Just wondering if we shall pull that doc in first to avoid an outdated doc? But completely up to you.

ping @rschmied as well to see if he still needs help with the merge.

Vagrantfile
```diff
@@ -1,35 +1,45 @@
...
+/sbin/ip link set eth1 up
+/sbin/ip link set eth2 up
+
 echo 'export GOPATH=#{netplugin_synced_gopath}' > /etc/profile.d/envvar.sh
 echo 'export GOBIN=$GOPATH/bin' >> /etc/profile.d/envvar.sh
 echo 'export GOSRC=$GOPATH/src' >> /etc/profile.d/envvar.sh
-echo 'export GOROOT=#{host_goroot_path}' >> /etc/profile.d/envvar.sh
-echo 'export PATH=$PATH:#{host_goroot_path}/bin:#{host_gobin_path}:$GOBIN' >> /etc/profile.d/envvar.sh
-if [ $# -gt 0 ]; then
-    echo "export $@" >> /etc/profile.d/envvar.sh
```
mapuri: Why not keep using the `CONTIV_ENV` var instead? it solves passing the required environment to vagrant in a general way and can be tooled from the caller, though I agree we only have a usecase for http_proxy for now.

Vagrantfile
```diff
@@ -75,16 +86,16 @@ fi
 # mv /tmp/consul /usr/bin) || exit 1
 
 # add vagrant user to docker group
-(usermod -a -G docker vagrant)
+# (usermod -a -G docker vagrant)
```
mapuri: I think @shaleman added this for ease of running docker commands as vagrant user without having to sudo. Don't we need it anymore?

Vagrantfile
```diff
@@ -109,11 +120,11 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
             end
         end
         config.vm.define node_name do |node|
-            node.vm.hostname = node_name
-            # create an interface for etcd cluster
-            node.vm.network :private_network, ip: node_addr, virtualbox__intnet: "true"
-            # create an interface for bridged network
-            node.vm.network :private_network, ip: "0.0.0.0", virtualbox__intnet: "true"
+            # node.vm.hostname = node_name
```
mapuri: minor: suggest to remove the commented code

Vagrantfile
```diff
@@ -109,11 +120,11 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
+            # # create an interface for etcd cluster
```
mapuri: minor: two hashes instead of one

systemtests/utils/utils.go
```diff
@@ -154,6 +153,8 @@ func StartNetPlugin(t *testing.T, nodes []TestbedNode, nativeInteg bool) {
 
 // StartNetmasterWithFlags starts netplugin on specified testbed nodes with specified flags
 func StartNetmasterWithFlags(t *testing.T, node TestbedNode, flags map[string]string) {
+	time.Sleep(5 * time.Second)
```
mapuri: Is a sleep needed before starting netmaster as well? This is will add an extra 10 sec for every test case :(

Makefile
erikh: the problem with that is that each build will trigger a vm restart -- this is very slow for a build/test cycle.

Makefile
erikh: this is just to ensure that all boxes are destroyed when the multihost system-tests run. I don't set CONTIV_NODES in that case, but there are still two boxes to clean up.

Makefile
erikh: at least for the testing tasks, using `stop` here is absolutely essential. if the binaries are not built on the right VM (e.g., ubuntu vs. centos), the whole thing explodes.

I can remove it from demo, but I figured if someone types `make demo` that they probably want a new set of boxes.

Makefile
erikh: In short, this is quite a bit slower but I think it's worth the trouble. Without it, the build will break between multi-os tests.

If you build on centos, you're going to get a centos-friendly binary in `bin`. Ubuntu will differ. The `clean` ensures the build does not break or produce false-positives because of an object mismatch.

I don't know what the HOST flags are solving here. The goal is to make our CI line up with our real-world environments. Host flags like this change the build toolchain, which may not be what's in the VM.

README.md
erikh: Let's wait for him to answer -- I think he said he was going on vacation. We can always merge that after the fact.

Vagrantfile
erikh: I think it's still in there... https://github.com/contiv/netplugin/pull/133/files#diff-23b6f443c01ea2efcb4f36eedfea9089R40

Vagrantfile
erikh: this is no longer necessary; baked into the image.

Vagrantfile
erikh: yeah I'll do that, sorry.

systemtests/utils/utils.go
erikh: IKR? It sucks.

This is however needed, because the services that netmaster rely on (consul and etcd) may not have fully coordinated/booted and thus be unavailable. I saw this happen numerous times.

Makefile
mapuri: ok, but if user were to just do `make build` and stop. They won't know that we left a vm running behind. And If they decide to yank out their workspace and pull a new one, this vm will stay there consuming resources. 

I might be coming out a bit paranoid but cleanup somewhat seems a lot safer at expense of spending some time :)

Makefile
mapuri: But if someone setups a demo environment with more than two nodes like `CONTIV_NODES=4 make demo`. In that case this will only clean 2 nodes out of 4, right?

Makefile
mapuri: ok, I guess both side of arguments are valid. I am arguing that a user might not want to destroy something that they have setup and are working on. Whereas your argument is also valid, that they might actually want to setup a new environment every time they do `make demo`

It should be fine to keep the `stop` here.

Makefile
mapuri: yeah, my concern is mostly around increased test times and you bring up a good point around consistent toolchain. 

We can keep this as is for now until we run into slow build/test times.

README.md
mapuri: sounds good. We can merge just this in for now then.

Vagrantfile
mapuri: No, I just meant we could use CONTIV_ENV to pass proxy info, something like below:

```
CONTIV_ENV="http_proxy=$http_proxy https_proxy=$https_proxy" make build
```

With current change we seem to be special casing proxy related environment info below. But thinking more, special casing these simplifies the instrumentation needed in the builds. So it shall be ok to keep what we have here, up to you.

systemtests/utils/utils.go
mapuri: From what I noticed during my debugging, it appeared a general raft leader-election instability in the new environment. From consul logs I saw that the leader-election would happen time and again and if a test would run around one of those times had more chances of failing.

Since we don't change anything around etcd/consul configuration itself so starting/stopping netmaster shall really not have any timing affect. So I am not sure if the sleeps are really helping or we are just being lucky.

I don't know if we have a better option atm. May be we can track this separately then and try playing with etcd/consul configuration knobs around keepalives etc and see if that helps.

erikh: So one of the changes I made here was that before, at least in the etcd case, all members of the cluster were actually on host #1 -- you should be able to verify this with `set -x` on in the vagrantfile provisioning script before etcd spawns.

I've watched the logs with this branch at this latest rev, this does not happen anymore, at least I can see. I'm guessing that what you might be seeing is normal periodic leader election that would not occur before.

mapuri: > I'm guessing that what you might be seeing is normal periodic leader election that would not occur before.

I remember seeing read request failures in consul logs with reason 'no leader' so I don't think it is due to a normal periodic election becasue I would assume these databases should stay available under normal conditions. I will retry and double check. We can go ahead with change for now.

Makefile
erikh: correct, let me adjust it to be a minimum of 2.

Makefile
erikh: I don't see how this is any different from typical vagrant usage -- if someone does that without stopping the VM, they will be stuck with the same solution, `pkill VBoxManage` roughly.

I understand your desire but we have to figure out whether keeping the build time down or protecting the user is more important here. I don't know if there's a way to get the best of both worlds.

Vagrantfile
erikh: The systemd stuff is very finicky (e.g., not even putting it into a shell profile and telling systemd to source it works) and was really hard to get right. I'd prefer to just leave it alone until someone really has a problem with it.

systemtests/utils/utils.go
erikh: Ah, I forgot to add this additional context:

so at some point during the b/t cycle for these features (packer, and this code) there was a largeish issue with vbox 5.0.0's networking which got us stuck plumbing the wrong interfaces. I think you hit this last week sometime, so maybe it's an artifact of that.

Makefile
mapuri: > I don't see how this is any different from typical vagrant usage

Yes, this is not different in behavior. The only difference is wrt user's experience. In case of doing `vagrant up` and then deleting the workspace we can consider it to be error on user's part if it leaves behind a vm. But if a user does `make build` and then delete their workspace which leaves behind a vm, it is not an error on their part.

You are right there is no easy way to achieve both in this case i.e. protect the user or improved build time. We can may be print a informative message about the running vm after build to alleviate this a bit. For now, I am ok starting with what we have here as it does solves the builds on mac.

ping @jainvipin @shaleman as well to get their thoughts.

mapuri:
LGTM except for the closure on `make build` behavior, it would be good to get an ok from others as well.

Otherwise good to merge after sanity passes.

erikh:
decided:

`make demo` will do an implicit build.

jainvipin:
@erikh  +1
Overall a great step for making it easy for anyone to try this out. Once checked in, I want to be the first user :-)
- I saw you and @mapuri talking about cleanup increasing the time. Do we know how much is the increase per test run? 
- How does it affect developer's build experience? Does he use the same VM based build? If so, how long does it take, it was taking few seconds earlier.
- I also feel that we need to update docs to suggest new ways of using netplugin: 4 step process: 1) git clone; 2) make demo (with netmaster/netplugin included); 3) Run netdcli with an intent; 4) launch containers, and see the connectivity.

erikh:
1. The cleanup itself is cheap; it takes 20-30 seconds to build the whole netplugin suite. The downside to this newer approach is that to avoid object format conflicts we're going to need to bring up and down a lot more VMs. Ultimately this is the majority of the build time, adn it's not too unpleasant on modern systems.
1. It makes the entire build more consistent at the expense of speed in testing. Yes, the build always comes from the VM now. It can take up to a minute or two to bring the VM up, but we are working on different approaches to making building easier on the guest VMs themselves.
1. I have made a task for this and will do so later today.

erikh:
@mapuri this is completing on everything but DIND, and we recently discussed disabling those tests.

Perhaps now is the time? :)

mapuri:
@erikh, for this particular PR if it's just the known ovs/timing failure I think we should let his one go in as it's a known failure.

For disabling the DIND I am fine turning them off. I am just wondering, if we should watch for a little bit more into recent and future sanity failures that relate to the DIND environment. Somehow I have a feeling that this failure is one off and can be hit in any environment. WDYT?

also ping @jainvipin, to conveniently punt the decision to him :)

erikh:
The test failures are DIND-only now. The heisenbug in the ovs code can be handled out-of-band.

Happy to punt to @jainvipin too :)

jainvipin:
@mapuri, @erikh: let's go without dind, which is not a production scenario we'd like to support anyways.

erikh:
Ok, then I believe this can be merged. If nobody objects I will tomorrow.

mapuri:
sure, SGTM. I have disabled DIND sanity in our jenkins CI. 

We can eventually cleanup the DIND related code as a separate PR. I will file an issue for the same.

PR #142, opened 2015-09-02
Sanity cases

abhinandanpb:
1) Fix has been added to create operEP once the container is deleted and recreated. 
2) Sanity test cases have been added .

### DIFF
--- a/netplugin/netd.go
+++ b/netplugin/netd.go
@@ -452,7 +452,7 @@ func createContainerEpOper(netPlugin *plugin.NetPlugin, contUUID, contName strin
 
 			log.Infof("updating container '%s' with id '%s' \n", contName, contUUID)
 		} else {
-      err = netPlugin.CreateEndpoint(epID)
+			err = netPlugin.CreateEndpoint(epID)
 			if err != nil {
 				log.Errorf("Endpoint creation failed. Error: %s", err)
 				return err
--- a/netplugin/netd.go
+++ b/netplugin/netd.go
@@ -452,7 +452,7 @@ func createContainerEpOper(netPlugin *plugin.NetPlugin, contUUID, contName strin
 
 			log.Infof("updating container '%s' with id '%s' \n", contName, contUUID)
 		} else {
-			err = netPlugin.CreateEndpoint(epID)
+      err = netPlugin.CreateEndpoint(epID)
 			if err != nil {
 				log.Errorf("Endpoint creation failed. Error: %s", err)
 				return err
@@ -499,7 +499,6 @@ func handleContainerStart(netPlugin *plugin.NetPlugin, crt *crt.CRT, opts *cliOp
 			log.Errorf("error attaching container: %v\n", err)
 		}
 	}
-
 	return err
 }
 
@@ -578,7 +577,6 @@ func handleDockerEvents(event *dockerclient.Event, retErr chan error,
 		// it until ep configuration is removed
 		// ep configuration as instantiated can be applied to another container
 		// or reincarnation of the same container
-
 	}
 
 	if err != nil {
--- a/systemtests/twohosts/sanity_test.go
+++ b/systemtests/twohosts/sanity_test.go
@@ -1,6 +1,6 @@
 /***
 Copyright 2014 Cisco Systems Inc. All rights reserved.
- 
+
 Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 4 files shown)


netplugin/netd.go
```diff
@@ -456,6 +456,18 @@ func createContainerEpOper(netPlugin *plugin.NetPlugin, contUUID, contName strin
 				log.Errorf("Endpoint creation failed. Error: %s", err)
 				return err
 			}
+
+			err = operEp.Read(epID)
```
mapuri: do we need to read operEp again? it looks like we already read operEp at line 439

systemtests/twohosts/sanity_test.go
```diff
@@ -1352,3 +1352,163 @@ func TestTwoHostsVxlanAddDelNetworkConsul_sanity(t *testing.T) {
...
+                } ]
+            } ]
+        } ]
+        }`
+
+	utils.ConfigSetupCommon(t, jsonCfg, testbed.GetNodes())
+
+	node1 := testbed.GetNodes()[0]
+	node2 := testbed.GetNodes()[1]
+	for i := 0; i < 10; i++ {
```
mapuri: just wondering if we should also test deletion of just one (random?) endpoint, instead of cleaning up all and recreating them?

systemtests/twohosts/sanity_test.go
```diff
@@ -1352,3 +1352,163 @@ func TestTwoHostsVxlanAddDelNetworkConsul_sanity(t *testing.T) {
...
+		func() {
```
mapuri: we really don't need this func() block here, right?

systemtests/twohosts/sanity_test.go
```diff
@@ -1352,3 +1352,163 @@ func TestTwoHostsVxlanAddDelNetworkConsul_sanity(t *testing.T) {
...
+                "Name"                  : "purple",
+                "Endpoints" : [
+                {
+                    "Container"         : "myContainer2",
+                    "Host"              : "host2"
+                }]
+            } ]
+        } ]
+        }`
+		utils.AddConfig(t, jsonCfg, testbed.GetNodes()[0])
```
mapuri: I think if we always Add same config back to back it will become a no-op starting from the iteration 2. May be we should also delete the config at the end of each loop?

netplugin/netd.go
abhinandanpb: The issue was caught with the new sanity test cases I added where containers are deleted and recreated. On a delete the operstate is completely wiped out. So on a recreate - 439 read is a failure. CreateEndpoint creates the operstate but ContUUID is not added as a part of this. Hence the explicit read to read the operstate and populating the ContID

systemtests/twohosts/sanity_test.go
abhinandanpb: Sure. This testcase was intended to just create and delete containers continously. I can add a new case to cover this.

systemtests/twohosts/sanity_test.go
abhinandanpb: The reason I added func block was to clean up after container create and ping. Just thought this was a better way of using defer cleaning after each iteration

systemtests/twohosts/sanity_test.go
abhinandanpb: Completely agree. Will do the same.

netplugin/netd.go
```diff
@@ -456,6 +456,18 @@ func createContainerEpOper(netPlugin *plugin.NetPlugin, contUUID, contName strin
...
+			if core.ErrIfKeyExists(err) != nil {
```
shaleman: We should not ignore key not found error here. We should write the oper state back only if we found the key

systemtests/twohosts/sanity_test.go
shaleman: Agree with @mapuri. cleanup at the end of the loop is sufficient. defer is useful if there are multiple return points like error handling.

systemtests/twohosts/sanity_test.go
shaleman: +1
Also please check myContainer2 can still ping myContaier1 after deleting the delta config.

unclejack:
This PR can be closed as we don't have that directory any more.

unclejack:
The go based sanity tests have been removed. I'm going to close this PR now.

Issue #30, opened 2015-03-17
relax the go version check make


mapuri:
This issue arises from PRs #17 and #29 .

Current test environment requires a strict go version match between dev (host) and test (vm) environment. This check is too strict and places unnecessary restriction on dev environment that have go ver >=1.4 .  Need to find a better way to address this.

PR #144, opened 2015-09-02
scripts/checks: adjust to ensure go 1.4.2 is in use

erikh:
Signed-off-by: Erik Hollensbe <erik+github@hollensbe.org>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,12 +87,12 @@ unit-test-centos: stop clean
 system-test: system-test-singlehost system-test-multihost
 
 # the `make stop` here and below are necessary because build leaves around a VM (intentionally)
-system-test-singlehost: stop clean
+system-test-singlehost: stop clean checks
 	make build stop
 	godep go test -v --timeout 30m -run "sanity" \
 					   github.com/contiv/netplugin/systemtests/singlehost
 
-system-test-multihost: stop clean
+system-test-multihost: stop clean checks build
 	make build stop
 	godep go test -v --timeout 80m -run "sanity" \
 					   github.com/contiv/netplugin/systemtests/twohosts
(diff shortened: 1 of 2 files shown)


mapuri:
@erikh 

Thinking more this change in itself might not actually work as the `checks` get triggered as part of build which happens inside the vm, so I think they will always succeed. May be we will need to add `checks` as part of `systemtest` target?
Other than that the change looks good.

++++++
And now some ranting ;)

I have a slight reservation with this stricter check that it might come as restricting for users running a slightly old or new version of go1.4 on host. We had similar check before and had to relax it as part of #30. Also I am not sure if current `systemtests` build fine for `mac` users, do you know? I remember @shaleman mentioned that systemtest still don't build on his mac but not sure if he is still hitting it.

Thinking about alternate approaches. I experimented with modifying the `system-test-singlehost` target as follows, which works on my server. But it will not work for `mac` users though. However, just wanted to share in case it seems better than a stricter version check and if we can't run `systemtests` mac for other reasons:

```
system-test-singlehost: clean run-build
     godep go test -v --timeout 30m -run "sanity" \
                      github.com/contiv/netplugin/systemtests/singlehost
```

erikh:
It works for the host build, which IIRC was the problem. The restriction lives to keep people from building bad copies.

The systemtests do not work on mac -- they need the netlink library.

mapuri:
> The restriction lives to keep people from building bad copies.

sgtm

we still need to add `checks` target as part of `systemtest`, right?

erikh:
Yeah, probably a good idea. Let me amend this, one minute.

erikh:
this should be ready after tests pass.

erikh:
Anything left to do here?

mapuri:
changes look good, it needs a rebase though.

erikh:
once CI passes, let's merge this

erikh:
@mapuri I merged this so we could move forward with @shaleman's changes. I'm going to do the same for the systemtests stuff today.

PR #145, opened 2015-09-02
Integrate contivModel and dockplugin into netplugin

shaleman:
Here is the new pull request with squashed commits..

This changes brings in contivModel and dockplugin into netplugin.

With this change you can create network using contivModel and start doing "docker run --publish-service foo.network". Netmaster creates a default tenant and a "private" network at boot up. So, you can just start netmaster/netplugin and immediately start running containers with "docker run --publish-service foo.private" which starts containers in a private vxlan overlay without any additional config.

Unfortunately, this works only on the host where net master is running since dockplugin still tries to use netdcli to talk to netmaster on localhost. Next phase of this change will remove this dependency and introduce an RPC between dockplugin and remote netmaster.

Summary of changes:

Integrate contivModel
merge dockplugin into netplugin. Change dockplugin to use .sock instead of .spec registration.
Move to Veth pairs instead of OVS internal ports
automatically create networks in docker daemon based on contivModel
Query docker daemon using REST interface for network and endpoint parameters during endpoint creating
Bring in libnetwork and associated libraries as godeps.
temporarily Import libnetwork client library from coreOs guys till its merged into libnetwork.


erikh:
in general, please make sure you are using this syntax whenever possible:

```go
if err := callThatErrors(); err != nil {
  panic(err)
}
```

Instead of:

```go
err := callThatErrors()
if err != nil {
  panic(err)
}
```

This reduces the necessity of `err` living in the outer scope, which isn't really a huge deal until you start to re-use err. Then, it becomes dangerous if you don't handle every set of err properly, so your checks don't validate old error conditions.

I can give some examples if you want, but please review your code and make the changes -- we do it elsewhere and I'll get around to fixing those cases eventually, but we should do this for new code.

erikh:
I'll need some more time for this, can we avoid merging this until tomorrow?

shaleman:
Incorporated review comments PTAL.
@erikh: I'll hold the merge till tomorrow. PTAL when you get a chance..

shaleman:
Folks, if there are no further comments, i'll merge this change tonight.

erikh:
I would really like it if this was corrected to use the `if err := ` syntax and all error cases were checked. Please do not merge it until this is done. I have also not made time to review the full source yet, I will tonight.

mapuri:
I agree that we need to follow a common code convention (not just for this project but across all our projects). It will help code look consistent no matter whether we have one or several contributors working on a project. 

How about we follow these (I have seen folks refer these in code review comments in some other projects): https://github.com/golang/go/wiki/CodeReviewComments. 

Regarding scoping of variables: note that the code review guidelines don't explicitly dictate scoping of variables, as there is nothing technically wrong with giving every variable a larger scope than it requires. Usually I have found (and seen it followed in other projects) that code is more readable if the variable (like `err`) can be scoped to a smallest block possible (like an `if` block in this case). So may be we just follow it wherever we can while following the code indent guideline (https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow)?

Regarding error checks: the review guidelines have something to say about it (https://github.com/golang/go/wiki/CodeReviewComments#handle-errors) and I agree we should follow it strictly.

Thinking back, I have myself violated these guidelines a few times. But may be we can all agree that we follow them and also cleanup wherever we find them amiss in existing code.

my $0.02

erikh:
https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

erikh:
Actually, nevermind. I'm not going to fight with you guys. If you want to do it this way, so be it, even though nobody else does it and neither of you have provided any rationale as to why you do it this way.

mapuri:
@erikh I am agreeing with you but may be it didn't come out well it my comment :relaxed: 

The rationale for keeping scope of variables limited is **readability** (which is I believe you are also asking for) and since golang allows variable definition within the conditional I think we should use that feature.

erikh:
Why was this merged? It needs to be fixed.

PR #146, opened 2015-09-06
create the default gateway for auto-allocated subnets

jainvipin:
The fix creates a default gateway to be the last valid unicast IP address in the subnet. Configures the gateway in 'non dockcontivnet' environment.

PTAL


erikh:
LGTM, but I'm not 100% sure of how the logic should flow here, just used my best intuition.

jainvipin:
@erikh - the change will mainly allow for containers to use routers for external connectivity

netmaster/master/netmaster.go
```diff
@@ -501,17 +501,26 @@ func CreateNetwork(network intent.ConfigNetwork, stateDriver core.StateDriver, t
...
 			return err
 		}
+		subnetIsAllocated = true
 	}
 
 	nwCfg.DefaultGw = network.DefaultGw
-	if nwCfg.DefaultGw == "" {
-		// TBD: allocate per global policy
+	// For auto derived subnets assign gateway ip be the last valid unicast ip the subnet
+	if nwCfg.DefaultGw == "" && subnetIsAllocated {
```
mapuri: I am wondering, how will this work when a container has two interfaces(aka endpoints) in different networks. I think if we derive a default-gw for both networks then this will break.

May be instead of default-gw, we should have gateway per network. And then have a tenant level default-gw config?

Also is it possible to have a few tests for this?

jainvipin: Good point @mapuri. In case there are multiple interfaces, even though it assigns the default gateway to both interfaces, the one added afterwards is going to overwrite the previous one. What we need is a mechanism that can allow specifying which one is the default interface among multiple. Can that be done as a separate PR?

About unit tests: mostly the routines being used (like GetSubnetIP, etc.) are unit tested. I am using them to generate default Gw. Should I move this code inside a function? That way a unit test can be written for the default gw/allocation.

mapuri: > What we need is a mechanism that can allow specifying which one is the default interface among multiple. Can that be done as a separate PR?

Sure, this change, though small, adds a feature imo. I am not sure if we should add just a part of feature, but up to you.

> About unit tests:

@jainvipin I actually meant adding a few system-tests. I was thinking we should be able to test this feature may be by a ping to default-gateway or routing between networks?

jainvipin: > Sure, this change, though small, adds a feature imo. I am not sure if we should add just a part of feature, but up to you.

I can attempt to define a new attribute that allows for tagging an endpoint as 'default'. Doing so would also mean that if there is only one interface this is not needed because that is the default interface.
Also note that this is only a problem when not integrating with dockcontivnet, because libnetwork takes care of adding the only interface as a default interface.

> @jainvipin I actually meant adding a few system-tests. I was thinking we should be able to test this feature may be by a ping to default-gateway or routing between networks?

We have routed-mode and bridge mode: 
- This change is for bridged mode, for example in ACI, where we can punt all traffic up and need default gateway there. For 'bridged mode' to test this I may have to simulate an external router that is outside. We need to cover those perhaps as part of those end-to-end system tests.
- In routed mode, @shaleman can confirm that we are already adding the default gateway appropriately, otherwise 'any to any' connectivity wouldn't work. I need to check if he is adding a default route and if so, can that be combined with this change.

mapuri: > Doing so would also mean that if there is only one interface this is not needed because that is the default interface.

I think this more of a state modelling decision. I am fine tagging an endpoint with a flag or having a tenant level flag. I feel latter needs less error checking but am completely ok with either as long as we test :)

>Also note that this is only a problem when not integrating with dockcontivnet, because libnetwork takes care of adding the only interface as a default interface.

I think libnetwork also depends on the driver to return the default gateway, so we need to solve this there as well.

> This change is for bridged mode, for example in ACI,  ... We need to cover those perhaps as part of those end-to-end system tests.

I am completely ok if we are deciding to add them later. But usually it might be a good practice to back every functionality with tests otherwise we risk having breakages as more changes come in.

jainvipin: > I think this more of a state modelling decision. I am fine tagging an endpoint with a flag or having a tenant level flag. I feel latter needs less error checking but am completely ok with either as long as we test :)

I understand tagging an endpoint as default, however I am not sure what you mean by 'tenant level flag'. How would, In your opinion, a tenant level flag be sufficient? Perhaps indicating that 'choose the first one as default' or 'choose the public-network leg as default'.
Can you shed some more light on what might be needed there?

In discussion with @shaleman it seems like we are covered for 'routed mode' forwarding because the subnet and gw are required elements there in the rest APIs. And it doesn't conflict with my change because the allocation of default gw only happens if subnet itself is dynamically allocated.

> I am completely ok if we are deciding to add them later. But usually it might be a good practice to back every functionality with tests otherwise we risk having breakages as more changes come in.

The ACI system testing is outside the scope of this change for sure. Because there are more elements needed in other repos that I am pushing the diffs in.

mapuri: > How would, In your opinion, a tenant level flag be sufficient? Perhaps indicating that 'choose the first one as default' or 'choose the public-network leg as default'.

I meant have tenant level default-gw config. The complexity with allowing a default-gateway per endpoint (`tagging an endpoint as default`) is that due to ordering of things (which endpoint comes first or later), it becomes tricky to consistently come up with same gateway value or fail in a consistent manner.

jainvipin:
@mapuri, @shaleman PTAL:
- added default 'network' as we discussed
- default network is added automatically when not specified to be the first available network
- if the network is removed, it is withdrawn as a default network
- added unit/system tests around this functionality

netmaster/master/netmaster.go
```diff
@@ -501,17 +498,34 @@ func CreateNetwork(network intent.ConfigNetwork, stateDriver core.StateDriver, t
...
+	defaultNwName, err := gCfg.AssignDefaultNetwork(network.Name)
+	if err != nil {
+		log.Errorf("error assigning the default network. Error: %s", err)
+		return err
+	}
+
+	if network.Name == defaultNwName {
+		nwCfg.DefaultGw = network.DefaultGw
+		// For auto derived subnets assign gateway ip be the last valid unicast ip the subnet
+		if nwCfg.DefaultGw == "" && subnetIsAllocated {
```
mapuri: just wondering, if this check is really required here.

Would `subnetIsAllocated` be false when there is a user-defined gateway provided? If so, I think in either case can treat it as a default gateway, right?

systemtests/singlehost/sanity_test.go
```diff
@@ -194,3 +195,72 @@ func TestSingleHostMultiVlanPingFailure_sanity(t *testing.T) {
 		utils.DockerCleanup(t, node, "myContainer2")
 	}()
 }
+
+// Default Network Assignment and freeing
+func TestSingleHostDefaultNetwork_sanity(t *testing.T) {
```
mapuri: It might be good to test the following:

- verify the `ip route` shows the default gateway as the one corresponding to `orange` network.
- And that when default-network is removed that `ip route` no longer has default gateway set.

netmaster/master/netmaster.go
jainvipin: @mapuri - as we talked, I will keep this as is for now

mapuri: yes, sounds good @jainvipin

systemtests/singlehost/sanity_test.go
jainvipin: Sure, will add the test case for `ip route` based confirmation as seen by the container. On the other hand, for the removal of the network, few questions popped up:
- We do not implicitly delete the endpoints when a network is deleted this means we'll not remove the associated default route as well.

So, perhaps we should take the endpoint deletion upon network deletion separately. For this, I can ensure that endpoint deletion removes the default route.

-

mapuri: sounds good.

jainvipin: @mapuri - PTAL

mapuri:
LGTM

Please squash and merge once sanity passes.

PR #143, opened 2015-09-02
Godeps,systemtests/utils: replace systemtests/utils with the new systemtests-utils repo.

erikh:

PR #148, opened 2015-09-17
First commit for policy support

shaleman:
This is the first commit for policy support. This is not fully complete. its work in progress. But, its reasonably stable that we can start playing with it.
Changes include:
- Group based policy support
- Move ofnet master to net master.
- add a discovery logic to discover peers and masters.
- Add some tests to verify policies


erikh:
The `Info` logs in dockplugin can be quite noisy with the `%+v`, can we move those to debug?

drivers/ovsdriver.go
```diff
@@ -377,44 +373,64 @@ func (d *OvsDriver) DeleteEndpoint(id string) (err error) {
...
-func (d *OvsDriver) CreatePeerHost(id string) error {
-	log.Infof("CreatePeerHost for %s", id)
+// AddPeerHost adds VTEPs if necessary
+func (d *OvsDriver) AddPeerHost(node core.ServiceInfo) error {
+	// Nothing to do if this is our own IP
+	if node.HostAddr == d.localIP {
+		return nil
+	}
+
+	log.Infof("CreatePeerHost for %+v", node)
```
erikh: Sorry, I should have commented here. These kinds of messages I think should be debug, that's all.

mgmtfn/dockplugin/dockplugin.go
```diff
@@ -39,8 +39,12 @@ import (
 )
 
 const pluginPath = "/run/docker/plugins"
+
+//const pluginPath = "/usr/share/docker/plugins"
 const driverName = "netplugin"
 
+const useUnixSocket = true
```
erikh: is there some reason we need to keep the old code path if this is always going to be true?

mgmtfn/dockplugin/dockplugin.go
```diff
@@ -74,21 +78,28 @@ func InitDockPlugin() error {
...
-		}
+		go func() {
+			l, err := net.ListenUnix("unix", &net.UnixAddr{Name: driverPath, Net: "unix"})
+			if err != nil {
+				panic(err)
+			}
+
+			log.Infof("docker plugin listening on %s", driverPath)
+			http.Serve(l, router)
+			l.Close()
```
erikh: I think the `l.Close()` should be in a `defer` on line 91 instead of here. I know this wasn't a part of this patch, but it seems like the right time to fix it.

mgmtfn/dockplugin/dockplugin.go
```diff
@@ -307,10 +319,15 @@ func createEndpoint(tenantName, hostname string) func(http.ResponseWriter, *http
 			return
 		}
 
-		// FIXME:
-		GetEndPointName(cereq.NetworkID, cereq.EndpointID)
+		// Get Service name
+		serviceName, err := GetEndPointName(cereq.NetworkID, cereq.EndpointID)
```
erikh: I know this is another surface error (sorry!) but `EndPoint` should probably be capitalized along with our convention elsewhere

netmaster/master/policy.go
```diff
@@ -0,0 +1,436 @@
...
+
+// PolicyAttach attaches a policy to an endpoint and adds associated rules to policyDB
+func PolicyAttach(epg *contivModel.EndpointGroup, policy *contivModel.Policy) error {
+	epgpKey := epg.Key + ":" + policy.Key
+
+	// See if it already exists
+	gp := FindEpgPolicy(epgpKey)
+	if gp != nil {
+		log.Errorf("EPG policy %s already exists", epgpKey)
+		return errors.New("Epg policy exists")
```
erikh: These errors would probably work really well as top-level `var`s that could be exported and compared against, making error handling more fine-grained. I still haven't looked into how this code is being used, though, so maybe this is irrelevant.

Example (the os package implements similar constants):

```go
var ErrExist = errors.New("EPG policy exists")
var ErrNotExist = errors.New("EPG policy does not exist")
```

netmaster/objApi/apiController.go
```diff
@@ -120,15 +131,131 @@ func (ac *APIController) AppDelete(app *contivModel.App) error {
...
+			// find the policy
+			policy := contivModel.FindPolicy(policyKey)
+			if policy == nil {
+				log.Errorf("Could not find policy %s", policyName)
+				return errors.New("Policy not found")
+			}
+
+			// attach policy to epg
+			err := master.PolicyAttach(endpointGroup, policy)
+			if err != nil && !strings.Contains(err.Error(), "Epg policy exists") {
```
erikh: this would be a good place to use the previously suggested constants.

core/core.go
```diff
@@ -87,8 +93,10 @@ type NetworkDriver interface {
 	DeleteNetwork(id string) error
 	CreateEndpoint(id string) error
 	DeleteEndpoint(id string) error
-	CreatePeerHost(id string) error
-	DeletePeerHost(id string) error
+	AddPeerHost(node ServiceInfo) error
```
mapuri: minor:
I think you can just embed `Clustering` interface here as well unless you want to differentiate it from the one in `Plugin`

mgmtfn/dockplugin/dockplugin.go
```diff
@@ -74,21 +78,28 @@ func InitDockPlugin() error {
...
+				panic(err)
+			}
+
+			log.Infof("docker plugin listening on %s", driverPath)
+			http.Serve(l, router)
+			l.Close()
+			log.Infof("docker plugin closing %s", driverPath)
+		}()
+	} else {
+		go func() {
```
mapuri: I think you can just do:

```
go http.ListenAndServe("localhost:4545", router)
```

netmaster/master/netmaster.go
```diff
@@ -773,6 +773,47 @@ func allocSetEpAddress(ep *intent.ConfigEP, epCfg *drivers.OvsCfgEndpointState,
 	return
 }
 
+// CreateEndpoint creates an endpoint
+func CreateEndpoint(stateDriver core.StateDriver, network *intent.ConfigNetwork,
```
mapuri: minor:
does `CreateEndpoint` need to be a exported/public function?

netmaster/master/policy.go
mapuri: +1

BTW we have `core.Error` interface defined for encoding errors and we can use that here as well. It adds some debug related info in error which can be useful.

netmaster/objApi/apiController.go
```diff
@@ -79,6 +80,16 @@ func NewAPIController(router *mux.Router) *APIController {
 	return ctrler
 }
 
+// Utility function to check if string exists in a slice
+func stringInSlice(a string, list []string) bool {
```
mapuri: minor:
As an search optimization we can maintain a `map[string]bool` than `[]string`. But totally optional.

netmaster/objApi/apiController.go
```diff
@@ -120,15 +131,131 @@ func (ac *APIController) AppDelete(app *contivModel.App) error {
...
 	log.Infof("Received EndpointGroupCreate: %+v", endpointGroup)
+
+	// assign unique endpoint group ids
+	endpointGroup.EndpointGroupID = globalEpgID
+	globalEpgID = globalEpgID + 1
+
+	// Find the tenant
+	tenant := contivModel.FindTenant(endpointGroup.TenantName)
+	if tenant == nil {
+		return errors.New("Tenant not found")
```
mapuri: suggest to use `core.Errorf()` instead and elsewhere below.

netplugin/cluster.go
```diff
@@ -0,0 +1,209 @@
...
+	// walk each node and add it as a PeerHost
+	for _, node := range nodeList {
+		// Ignore if its our own info
+		if node.HostAddr == localIP {
+			continue
+		}
+		// add the node
+		err := netplugin.AddPeerHost(core.ServiceInfo{
+			HostAddr: node.HostAddr,
+			Port:     ofnet.OFNET_AGENT_PORT,
```
mapuri: Just wondering, why can't we use the `Port` information from the objstore read here or watch event later than hardcoding it here? Isn't hardcoding makes it less flexible here wrt service discovery?

netplugin/cluster.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	log.Infof("Got netmaster service list: %+v", masterList)
+
+	// Walk each master and add it
+	for _, master := range masterList {
+		// Add the master
+		err := netplugin.AddMaster(core.ServiceInfo{
+			HostAddr: master.HostAddr,
+			Port:     ofnet.OFNET_MASTER_PORT,
```
mapuri: same as above it might be better to use `Port` info from objstore, if we can that is.

scripts/python/testbedApi.py
```diff
@@ -0,0 +1,329 @@
+# Python testing utility classes and functions
```
mapuri: Going through this file I see some overlap with the `systemtests` package that we have. In long run is it intended to replace go-based systemtests for netplugin? Or we will be maintaining both?

I have no strong opinion for go tests v/s python tests per-se but from maintenance purposes I feel it might be good to have one test-infra per project. Again totally up to you.

scripts/python/testbedApi.py
```diff
@@ -0,0 +1,329 @@
...
+import paramiko
+import threading
+import sys
+import os
+
```
mapuri: This comment is meant for `scripts/python/testbedApi.pyc` and `scripts/python/objmodel.pyc` files. Looks like they sneaked in despite the `.gitignore` change.

scripts/python/testbedApi.py
erikh: I don't know python and will be unable to maintain this code. I'd prefer to keep the go code and not introduce new tools if not necessary.

core/core.go
shaleman: I did embed the clustering APIs in NetworkDriver interface. I converted 'PeerHost' interface to 'Clustering' interface for plugin interface. Similar to how there is separate 'Network' and 'Endpoint' interfaces.

drivers/ovsdriver.go
shaleman: These are very useful for debugging right now. Can we make them debug messages later when things are stable.

mgmtfn/dockplugin/dockplugin.go
shaleman: I added it to debug a problem when they changed lib network API from version v1 to v1.1.
Sure. I can remove it.

mgmtfn/dockplugin/dockplugin.go
shaleman: Sure. I'll make the change

mgmtfn/dockplugin/dockplugin.go
shaleman: http.ListenAndServe() blocks forever. So, i had to create a go routine for it.
Anyway, I'll remove this block of code since we'll just use units sockets for now.

mgmtfn/dockplugin/dockplugin.go
shaleman: Are you saying 'P' should not be capitalized in Endpoint?
I think it was capitalized by accident and then got cut and pasted..

netmaster/master/netmaster.go
shaleman: Its called from newly created objApi package. So, it had to be exported.

netmaster/master/policy.go
shaleman: I have seen too many projects fail trying to create central place for all possible error codes. As a compromise lets try creating well known errors per go package( I think thats what Erik was hinting at too) and see if that approach scales. At least when an error needs to be referenced outside the package, it sounds reasonable to say the package needs to export the error.

netmaster/objApi/apiController.go
shaleman: This was required for searching list of policy names in EndpointGroup object. I couldn't build the map since the object comes from contivModel(or i had to build the map each time i need to search it), We don't expect hundreds of policy names per endpoint group. so, performance here should be OK.

netmaster/objApi/apiController.go
shaleman: As i mentioned above. I'm not sure centralized error codes are a good idea. Lets discuss it offline.

netmaster/objApi/apiController.go
shaleman: I'll make the change as discussed above.

netplugin/cluster.go
shaleman: Yes, this is a good point. I'll take this up in next commit.
We are using two separate set of ports for net master. one for REST APIs and one for RPC calls. Either we need to unify the ports or advertise the two as separate 'services'.

scripts/python/testbedApi.py
shaleman: Yeah, I need to figure out how to remove these two binary files.

scripts/python/testbedApi.py
shaleman: These are quick and dirty tests i created so that I can run them on Macs and Linux. I wouldn't read much into it at the moment. We can have bigger discussion on what to do about running system tests on Macs going forward.

Erik, I didn't know python till last week. Learning it was surprisingly easy since syntax is somewhat similar to Golang. Now, i'm feeling like may be I should try ruby too :)

drivers/ovsdriver.go
erikh: That’s fine, but please put in an issue for it.

-Erik

mgmtfn/dockplugin/dockplugin.go
erikh: Yeah, more or less. Sorry for the nit!

-Erik

netmaster/master/policy.go
erikh: I think @mapuri means that we can use the core.Error struct in place of `error` (`fmt.Errorf` or `errors.New`) in our new errors. It implements the `error` interface so I think we'll be ok.

It is a convention in this project, I think it's best we stick with it.

scripts/python/testbedApi.py
erikh: hah. well, I still think if we can avoid introducing new tools, that we should avoid it. The biggest problem with python in our code is that python on the host will differ wildly, anywhere from 2.x to 3.x (which as I understand it are wildly different languages).

To keep a quality build experience we check the go env before we build. We could do this for python too, and in fact I would expect it if we keep this code.

Perhaps this would be better put in a contrib, etc dir in the systemtests-utils library, which is where the systemtests are going,  see #143 .

scripts/python/testbedApi.py
erikh: I can help you with this today, just ping me.

netmaster/master/policy.go
mapuri: yep that's what I meant. so, just morphing @erikh example and adding a suggestion of my own :)
```
var ErrExist = core.Errorf("EPG policy exists")
var ErrNotExist = core.Errorf("EPG policy does not exist")

func ErrExist(type, name string) error { return core.Errorf("%s %q policy exists", type, name) }
...
...
```

Using a common error interface throughout will also enable us to enhance/fix/standardize debug related stuff at a single place.

core/core.go
mapuri: oh yeah I just meant embed it directly as follows but the current form is fine as well:

```
type NetworkDriver interface {
...
Clustering
}
```

mgmtfn/dockplugin/dockplugin.go
mapuri: yep, I understand. Notice the `go` keyword before `http.ListenAndServe()`, it will also run it as a separate go routine.

netplugin/cluster.go
mapuri: ok, sounds good

netmaster/master/policy.go
shaleman: ok. Thanks. Core.Errorf seems like a nice thing to use. I'll change it.

drivers/ovsdriver.go
shaleman: Once the team size grows, we can follow this kind of process. In early stages of the project, its best to let the module owners do the right thing. I have full confidence in this team that all of us will do the right thing :)

shaleman:
I incorporated all review comments PTAL.

drivers/ovsdriver.go
mapuri: just for future reference, `netmaster` and `netplugin` take `--debug` flag which if turned on will print all the debug messages as well.

netmaster/objApi/apiController.go
mapuri: @shaleman do you want to convert errors in these files to `core.Errorf()`?

mapuri:
LGTM, other than one minor comment around leftover `errors.New()` calls.

mapuri:
oh, also whenever we decide to merge please squash before that. This PR has a lot of commits and they won't look clean in master branch's commit logs.

netmaster/objApi/apiController.go
shaleman: I thought i did already. sorry, its sitting in my tree.

drivers/ovsdriver.go
erikh: Ah, I forgot about that! Let’s fix these please since it’s trivial 
to get at them with this flag.

shaleman: --debug flag is not turned on in test runs. so, keeping the logs for now.

shaleman:
what is the problem if master branch has too many commits? isn't that what version control systems for? :)

erikh:
This is an open source project. The commit log is representative of the work we have done between versions.

mapuri:
yeah, commits to master branch shall reflect the final fix/feature that was added. While it is good to track commits in private branch during development but they don't look clean in the master branch itself with multiple developers committing their interim commits there. 

The first response to a similar question below explains it well I think:

http://programmers.stackexchange.com/questions/263164/why-squash-git-commits-for-pull-requests

jainvipin:
I think we should squash commits to make it look like functional atomic blocks of code; I haven't been diligent about it myself but we should do that in order to make it readable for someone to looks at the logs.

shaleman:
I've squashed the commits. I'm going to merge this PR once sanity passes if there are no objections.

PR #150, opened 2015-09-22
add lldp utils

jainvipin:
This commit adds lldp utilities to fetch the lldp information, for information about bindings, etc. The information can be used to configure connected device(s)

### DIFF
--- a/crtclient/docker/docker.go
+++ b/crtclient/docker/docker.go
@@ -373,3 +373,15 @@ func (d *Docker) GetContainerName(contID string) (string, error) {
 
 	return contInfo.Name, nil
 }
+
+// ExecContainer executes a specified in the container's namespace
+func (d *Docker) ExecContainer(contName string, cmdParams ...string) ([]byte, error) {
+	newCmdParams := append([]string{"exec", contName}, cmdParams...)
+	output, err := exec.Command("docker", newCmdParams...).CombinedOutput()
+	if err != nil {
+		log.Errorf("Unable to execute in container namespace. Cmd: %v Error: %s Output: \n%s\n", newCmdParams, err, output)
+		return nil, err
+	}
+
+	return output, err
+}
--- a/utils/lldp.go
+++ b/utils/lldp.go
@@ -1,0 +1,186 @@
+/***
+Copyright 2015 Cisco Systems Inc. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package utils
+
+import (
+	"strings"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/contiv/netplugin/core"
+	"github.com/contiv/netplugin/crt"
+)
+
+const (
+	invalidAttr         = ""
+	lldpadContainerName = "lldpad"
+)
+
+// Implement utilities to fetch lldp information aobut connecting neighbors.
+// The lldp information can be used in non cloud environment to automate
+// networking configuration on the physical devices
+
+// LLDPNeighbor lists the neighbor attributes as learned by the host
+type LLDPNeighbor struct {
+	SystemName        string
+	SystemDescription string
+	PortID            string
+	PortDesc          string
+	CiscoACIPodID     string
+	CiscoACINodeID    string
+}
+
+// FetchLLDPInfo reads the lldp information using lldptool
+// A native go lib could be done but it requires that lldpad daemon is
+// accessible natively on the lldp ipc path
+func FetchLLDPInfo(crt *crt.CRT, ifname string) (*LLDPNeighbor, error) {
+	lldp := &LLDPNeighbor{}
+
+	if crt.GetContainerID(lldpadContainerName) == "" {
+		log.Errorf("Unable to fetch %q container information - please check if it is running", lldpadContainerName)
+		return nil, core.Errorf("unable to fetch container %s info", lldpadContainerName)
+	}
+
+	cmdWithArgs := []string{"lldptool", "-n", "-t", "-i", ifname}
+	output, err := crt.ExecContainer(lldpadContainerName, cmdWithArgs...)
+	if err != nil {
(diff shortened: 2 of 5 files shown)


utils/lldp.go
```diff
@@ -0,0 +1,179 @@
...
+	SystemDescription string
+	PortID            string
+	PortDesc          string
+	CiscoACIPodID     string
+	CiscoACINodeID    string
+}
+
+// FetchLLDPInfo reads the lldp information using lldptool
+// A native go lib could be done but it requires that lldpad daemon is
+// aceessible natively on the lldp ipc path
```
mapuri: s / aceessible / accessible

utils/lldp.go
```diff
@@ -0,0 +1,179 @@
...
+func FetchLLDPInfo(ifname string) (*LLDPNeighbor, error) {
+	lldp := &LLDPNeighbor{}
+
+	output, err := exec.Command("docker", "exec", "lldpad", "lldptool",
+		"-n", "-t", "-i", ifname).CombinedOutput()
+	if err != nil {
+		log.Errorf("Error reading lldp information. Error: %s Output: \n%s\n", err, output)
```
mapuri: since this utility depends on running lldp running in a container, it might also be good to include `Dockerfile` somewhere (`scripts/utils/lldp`?). And perhaps also add checks for container is running and return a more specific (user friendly) error than a more cryptic `docker exec` failure.

utils/lldp.go
```diff
@@ -0,0 +1,179 @@
...
+		switch {
+		case strings.Contains(line, "System Name TLV"):
+			lldp.SystemName, err = lldpReadSystemName(tlvLines)
+		case strings.Contains(line, "System Description TLV"):
+			lldp.SystemDescription, err = lldpReadSystemDescription(tlvLines)
+		case strings.Contains(line, "Port ID TLV"):
+			lldp.PortID, err = lldpReadPortID(tlvLines)
+		case strings.Contains(line, "Port Description TLV"):
+			lldp.PortDesc, err = lldpReadPortDescription(tlvLines)
+		default:
```
mapuri: minor:
I think we don't need the `default` block if we initialize `err` to nil before `switch`

utils/lldp.go
```diff
@@ -0,0 +1,179 @@
...
+			continue
+		}
+		if err != nil {
+			log.Errorf("Error parsing lldp information. Error '%v' \n", err)
+		}
+	}
+	log.Infof("VJ before lldpInfo = %#v \n", lldp)
```
mapuri: suggest to remove these logs or may be make them Debug level. Also please strip the `VJ` prefix :)

utils/lldp.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		log.Errorf("Error parsing lldp information. Error '%v' \n", err)
+	}
+
+	return err
+}
+
+func lldpReadSystemName(attrLines []string) (string, error) {
+	if len(attrLines) < 1 {
+		return "", errors.New("empty attributes")
```
mapuri: suggest to use `core.Error` package here and elsewhere as that is used across this project.

Also you can define a variable for the `empty attributes` and use it elsewhere below.

jainvipin:
@mapuri - thanks for the comments
I took care of vairous comments except the one about lldp's Dockerfile:
- How about if we keep the lldp Dockerfile in the packaging entity, especially if we do not do do docker build in the makefile and/or package the same with netplugin.

I also moved the docker-exec to crt, which is where it belongs IMO, so keeping lldp clean of docker vs. runC formats.

PTAL

crtclient/docker/docker.go
```diff
@@ -373,3 +373,15 @@ func (d *Docker) GetContainerName(contID string) (string, error) {
 
 	return contInfo.Name, nil
 }
+
+// ExecContainer executes a specified in the container's namespace
+func (d *Docker) ExecContainer(contName string, cmdParams ...string) ([]byte, error) {
+	newCmdParams := append([]string{"exec", contName}, cmdParams...)
+	output, err := exec.Command("docker", newCmdParams...).CombinedOutput()
+	if err != nil {
+		log.Errorf("Error reading lldp information. Error: %s Output: \n%s\n", err, output)
```
mapuri: The error string should be more generic here and not refer lldp. Also no need for the extra `\n` at the end, the logger takes care of that.

utils/lldp.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// FetchLLDPInfo reads the lldp information using lldptool
+// A native go lib could be done but it requires that lldpad daemon is
+// accessible natively on the lldp ipc path
+func FetchLLDPInfo(crt *crt.CRT, ifname string) (*LLDPNeighbor, error) {
+	lldp := &LLDPNeighbor{}
+
+	if crt.GetContainerID(lldpadContainerName) == "" {
+		log.Errorf("Error fetching '%s' container information - not sure if it is running", lldpadContainerName)
```
mapuri: I think error should be more certain, so following might be one option. But up to you:

```
Error fetching %q container information - please check if it is running.
```

utils/lldp.go
```diff
@@ -0,0 +1,186 @@
...
+		return nil, core.Errorf("unable to fetch container %s info", lldpadContainerName)
+	}
+
+	cmdWithArgs := []string{"lldptool", "-n", "-t", "-i", ifname}
+	output, err := crt.ExecContainer(lldpadContainerName, cmdWithArgs...)
+	if err != nil {
+		log.Errorf("Error reading lldp information. Error: %s Output: \n%s\n", err, output)
```
mapuri: you can remove the extra `\n` at the end as logger takes care of that.

utils/lldp.go
```diff
@@ -0,0 +1,186 @@
...
+			lldp.SystemName, err = lldpReadSystemName(tlvLines)
+		case strings.Contains(line, "System Description TLV"):
+			lldp.SystemDescription, err = lldpReadSystemDescription(tlvLines)
+		case strings.Contains(line, "Port ID TLV"):
+			lldp.PortID, err = lldpReadPortID(tlvLines)
+		case strings.Contains(line, "Port Description TLV"):
+			lldp.PortDesc, err = lldpReadPortDescription(tlvLines)
+		}
+		if err != nil {
+			log.Errorf("Error parsing lldp information. Error '%v' \n", err)
```
mapuri: you can remove the extra `\n` at the end as logger takes care of that.

utils/lldp.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+	}
+	err = deriveLLDPArgs(lldp)
+	if err != nil {
+		log.Errorf("Error parsing lldp information. Error '%v' \n", err)
```
mapuri: you can remove the extra `\n` at the end as logger takes care of that.

mapuri:
LGTM except for a few minor comments.

ok to merge after squash and sanity pass.

PR #153, opened 2015-09-28
Create contivctl.md

shaleman:
This adds a Readme file for how to use contivctl command line tool

### DIFF
--- a/docs/contivctl.md
+++ b/docs/contivctl.md
@@ -1,0 +1,226 @@
+## contivctl
+
+contivctl command line tool lets you add/modify/delete objects into contiv object model using REST api.
+
+Command can be found under $GOPATH/bin directory (/opt/gopath/bin in vagrant VMs) after a build. 
+Currently, contivctl needs to be run on the same host where netmaster is running
+
+## Usage
+
+```
+vagrant@netplugin-node1:~$ contivctl --help
+usage: contivctl [-h] [--version] {policy,rule,group,network} ...
+
+positional arguments:
+  {policy,rule,group,network}
+    policy              Policy operations
+    rule                Rule add/delete
+    group               Endpoint group operations
+    network             Network operations
+
+optional arguments:
+  -h, --help            show this help message and exit
+  --version             show program's version number and exit
+```
+
+### Creating networks
+
+```
+vagrant@netplugin-node1:~$ contivctl network create --help
+usage: contivctl network create [-h] [-tenantName TENANTNAME]
+                                [-public {yes,no}] [-encap {vlan,vxlan}]
+                                -subnet SUBNET -defaultGw DEFAULTGW
+                                networkName
+
+positional arguments:
+  networkName           Network name
+
+optional arguments:
+  -h, --help            show this help message and exit
+  -tenantName TENANTNAME
+  -public {yes,no}      Is this a public network
+  -encap {vlan,vxlan}   Packet tag
+  -subnet SUBNET        Subnet addr/mask
+  -defaultGw DEFAULTGW  default GW
+```
+
+#### Examples
+
+Following creates a vxlan network called production under default tenant and assigns a subnet to it
+
+```
+vagrant@netplugin-node1:~$ contivctl network create production -encap=vxlan -subnet=20.20.20.0/24 -defaultGw=20.20.20.254
+Creating network default:production
+Network Create response is: {"key":"default:production","defaultGw":"20.20.20.254","encap":"vxlan","isPrivate":true,"networkName":"production","subnet":"20.20.20.0/24","tenantName":"default","link-sets":{},"links":{"Tenant":{"type":"tenant","key":"default"}}}
+```
+
+Following lists all networks under default tenant
+
+```
+vagrant@netplugin-node1:~$ contivctl network list
+Listing all networks for tenant default
+Network		Public	Encap	Subnet			Gateway
+private		No	vxlan	10.1.0.0/16		10.1.254.254
+public		Yes	vlan	192.168.1.0/24		192.168.1.254
+production		No	vxlan	20.20.20.0/24		20.20.20.254
+```
+
+Following deletes a network from default tenant
+
+```
+vagrant@netplugin-node1:~$ contivctl network delete production
+Deleting network default:production
+```
+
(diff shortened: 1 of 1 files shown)


docs/contivctl.md
```diff
@@ -0,0 +1,226 @@
+## contivctl
+
+contivctl command line tool lets you add/modify/delete objects into contiv object model using REST api.
```
mapuri: suggest to use the back-quote markup for keywords like `contivctl`, `netmaster` and paths like `$GOPATH/bin` etc. It brings out the emphasis pretty nicely and makes docs a lot more readable.

docs/contivctl.md
```diff
@@ -0,0 +1,226 @@
...
+
+Command can be found under $GOPATH/bin directory (/opt/gopath/bin in vagrant VMs) after a build.
```
mapuri: suggest to add a **Note:** tag before this line so that it get's necessary attention of the reader.

docs/contivctl.md
```diff
@@ -0,0 +1,226 @@
...
+    rule                Rule add/delete
+    group               Endpoint group operations
+    network             Network operations
+
+optional arguments:
+  -h, --help            show this help message and exit
+  --version             show program's version number and exit
+```
+
+### Creating networks
```
mapuri: May be this section should be called just `Networks` as it walk through examples of creation, listing and deletion.

Same goes for section on policies, rules an groups.

Also wondering if these sections shall start with one-liner descriptions of what a network, group, policy and rule mean. That will give reader some insight into the concepts. We can also put cross-reference to more detailed write-ups in future for interested readers.

docs/contivctl.md
```diff
@@ -0,0 +1,226 @@
...
+  -tenantName TENANTNAME
+  -public {yes,no}      Is this a public network
+  -encap {vlan,vxlan}   Packet tag
+  -subnet SUBNET        Subnet addr/mask
+  -defaultGw DEFAULTGW  default GW
+```
+
+#### Examples
+
+Following creates a vxlan network called production under default tenant and assigns a subnet to it
```
mapuri: It will be nice to use back-quote for keywords/proper-nouns like `production` and `default`, same goes for examples below.

docs/contivctl.md
```diff
@@ -0,0 +1,226 @@
...
+
+optional arguments:
+  -h, --help            show this help message and exit
+  -tenantName TENANTNAME
+
+```
+
+#### Example
+
+Following creates a policy names 'webTier'
```
mapuri: s / a policy names 'webTier' / a policy named `webTier`

docs/contivctl.md
```diff
@@ -0,0 +1,226 @@
...
+  -network NETWORK      Name of network
+  -ipAddress IPADDRESS  IP address/mask
+  -protocol {tcp,udp,icmp,igmp}
+                        IP protocol
+  -port PORT            tcp/udp port number
+  -action {accept,deny}
+                        Accept or deny
+```
+
+All rules have a  rule Id which uniquely identifies the rule under a policy. Rule Id is a string that can be anything user chooses. Rule id does not specify the order in which rules are applied. User has to specify the priority for the rule. Higher the priority, higher preference a rule gets. In following example, we add a default deny rule which uses default priority 1 and a allow web traffic rule which is at higher priority 10.
```
mapuri: s / a  rule Id / a rule Id 

there is an extra space before `rule`

mapuri:
LGTM except for some minor comments.

Also while reviewing I got this idea (not sure if it too crazy). I think we should consider moving `contivctl` to `contiv/objmodel` project as we should be able to use this for volplugin intent as well.

More so it will be super useful if we can auto-generate the the cli sub-commands (i.e. `object` and `fields` keywords below) and respective logic from the model meta-data.

```
contivctl [create|delete|list] {object} [{field} value] <object-key(s)>
```

docs/contivctl.md
shaleman: sounds good

docs/contivctl.md
shaleman: thanks. Sounds good

docs/contivctl.md
shaleman: Thanks. I'll change the heading.

Yeah, i think i need to create a read me for contiv object model so that we can point it to people and refer it from this document

docs/contivctl.md
shaleman: Sure. Thx

docs/contivctl.md
```diff
@@ -0,0 +1,227 @@
+## contivctl
+
+contivctl command line tool lets you add/modify/delete objects into contiv object model using REST api.
+
+*Note:*
+`contivctl` Command can be found under `$GOPATH/bin` directory (`/opt/gopath/bin` in vagrant VMs) after a build. 
+Currently, `contivctl` needs to be run on the same host where `netmaster` is running
+
+## Usage
+
```
jainvipin: Would you think it makes sense to provide with 3 steps for someone to try out a policy between two containers?
I think it makes perfect sense to keep the usage, etc. but before any of that, we can let users try as they understand the concepts.

shaleman: Yeah, let me add a quick start section at the top that shows few simple steps to use the policies

mapuri: @shaleman @jainvipin  

I think if we do consider putting an example deployment configuration in this document then may be we should also consider renaming this file to some thing like `ConfiguringPolicies.md` or something? otherwise it might be hard for the users to relate the contents from the file-name.

BTW while looking at `docs` directory, I think `docs/ConfigDetails.md` is now obsolete, is that correct?

jainvipin: @mapuri - agree about renaming
docs/ConfigDetails.md is obsolete I think, we shouldn't encourage using that interface. Removing that and/or replacing a reference to contivctl would be helpful in main README.md

shaleman: @mapuri thats a good point.
@jainvipin 
This file is too detailed for an end user. Let me add a separate user facing document about contiv object model and how to use contivctl command. I'll structure it the way you mentioned above.
after that we can remove references to docs/ConfigDetails.md

PR #155, opened 2015-09-28
Update systemtests

erikh:
@mapuri this updates the system tests from systemtests-utils to use the new branch with the netplugin stuff removed. This branch is in a PR at https://github.com/contiv/systemtests-utils/pull/2.

The code contains no logic changes, just package name adjustments for all edits that don't change whole files.

I attempted to partition the commits, but I goofed up and forgot to remove the ovs.go file the first time around in the original systemtests-utils PR. If it's not a huge deal, rewriting it would kind of be a pain to do.

PTAL and let me know if adjustments need to be made.


systemtests/utils/utils.go
```diff
@@ -53,14 +54,14 @@ func StopOnError(testFailed bool) {
...
 // ConfigCleanupCommon performs common cleanup after each test
-func ConfigCleanupCommon(t *testing.T, nodes []TestbedNode) {
+func ConfigCleanupCommon(t *testing.T, nodes []stu.TestbedNode) {
 
 	if !OkToCleanup(t.Failed()) {
 		return
 	}
 
 	for _, node := range nodes {
-		cmdStr := "sh -c 'sudo $GOSRC/github.com/contiv/netplugin/scripts/cleanup'"
```
mapuri: I vaguely remember these changes were needed when we brought in  DIND tests. I think it is ok to revert these now since we don't run DIND sanity anymore, but just making a note of it for myself in case we need to revisit DIND sanity in future.

mapuri:
thanks @erikh 

The changes look good to me, we can merge once the sanities pass. 

I didn't completely get the comment regarding `ovs.go` file, but it looks fine to me. I think you want to squash the changes before merge.

systemtests/utils/utils.go
erikh: You know, I'm surprised that change is in there, must be an artifact of changes I did earlier that didn't get vendored in or something.

I can get rid of it if it is an issue.

erikh:
@mapuri I just added the ovs.go late by accident is all. I'll squash in a second.

PR #161, opened 2015-10-09
tcp flags changes

shaleman:
Add TCP flags to policy os that we can match on syn/ack flags
Set container link MTU to 1450 to account for vxlan overhead
Add a reverse proxy to access container web pages
minor fixes to contivctl and other python scripts

### DIFF
--- a/drivers/ovsSwitch.go
+++ b/drivers/ovsSwitch.go
@@ -30,7 +30,7 @@ import (
 )
 
 const useVethPair = true
-const vxlanVethMtu = 1450
+const vxlanEndpointMtu = 1450
 
 // OvsSwitch represents on OVS bridge instance
 type OvsSwitch struct {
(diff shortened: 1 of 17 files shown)


shaleman:
If there are no comments, i'd like to merge this PR.

drivers/ovsSwitch.go
```diff
@@ -223,6 +233,14 @@ func (sw *OvsSwitch) CreatePort(intfName string, cfgEp *OvsCfgEndpointState, pkt
 			log.Errorf("Error setting link %s up. Err: %v", ovsPortName, err)
 			return err
 		}
+
+		// Set the link mtu to 1450 to allow for 50 bytes vxlan encap
+		// (inner eth header(14) + outer IP(20) outer UDP(8) + vxlan header(8))
+		err = setLinkMtu(intfName, vxlanVethMtu)
```
mapuri: just wondering if we need to worry about the mtu of the host link itself as well?

Also should we set mtu even for internal ports?

netmaster/main.go
```diff
@@ -209,6 +212,9 @@ func (d *daemon) ListenAndServe() {
 		})
 	}
 
+	// proxy Handler
+	router.PathPrefix("/proxy/").HandlerFunc(proxyHandler)
```
mapuri: I am not very clear on the use case for having proxy embedded inside netmaster. 

Shouldn't it be provided as a separate service like kube-proxy in kubernetes cluster? This will allow us to use other proxies later like nginx, ha-proxy etc, right?

drivers/ovsSwitch.go
shaleman: Good points.
- Yes, we need to make container MTU = host interface MTU - 50. I'm not sure how to do this yet.
- I'll make the change to set internal port MTU too.

netmaster/main.go
shaleman: - This is a hack to access the overlay networks in vagrant environment. Thats how Vipin demo'd accessing redid/web application from the browser on his laptop.
- Yes, we'll create a separate service routing proxy like kube-proxy

mapuri: ok, so do we need this code in 'master' then?

PR #162, opened 2015-10-15
Update README.md

shaleman:
update GOPATH instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,9 @@ that lives behind an OVS bridge and has its own unique interfaces.
 #### Step 1: Clone the project:
 
 ```
+$ export GOPATH=`pwd`
+$ mkdir src/github.com/contiv
+$ cd src/github.com/contiv
 $ git clone https://github.com/contiv/netplugin
 $ cd netplugin; make build demo ssh
 ```


README.md
```diff
@@ -22,6 +22,9 @@ that lives behind an OVS bridge and has its own unique interfaces.
 #### Step 1: Clone the project:
 
 ```
+$ export GOPATH=`pwd`
```
mapuri: This is minor and feel free to ignore. Otherwise change LGTM.

Most of the users with go environment will have $GOPATH set and if not we can let standard go utilities to test and fail if needed than suggesting that user set their GOPATH to `pwd`.

I think we can have instructions as simple as:

```
go get -d https://github.com/contiv/netplugin
cd $GOPATH/src/github.com/contiv/netplugin
make build demo ssh
```

shaleman: I'm thinking being explicit is better for beginners who are not as familiar with golang..
Folks who are familiar with golang will know how to ignore the GOPATH part..

-sukhesh

mapuri: SGTM

jainvipin:
LGTM

PR #165, opened 2015-10-16
README: use mkdir -p to create directory

unclejack:
The `src/github.com` directories might not exist and we should tell the users to use `mkdir -p` to create the whole directory structure.


jainvipin:
LGTM

PR #168, opened 2015-10-21
Fixed golint error

golisai:
Signed-off-by: Sri Goli <srgoli@cisco.com>

### DIFF
--- a/utils/sysutils.go
+++ b/utils/sysutils.go
@@ -37,7 +37,7 @@ type SystemAttributes struct {
 	TotalNetBw  int
 }
 
-// Exported system attributes
+// SysAttrs are the exported system attributes
 var SysAttrs SystemAttributes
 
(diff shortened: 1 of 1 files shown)


utils/sysutils.go
```diff
@@ -37,7 +37,7 @@ type SystemAttributes struct {
 	TotalNetBw  int
 }
 
-// Exported system attributes
+// SysAttrs Exported system attributes
```
mapuri: @golisai thanks for the fix. Minor nit, may be we should reword this to:

```
SysAttrs are the exported system attributes
```

jainvipin:
LGTM

utils/sysutils.go
golisai: Sure. I will reword and resubmit.

mapuri:
LGTM

cc/ @contiv/core-network for final review and merge

PR #174, opened 2015-11-03
docs/CONTRIBUTING: rewrite & add commit guidelines

unclejack:


mapuri:
LGTM

docs/CONTRIBUTING.md
```diff
@@ -102,12 +133,23 @@ By making a contribution to this project, I certify that:
     this project or the open source license(s) involved.
 ```
 
-Then you just add a line to every git commit message:
+Every git commit message must have the following at the end on a separate line:
```
erikh: maybe instructions to `git commit -s` here?

docs/CONTRIBUTING.md
```diff
@@ -1,66 +1,97 @@
...
-are ripe to become a maintainer. If you think you are ready, drop an eamil
-to any of the maintainers.
+Play with the code and know it inside out. Once you think you are comfortable
+with the code and you think you are ready to become a maintainer, you can send
+an email to one of the maintainers.
+
+### Commit message format guidelines
+The commit message should have a short summary of no more than 50
+characters on the first line. The description should use verbs in the imperative
+(e.g. `netmaster: fix bug`, not `netmaster: fixed bug`).
```
jainvipin: Wonder why is this so critical... instead should we ask folks to specify the one liner to be succinct description of the fix itself. For example 'netmaster: fixed issue #114", or "netmaster: fix for race condition described in #114"

erikh:
Looks great other than that nit

docs/CONTRIBUTING.md
erikh: I've never been a big fan of the git first-line conventions, either, but I agree that subsections should be noted.

Perhaps we can do what we do in volplugin here? https://github.com/contiv/volplugin/blob/master/CONTRIBUTING.md#contributing-to-volplugin

docs/CONTRIBUTING.md
unclejack: @erikh: Good point, I'll add that.

docs/CONTRIBUTING.md
unclejack: @jainvipin @erikh Having commits which follow those rules makes it a lot easier when writing a change log, when doing bisects and when trying to make sense of what's going on. Some changes sometimes require some more details which are provided as a more detailed comment. Over time, having better commit messages helps.

The second line has to be empty in order to have the signature anyway, so this applies even when there's no longer description.

The main issue here is having the verbs written as `netmaster: fix issue #114` or `netmaster: fixes issue #114` and avoid past tenses because it describes exactly what the commit does.

@erikh: Thank you for the suggestion. I'll take a look and add it here.

unclejack:
The PR has been updated to make the following changes:
- explain that the detailed description of the change is optional and should be provided only when it's deemed necessary
- provide instructions on how to use `git commit -s` and how to set up the required git config options
- add another example of a commit with a sign off and without any detailed description

PTAL

erikh:
LGTM

PR #176, opened 2015-11-04
integrate with docker1.9 libnetwork

shaleman:
This change catches up to latest lib network

- Make dockplugin compatible with latest libnetwork
- Add IPAM driver
- Create docker network per EPG
- Make endpoint creation synchronous REST call from dockplugin to netmaster
- Move OvsCfg state to netmaster from ovs driver
- Misc code cleanup
- move EPG under network hierarchy


shaleman:
Folks,
I'm going to merge this in a hurry since service discovery integration and some aci work is waiting on this.
I'll take care of any comments in next commit.

PR #178, opened 2015-11-05
Makefile,contivctl: initial port of contivctl to go and codegangsta/cli

erikh:
**WORK IN PROGRESS** -- for evaluation purposes only

here's a port of `contivctl` to golang. it should work on all platforms and has no stdlib-external dependencies other than `Sirupsen/logrus` and `codegangsta/cli`.

I really need people to verify the functionality while I work on other parts of this. I'm going to make larger edits to refactor repetitive code out, but this works and will work largely the same after these other edits are finished.

/cc @shaleman @jainvipin

### DIFF
--- a/contivctl/commands.go
+++ b/contivctl/commands.go
@@ -1,0 +1,193 @@
+package contivctl
+
+import (
+	"github.com/codegangsta/cli"
+)
+
+var tenantFlag = cli.StringFlag{
+	Name:  "tenant, t",
+	Value: "default",
+	Usage: "Name of the tenant",
+}
+
+var allFlag = cli.BoolFlag{
+	Name:  "all, a",
+	Usage: "List all items",
+}
+
+var jsonFlag = cli.BoolFlag{
+	Name:  "json, j",
+	Usage: "Output list in JSON format",
+}
+
+// Commands are all the commands that go into `contivctl`, the end-user tool.
+// These are represented as cli.Command objects.
+var Commands = []cli.Command{
+	{
+		Name:  "group",
+		Usage: "Endpoint Group manipulation tools",
+		Subcommands: []cli.Command{
+			{
+				Name:      "create",
+				Usage:     "Create an endpoint group",
+				ArgsUsage: "[group] [network]",
+				Flags: []cli.Flag{
+					tenantFlag,
+					cli.StringFlag{
+						Name:  "policy, p",
+						Usage: "Policy List (separated by commas)",
+					},
+				},
+				Action: createEndpointGroup,
+			},
+			{
+				Name:      "delete",
+				Usage:     "Delete an endpoint group",
+				ArgsUsage: "[group]",
+				Flags:     []cli.Flag{tenantFlag},
+				Action:    deleteEndpointGroup,
+			},
+			{
+				Name:      "list",
+				Usage:     "List endpoint groups",
+				ArgsUsage: " ",
+				Flags:     []cli.Flag{tenantFlag, allFlag, jsonFlag},
+				Action:    listEndpointGroups,
+			},
+		},
+	},
+	{
+		Name:  "net",
+		Usage: "Network manipulation tools",
+		Subcommands: []cli.Command{
+			{
+				Name:      "list",
+				Usage:     "List networks",
+				ArgsUsage: " ",
+				Flags:     []cli.Flag{tenantFlag, allFlag, jsonFlag},
+				Action:    listNetworks,
+			},
+			{
+				Name:      "delete",
+				Usage:     "Delete a network",
+				ArgsUsage: "[network]",
+				Flags:     []cli.Flag{tenantFlag},
+				Action:    deleteNetwork,
+			},
+			{
+				Name:      "create",
+				Usage:     "Create a network",
+				ArgsUsage: "[network]",
+				Flags: []cli.Flag{
+					tenantFlag,
+					cli.BoolFlag{
+						Name:  "public, p",
+						Usage: "Indicate this network is public",
+					},
+					cli.StringFlag{
+						Name:  "encap, e",
+						Usage: "Packet tag",
+						Value: "vxlan",
+					},
+					cli.StringFlag{
+						Name:  "subnet, s",
+						Usage: "Subnet CIDR - REQUIRED",
+					},
+					cli.StringFlag{
+						Name:  "default-gw, g",
+						Usage: "Default Gateway - REQUIRED",
+					},
+				},
+				Action: createNetwork,
+			},
+		},
+	},
+	{
+		Name:  "rule",
+		Usage: "Rule manipulation tools",
+		Subcommands: []cli.Command{
+			{
+				Name:      "list",
+				Usage:     "List rules for a given tenant/policy",
+				ArgsUsage: "[policy]",
+				Flags:     []cli.Flag{tenantFlag, allFlag, jsonFlag},
+				Action:    listRules,
+			},
+			{
+				Name:      "delete",
+				Usage:     "Delete an existing rule.",
+				ArgsUsage: "[policy] [rule id]",
+				Flags:     []cli.Flag{tenantFlag},
+				Action:    deleteRule,
+			},
+			{
+				Name:      "add",
+				Usage:     "Add a new rule.",
+				ArgsUsage: "[policy] [rule id]",
+				Flags: []cli.Flag{
+					tenantFlag,
+					cli.IntFlag{
+						Name:  "priority, p",
+						Usage: "Priority Indicator",
+					},
+					cli.StringFlag{
+						Name:  "direction, d",
+						Usage: "Direction of traffic (in/out)",
+					},
+					cli.StringFlag{
+						Name:  "group, g",
+						Usage: "Endpoint Group Name",
+					},
+					cli.StringFlag{
+						Name:  "network, n",
+						Usage: "Network name",
+					},
+					cli.StringFlag{
+						Name:  "ip-address, a",
+						Usage: "IP Address",
+					},
+					cli.StringFlag{
+						Name:  "protocol, l",
+						Usage: "Protocol (e.g., tcp)",
+					},
+					cli.IntFlag{
+						Name:  "port, P",
+						Usage: "Port",
+					},
+					cli.StringFlag{
+						Name:  "action, j",
+						Usage: "Action to take (e.g., deny)",
+					},
+				},
+				Action: addRule,
+			},
+		},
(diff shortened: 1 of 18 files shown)


contivctl/contivctl/main.go
```diff
@@ -0,0 +1,195 @@
...
+					Action: contivctl.ListRules,
+				},
+				{
+					Name:   "delete",
+					Usage:  "Delete an existing rule.",
+					Flags:  []cli.Flag{tenantFlag},
+					Action: contivctl.DeleteRule,
+				},
+				{
+					Name:  "add",
```
mapuri: minor nit, should this sub-command be called `create` as well just like sub-commands for policy, network etc?

mapuri:
looks good @erikh 

Just wanted to share the thought I had posted on the initial python based implementation of [contivctl here](https://github.com/contiv/netplugin/pull/153#issuecomment-143833431), do you think we should explore auto-generating these clis from the objmodel that will help keeping the UX consistent across objects and also save us a lot of hand written code?

shaleman:
Erik,

This is a good start.
Most things seem to work.

Few things I noticed while playing with it:


  1.  "Group create" doesn't seem to work. It keeps saying invalid argument
     *   $ contivctl group create test --policy=first

ERRO[0000] Invalid Arguments

NAME:

   group create - Create an endpoint group


USAGE:

   command group create [command options] [arguments...]


OPTIONS:

   --tenant, -t "default" Name of the tenant to create the policy on.

   --policy, -p Policy List (separated by commas)

  2.  Help screen for positional arguments(like group name) is not shown. See below for an example.
     *   $ contivctl group create --help

NAME:

   group create - Create an endpoint group


USAGE:

   command group create [command options] [arguments...]


OPTIONS:

   --tenant, -t "default" Name of the tenant to create the policy on.

   --policy, -p Policy List (separated by commas)

  3.  Empty values are shown as <nil> instead of empty string or "-" string
     *

$ contivctl rule list first

Rule  Direction  Priority  EndpointGroup  Network  IpAddress  Protocol  Port   Action

----  ---------  --------  -------------  -------  ---------  --------  ----   ------

1     in         1         <nil>          <nil>    <nil>      tcp       <nil>  deny

  4.  Some commands show duplicate entries like:
     *

$ contivctl rule list first

Rule  Direction  Priority  EndpointGroup  Network  IpAddress  Protocol  Port   Action

----  ---------  --------  -------------  -------  ---------  --------  ----   ------

1     in         1         <nil>          <nil>    <nil>      tcp       <nil>  deny

2     in         1         <nil>          <nil>    <nil>      tcp       80     accept

1     in         1         <nil>          <nil>    <nil>      tcp       <nil>  deny

2     in         1         <nil>          <nil>    <nil>      tcp       80     accept

     *   Note that rule-1 and rule-2 are shown twice

-sukhesh

shaleman:
group creation requires a network name as well as we discussed prior, I 
presumed this to be mandatory but I can make an option if not.

the argument help is missing on all the commands, I will rectify this 
later tonight or tomorrow.

I'm looking into the rule listing code now, can you show me what you did 
to create them?

-Erik

contivctl/contivctl/main.go
erikh: up to @shaleman; I was just copying from the python.

erikh:
I'm on a short timeline here, so I can't explore that idea this 
iteration. I also think the objmodel will need a lot of work to make 
this happen.

erikh:
Hey, I fixed the arguments usage and that's reflected on the latest 
version.

shaleman:
@erikh changes look good to me. If you can fix the sanity failure, we can quickly merge this PR

shaleman:
@mapuri 
+1 on auto generation. I can take this up after dockercon.

mapuri:
@shaleman :+1:

contivctl/commands.go
```diff
@@ -0,0 +1,191 @@
...
+
+var tenantFlag = cli.StringFlag{
+	Name:  "tenant, t",
+	Value: "default",
+	Usage: "Name of the tenant to create the policy on.",
+}
+
+var Commands = []cli.Command{
+	{
+		Name:  "group",
```
jainvipin: sorry for nitpicking: should we call 'group' or 'policy-domain' or 'endpoint group' because containers can belong to two different groups at a given time.

contivctl/commands.go
```diff
@@ -0,0 +1,191 @@
...
+		Usage: "Endpoint Group manipulation tools",
+		Subcommands: []cli.Command{
+			{
+				Name:      "create",
```
jainvipin: @erikh, @unclejack @mapur: do you see the first verb being 'object' and verbs/actions as second subcommand a right model given other things in container ecosystem. I'd think so... but wanted to confirm

contivctl/commands.go
```diff
@@ -0,0 +1,191 @@
...
+				Flags:     []cli.Flag{tenantFlag},
+				Action:    DeleteEndpointGroup,
+			},
+			{
+				Name:      "list",
+				Usage:     "List endpoint groups",
+				ArgsUsage: " ",
+				Flags: append([]cli.Flag{tenantFlag},
+					cli.BoolFlag{
+						Name:  "all, a",
```
jainvipin: If we allow 'all' - does it mean we can specify a specific element if '-a' flag is not specified? Or is that a default flag?

contivctl/commands.go
```diff
@@ -0,0 +1,191 @@
...
+					cli.IntFlag{
+						Name:  "priority, p",
+						Usage: "Priority Indicator",
+					},
+					cli.StringFlag{
+						Name:  "direction, d",
+						Usage: "Direction of traffic (in/out)",
+					},
+					cli.StringFlag{
+						Name:  "epg, e",
```
jainvipin: here we use 'epg', but 'group' earlier... need to be consistent

erikh:
I really don't think binding to the object model is a good idea. We're going to have a UI that is ever changing, inflexible with regard to user help for error conditions (which is when people are confused the most), and most importantly, impossible to document.

Please reconsider this.

contivctl/commands.go
```diff
@@ -0,0 +1,191 @@
...
+				Usage:     "Add a new rule.",
+				ArgsUsage: "[policy] [rule id]",
+				Flags: []cli.Flag{
+					tenantFlag,
+					cli.IntFlag{
+						Name:  "priority, p",
+						Usage: "Priority Indicator",
+					},
+					cli.StringFlag{
+						Name:  "direction, d",
```
jainvipin: Is there a way to specify the  values of direction 'in' and 'out' here? instead of doing inline validation i.e. enums in codegangsta/cli

erikh: This is only safely handled in the master daemon code, not here. We shouldn't try to filter it here, instead preferring the master's response back to the client.

curl gets right around this, so if we don't want to panic the daemon we should ensure all our validation happens there.

contivctl/contivctl.go
```diff
@@ -0,0 +1,287 @@
...
+	argCheck(1, ctx)
+
+	tenant := ctx.String("tenant")
+	policy := ctx.Args()[0]
+
+	rules := getList(ctx, ruleURL(ctx))
+
+	writer := tabwriter.NewWriter(os.Stdout, 0, 2, 2, ' ', 0)
+	defer writer.Flush()
+	writer.Write([]byte("Rule\tDirection\tPriority\tEndpointGroup\tNetwork\tIpAddress\tProtocol\tPort\tAction\n"))
```
jainvipin: Should we display the output in 'json' like output for example how 'docker inspect' gives the information? I understand that 'docker ps' and 'docker inspect' shows different formats, but 
inspect allows a lot more hierarchical output. I wonder if we need 'inspect' version of a network i.e. epgs and policy details within network, etc.

erikh: I'll implement this now.

contivctl/commands.go
erikh: yes.

jainvipin:
@erikh - great start... will fill in things as we go, but definitely a unified complete picture of various objects

should we do bash autocompletion for some of our commands? I was playing with autocompletion support in codegangsta and it looks quite promising.
Most of my comments do not fall in 'must have' for this checkin but something that we must address...

mapuri:
@erikh 

> I really don't think binding to the object model is a good idea. We're going to have a UI that is ever changing, inflexible with regard to user help for error conditions (which is when people are confused the most), and most importantly, impossible to document.

I might be completely missing your concern. But I think auto-generation of UX (and if possible docs) from object model should addresses the inflexibility and documentation concerns to a large extent.

I ask for auto generation from object model as imo that should be defining our UX (REST, Web, CLI etc). So yes, if object definition changes then UX shall change accordingly. And I feel that actually takes away the chances of confusion that might arise if we have to to manual updates to UX and docs every time an object changes.

Having said that I am in now way underestimating the amount of work it will take to achieve this. And in the beginning we might actually breaking things by not being backward compatible etc, when we change object definitions but I think once we are able to automate integration of objmodel in all our projects it will actually prove pretty useful when we decide to introduce new constructs and interfaces in future.

contivctl/commands.go
mapuri: I am fine either way as long as we consistent with the `verbs` and their placement across all our clis.

erikh:
ok, I think this is ready for merge; just incorporated the JSON changes. PTAL and **test** if you have time.

erikh:
@mapuri respectfully to all involved, I don't think objmodel is in the shape (and it needs a lot of work, a lot lot lot) to pepper everywhere in the code right now. Currently there's one member of our team who knows how to even import it into netplugin, and that takes several days to get through the pipe.

The code itself, (which is at least partially my fault) is not flexible, either. The principal argument for the object model so far has been handling responses from foreign clients but there are none yet other than the web interface, which arguably does not need this to function since there is no formal client for it right now.

I would really like to get back to static structs with reflection / struct tags and use them to generate alternative clients if we actually realize that in the future. Right now the typeless nature of the object model means that I must use `map[string]interface{}{}` everywhere just to keep things afloat and check for nils everywhere -- this is not how an object model should work, considering it's supposed to be typed.

But ultimately, my code is not better or more flexible for using it.

Sorry if this post seems inflammatory, I just want to be sure we're making the right decision.

contivctl/contivctl.go
erikh: This is implemented now, PTAL

erikh:
I think we should, but not in this patch. It will be a little too 
complicated right now.

PR #180, opened 2015-11-06
fix network delete; fix EPG keys; code cleanup

shaleman:
- fix network delete
- fix EPG keys
- code cleanup


jojimt:
I see a bunch of deletes related to volumes. Hope those are intentional/ok.
The rest LGTM.

shaleman:
yes, cleaning up code thats not used by netmaster

PR #181, opened 2015-11-06
Support app delete and exposed ports

jojimt:
Changes to support app delete and propagate exposed ports during create.


shaleman:
LGTM

PR #184, opened 2015-11-10
libcompose + service discovery integration

shaleman:
- publish DNS server address for each network so that deploy can query it
- minor bug fixes
- start/stop script changes


netmaster/master/netmaster.go
```diff
@@ -138,11 +142,15 @@ func startServiceContainer(tenantName string) error {
 		return err
 	}
 
+	// pull the skydns image if it does not exist
 	imageName := "skynetservices/skydns:latest"
```
jainvipin: should this be hardcoded?

jainvipin:
minor nit; LGTM otherwise

PR #185, opened 2015-11-10
Vagrantfile: Use a better strategy for getting buidls off VMs than ov…

erikh:
…erwriting gopath

this should be merged before we go to demo, PTAL


erikh:
closed in favor of #186, which incorporates this change.

PR #186, opened 2015-11-10
Netctl fixes

erikh:
see commits. Found these while testing with a demo script.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -118,7 +118,13 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
             end
 
             # mount the host directories
-            node.vm.synced_folder "../../../../", gopath_folder
+            node.vm.synced_folder "bin", File.join(gopath_folder, "bin")
+            if ENV["GOPATH"]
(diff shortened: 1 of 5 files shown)


Vagrantfile
```diff
@@ -118,7 +118,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
             end
 
             # mount the host directories
-            node.vm.synced_folder "../../../../", gopath_folder
+            node.vm.synced_folder "bin", File.join(gopath_folder, "bin")
+            node.vm.synced_folder ".", File.join(gopath_folder, "src/github.com/contiv/netplugin")
```
shaleman: @erikh most of us work on multiple Repos like deploy/libcompose/netplugin/objmodel/ofnet in same workspace.
So, we'd need entire source tree available inside the VM for builds.
Can we mount synced_folder $GOPATH/src as /opt/gopath/src inside VM?
this way, binaries don't conflict between Mac and Linux and we can still have entire source tree inside the VM

erikh: Sure, seems good. Just not the bins or pkg.

erikh:
@shaleman I incorporated some changes around your suggestions. I also managed to keep the existing behavior for those who are not go devs and will not have a GOPATH configured. I hope this helps!

shaleman:
LGTM

PR #187, opened 2015-11-11
trim default tenant; fix policy bug

shaleman:
- Trim default tenant name from docker network names
- Fix a bug in policy code
- Add mutexes around endpoint and address allocation
- Ugly hack to setup proxy interface

### DIFF
--- a/mgmtfn/dockplugin/netDriver.go
+++ b/mgmtfn/dockplugin/netDriver.go
@@ -18,6 +18,7 @@ package dockplugin
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"strings"
@@ -32,6 +33,8 @@ import (
 	"github.com/docker/libnetwork/drivers/remote/api"
 	"github.com/samalba/dockerclient"
 )
+
+const defaultTenantName = "default"
 
 func getCapability() func(http.ResponseWriter, *http.Request) {
 	return func(w http.ResponseWriter, r *http.Request) {
@@ -400,7 +403,6 @@ func netdGetNetwork(networkID string) (*mastercfg.CfgNetworkState, error) {
 	nwCfg.StateDriver = stateDriver
 	err = nwCfg.Read(networkID)
 	if err != nil {
-		log.Errorf("network %s is not operational", networkID)
 		return nil, err
 	}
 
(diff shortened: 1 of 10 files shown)


mgmtfn/dockplugin/netDriver.go
```diff
@@ -433,14 +435,27 @@ func GetDockerNetworkName(nwID string) (string, string, string, error) {
 			var tenantName, netName, serviceName string
 			names := strings.Split(nw.Name, ".")
 			if len(names) == 2 {
-				// has only network.tenant format
-				tenantName = names[1]
-				netName = names[0]
+				// determine if this is service.network on default tenant or network.tenant
```
jainvipin: should we use 'service.network/tenant' to avoid this confusion, then 'service.network' is valid and 'network/tenant' is valid and 'service.network/tenant' is valid too. This is assuming that '/' is allowed in the list as well.

mgmtfn/dockplugin/netDriver.go
```diff
@@ -433,14 +435,27 @@ func GetDockerNetworkName(nwID string) (string, string, string, error) {
...
+				}
 			} else if len(names) == 3 {
 				// has service.network.tenant format
 				tenantName = names[2]
 				netName = names[1]
 				serviceName = names[0]
+			} else if len(names) == 1 {
+				// If only network is specified, use default tenant
+				tenantName = defaultTenantName
+				netName = names[0]
```
jainvipin: need system tests for 'service.network/tenant', 'service.network', 'network', 'network/tenant' independently.

netmaster/master/api.go
```diff
@@ -88,6 +92,10 @@ func AllocAddressHandler(w http.ResponseWriter, r *http.Request, vars map[string
 
 	log.Infof("Received AddressAllocRequest: %+v", allocReq)
 
+	// Take a global lock for address allocation
+	addrMutex.Lock()
+	defer addrMutex.Unlock()
```
jainvipin: just curious: for my understanding the locks are needed for various objects because multiple netmasters are allowed on the same cluster?

scripts/python/setupProxy.py
```diff
@@ -68,7 +67,7 @@ def setupProxy():
 
             # Look for pro
             for ep in epList:
-            	if ep['id'] == "private-myContainer1":
+            	if ep['id'] == "private.default-proxyPort":
```
jainvipin: does proxy work with these changes?

mgmtfn/dockplugin/netDriver.go
shaleman: I haven't tried '/' character. We could try

mgmtfn/dockplugin/netDriver.go
shaleman: yes, i'll add system test for these

netmaster/master/api.go
shaleman: No, its for multiple REST calls coming into same net master at the same time. This was seen while testing "docker-compose scale" command. docker-compose multiple containers in parallel and some of them end up getting duplicate addresses.

scripts/python/setupProxy.py
shaleman: Yes, proxy works now.

PR #191, opened 2015-11-12
Minor fixes to netctl

shaleman:
- move EPG under network hierarchy
- eliminate 'public' field in network command
- add packet tag field to specify clan-id/vni
- rename default-gw to gateway
- minor fix to json parsing

Appreciate a quick review since "create net" wouldn't work without this change.

### DIFF
--- a/netctl/commands.go
+++ b/netctl/commands.go
@@ -38,7 +38,7 @@ var Commands = []cli.Command{
 			{
 				Name:      "create",
 				Usage:     "Create an endpoint group",
-				ArgsUsage: "[group] [network]",
+				ArgsUsage: "[network] [group]",
 				Flags: []cli.Flag{
 					tenantFlag,
 					cli.StringFlag{
@@ -51,7 +51,7 @@ var Commands = []cli.Command{
 			{
 				Name:      "delete",
 				Usage:     "Delete an endpoint group",
-				ArgsUsage: "[group]",
+				ArgsUsage: "[network] [group]",
 				Flags:     []cli.Flag{tenantFlag},
 				Action:    deleteEndpointGroup,
 			},
@@ -88,22 +88,22 @@ var Commands = []cli.Command{
 				ArgsUsage: "[network]",
 				Flags: []cli.Flag{
 					tenantFlag,
-					cli.BoolFlag{
-						Name:  "public, p",
-						Usage: "Indicate this network is public",
-					},
 					cli.StringFlag{
 						Name:  "encap, e",
-						Usage: "Packet tag",
+						Usage: "Encap type (vlan or vxlan)",
 						Value: "vxlan",
+					},
+					cli.StringFlag{
+						Name:  "pkt-tag, p",
+						Usage: "Packet tag (Vlan/Vxlan ids)- REQUIRED",
 					},
 					cli.StringFlag{
 						Name:  "subnet, s",
 						Usage: "Subnet CIDR - REQUIRED",
 					},
 					cli.StringFlag{
-						Name:  "default-gw, g",
-						Usage: "Default Gateway - REQUIRED",
+						Name:  "gateway, g",
+						Usage: "Gateway - REQUIRED",
 					},
 				},
 				Action: createNetwork,
--- a/netctl/netctl.go
+++ b/netctl/netctl.go
@@ -51,8 +51,6 @@ func listPolicies(ctx *cli.Context) {
 	policies := getList(ctx, policyURL(ctx))
 	filtered := []map[string]interface{}{}
 
-	filtered = policies
-
 	if !ctx.Bool("all") {
 		for _, policy := range policies {
 			if policy["tenantName"] == tenant {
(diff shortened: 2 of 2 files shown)


netctl/netctl.go
```diff
@@ -237,23 +235,18 @@ func listNetworks(ctx *cli.Context) {
 	} else {
 		writer := tabwriter.NewWriter(os.Stdout, 0, 2, 2, ' ', 0)
 		defer writer.Flush()
-		writer.Write([]byte("Tenant\tNetwork\tPublic\tEncap\tSubnet\tGateway\n"))
-		writer.Write([]byte("------\t-------\t------\t-----\t------\t-------\n"))
+		writer.Write([]byte("Tenant\tNetwork\tEncap\tTag\tSubnet\tGateway\n"))
```
mapuri: Just for my understanding, how is `tag` different from `encap`?

mapuri:
LGTM

netctl/netctl.go
shaleman: Encap is vlan/vxlan. Tag is vlan-id or vni
suggestion for better name are welcome

mapuri: ah, ok..

May be 'encap-type' and 'encap-value' might be more clear?

shaleman: Here is a sample output. I think its reasonable.
```
$ netctl net list
Tenant   Network     Encap  Tag     Subnet  Gateway
------   -------     -----  ------  -------
default  private     vxlan  1001    10.1.0.0/16     10.1.254.254
default  public      vlan   1       192.168.1.0/24  192.168.1.254
default  contiv-net  vxlan  1005    20.1.1.0/24     20.1.1.1
```

netctl/commands.go
```diff
@@ -88,22 +88,22 @@ var Commands = []cli.Command{
...
-						Usage: "Indicate this network is public",
-					},
 					cli.StringFlag{
 						Name:  "encap, e",
-						Usage: "Packet tag",
+						Usage: "Packet encap",
 						Value: "vxlan",
 					},
 					cli.StringFlag{
+						Name:  "pktTag, p",
```
erikh: can we keep consistent case with our `--` arguments? this is the only one that has a capital letter in it.

netctl/netctl.go
mapuri: sgtm

shaleman: @mapuri our comments crossed. I'm ok with "Encap Type" and "Packet Tag" also. encap-type encamp-value might be too geeky :)

netctl/netctl.go
```diff
@@ -317,7 +311,11 @@ func listEndpointGroups(ctx *cli.Context) {
 		for _, group := range filtered {
 			policies := ""
 			if group["policies"] != nil {
-				policies = strings.Join(group["policies"].([]string), ",")
+				policyList := []string{}
+				for _, p := range group["policies"].([]interface{}) {
```
erikh: was this broken before? `[]interface{}` usually leads to problems.

netctl/commands.go
shaleman: Hmm. i'll change it to --pkt-tag

netctl/netctl.go
shaleman: yes, it was crashing with this message:
$ netctl group list 
Tenant  Group  Network  Policies
------  -----  -------  --------
panic: interface conversion: interface is []interface {}, not []string

goroutine 1 [running]:
github.com/contiv/netplugin/netctl.listEndpointGroups(0xc82008aea0)
	/opt/gopath/src/github.com/contiv/netplugin/netctl/netctl.go:318 +0x644
github.com/codegangsta/cli.Command.Run(0x7bb4c0, 0x4, 0x0, 0x0, 0x0, 0x0, 0x0, 0x7f2bd0, 0x14, 0x0, ...)
	/opt/gopath/src/github.com/contiv/netplugin/Godeps/_workspace/src/github.com/codegangsta/cli/command.go:131 +0x1055

erikh: ah ok.

erikh:
LGTM after `--pkg-tag` change

netctl/netctl.go
mapuri: +1.
I would like the headings to be clear even when table had no values. "Encap Type" and "Packet Tag" are better, as they are also more explicit about what to expect in those columns.

Just as side, do we need same for the corresponding cli flags as well? Right now they are `encap` and `pkt-tag` but may be they should also be `encap-type` and `pkt-tag`. But upto you.

shaleman: I made it look like this below.
Note that description match. but, arg option is in abbreviated form

```
vagrant@netplugin-node1:~$ netctl net create --help
NAME:
   netctl net create - Create a network

USAGE:
   netctl net create [command options] [network]

OPTIONS:
   --tenant, -t "default"	Name of the tenant
   --encap, -e "vxlan"		Encap type (vlan or vxlan)
   --pkt-tag, -p 		Packet tag (Vlan/Vxlan ids)- REQUIRED
   --subnet, -s 		Subnet CIDR - REQUIRED
   --gateway, -g 		Gateway - REQUIRED
   
vagrant@netplugin-node1:~$ netctl net list
Tenant   Network     Encap type  Packet tag  Subnet          Gateway
------   -------     ----------  ----------  -------         ------
default  private     vxlan       1001        10.1.0.0/16     10.1.254.254
default  public      vlan        1           192.168.1.0/24  192.168.1.254
default  contiv-net  vxlan       1005        20.1.1.0/24     20.1.1.1
```

mapuri: :+1: lgtm

PR #195, opened 2015-11-18
Svcplugin cleanup

DivyaVavili:
Changes done:
- Added support to cleanup skydns state during network delete and tenant delete
- Removed code dependency on go-dockerclient 
- Added check to create public/private networks for only default tenant
- Added support to retry pulling skydns image if pull-image failed


netmaster/master/netmaster.go
```diff
@@ -175,6 +185,30 @@ func startServiceContainer(tenantName string) error {
 	return err
 }
 
+func stopServiceContainer(tenantName string) error {
```
mapuri: minor nit, may be this should be called `stopAndRemoveServiceContainer` to indicate the fact that it also destroys the service container.

netmaster/master/network.go
```diff
@@ -163,6 +164,57 @@ func deleteDockNet(networkName string) error {
 	return nil
 }
 
+func detachServiceContainer(networkName, tenantName string) error {
```
mapuri: Sorry if this is orthogonal to this PR and may be this discussion need to happen somewhere else, so kindly correct me. But I think by wrapping up service-discovery container inside netmaster we are tying state to a specific netmaster instance. For example, how will service-discovery work if this particular netmaster instance were to get killed and another  netmaster instance now needs to take over?

netmaster/master/network.go
```diff
@@ -421,6 +473,13 @@ func DeleteNetworkID(stateDriver core.StateDriver, netID string) error {
 		return err
 	}
 
+	err = detachServiceContainer(netID, nwCfg.Tenant)
```
mapuri: Should this code block be before line 468 above? Otherwise if reading tenant fails for some reason then we won't be able to do this cleanup as well.

netmaster/objApi/apiController.go
```diff
@@ -737,38 +737,40 @@ func (ac *APIController) TenantCreate(tenant *contivModel.Tenant) error {
...
-		Subnet:      "10.1.0.0/16",
-		Gateway:     "10.1.254.254",
-		NetworkName: "private",
-		TenantName:  tenant.TenantName,
-	})
-	if err != nil {
-		log.Errorf("Error creating privateNet for tenant: %+v. Err: %v", tenant, err)
-		return err
-	}
+	if tenant.TenantName == "default" {
```
mapuri: just for my understanding, what is the reason to create a private network just for 'default' tenant? Should we follow the same for public network as well below?

svcplugin/consulextension/consul.go
```diff
@@ -19,11 +18,13 @@ func init() {
 	bridge.Register(new(Factory), "consul")
 }
 
+/* TODO: Fix this for check_script
```
mapuri: I suggest that we remove this commented out code. It doesn't look good in an open sourced repo. If we need to fix a script to make it work then may be also fix the script as part of this PR?

netmaster/objApi/apiController.go
shaleman: @mapuri: Just to give you some context, I'm working on removing auto creation of these networks completely(including from default tenant). I asked Divya to make a quick change to create these networks only for default tenant for now. She was running into problems with testing multiple tenants because of overlapping Vlan/Vxlan Ids for these auto created networks.

mapuri: oh, k, sgtm. thanks @shaleman

Is this supposed to be checked in then, if it was just needed for @DivyaVavili testing? Either way is fine with me as you also seem to be working on changing this logic.

netmaster/master/network.go
shaleman: This is something we need to think about as we'll have more such co-processing containers like ACI-gateway and skydns. Current thinking for skydns is to let schedulers(E.g. Swarm) run it anywhere in the cluster where resources are available. ACI-gateway might have slightly different requirements around locality with active-netmaster.

mapuri: yep, agree we need to think this through.

> Current thinking for skydns is to let schedulers(E.g. Swarm) run it anywhere in the cluster where resources are available

ok, I think for now we should use tcp socket here instead to be able to make use of swarm.

> ACI-gateway might have slightly different requirements around locality with active-netmaster.

hmm, do you have specific requirement in mind? Regarding `locality with active-netmaster` could we make use of REST endpoints (behind a VIP) to solve that?

DivyaVavili:
Will reopen the Pull request after including some additional fixes

PR #197, opened 2015-11-20
Provide a link to documentation page

shaleman:
add a link to docs.contiv.io


mapuri:
LGTM... 

unlrelated, but I just noticed that github specific links in `http://docs.contiv.io/` are not working  for both netplugin and volplugin. May be something broke recently.

PR #201, opened 2015-12-03
svcplugin cleanup

DivyaVavili:
- Added support during delete network and delete tenant
- Removed unused code/features
- Using docker swarm socket instead of default unix socket
- Incorporated previous PR comments
- Including netctl changes for tenant manipulation

netctl tenant tools usage:
vagrant@netplugin-node1:~$ netctl tenant create -p 100.1.0.0/16 -v 2000-2100 -x 12000-12100 tenant1
INFO[0000] Creating tenant: tenant1
vagrant@netplugin-node1:~$ netctl tenant list
Name Subnet Pool Subnet Len vlans vxlans

default 10.1.1.1/16 24 100-1100 1001-1100
tenant1 100.1.0.0/16 24 2000-2100 12000-12100
vagrant@netplugin-node1:~$ netctl tenant delete tenant1
INFO[0000] Deleting tenant tenant1
vagrant@netplugin-node1:~$ netctl tenant list
Name Subnet Pool Subnet Len vlans vxlans

default 10.1.1.1/16 24 100-1100 1001-1100

Arguments help:
vagrant@netplugin-node1:~$ netctl tenant create -h
NAME:
netctl tenant create - Create a tenant

USAGE:
netctl tenant create [command options] [tenant]

OPTIONS:
--subnet-pool, -p Subnet CIDR - REQUIRED
--subnet-len, -l "24" Subnet length
--vlans, -v Vlan range - REQUIRED
--vxlans, -x Vxlan range - REQUIRED

vagrant@netplugin-node1:~$ netctl tenant list -h
NAME:
netctl tenant list - List tenants

USAGE:
netctl tenant list

vagrant@netplugin-node1:~$ netctl tenant delete -h
NAME:
netctl tenant delete - Delete a tenant

USAGE:
netctl tenant delete [tenant]


erikh:
LGTM aside from comments

netctl/netctl.go
```diff
@@ -252,6 +252,71 @@ func listNetworks(ctx *cli.Context) {
...
+func createTenant(ctx *cli.Context) {
+	argCheck(1, ctx)
+
+	tenant := ctx.Args()[0]
+	subnetPool := ctx.String("subnet-pool")
+	subnetLen := ctx.Int("subnet-len")
+	vlans := ctx.String("vlans")
+	vxlans := ctx.String("vxlans")
+
+	if subnetPool == "" || vlans == "" || vxlans == "" {
```
mapuri: do we need a check for `subnet-len`?

On a related note, may be these validations should be be left to objmodel and it's client interface? May be it is already planned as we are moving towards a shared objmodel with code generation. Just want to confirm my understanding.

netctl/netctl.go
```diff
@@ -252,6 +252,71 @@ func listNetworks(ctx *cli.Context) {
...
+	deleteURL(ctx, url)
+}
+
+func listTenants(ctx *cli.Context) {
+	argCheck(0, ctx)
+
+	list := getList(ctx, tenantURL(ctx))
+
+	if ctx.Bool("json") {
+		dumpList(ctx, list)
```
mapuri: minor, if we add a `return` in the `if` block, then we can get rid of the `else` block and so the indentation.

netmaster/master/network.go
```diff
@@ -163,6 +164,57 @@ func deleteDockNet(networkName string) error {
...
+	nwCfg := &mastercfg.CfgNetworkState{}
+	nwCfg.StateDriver = stateDriver
+	err = nwCfg.Read(networkName)
+	if err != nil {
+		return err
+	}
+
+	log.Infof("dnsServerIP: %+v, nwCfg.dnsip: %+v", dnsServerIP, nwCfg.DNSServer)
+	// Remove dns container from network if all other endpoints are withdrawn
+	if len(nwState.Containers) == 1 && (dnsServerIP == nwCfg.DNSServer) {
```
mapuri: minor nit, can we call this function `checkAndDetachServiceContainer()`? Only after reading the code till this condition I was able to figure out that this function conditionally detaches a container rather than always, may be it's just me :)

scripts/cleanup
```diff
@@ -34,6 +34,7 @@ killall netmaster
 etcdctl rm -recursive /contiv > /dev/null 2>&1
 etcdctl rm -recursive /contiv.io > /dev/null 2>&1
 curl -X DELETE http://localhost:8500/v1/kv/contiv?recurse > /dev/null 2>&1
+curl -X DELETE http://localhost:8500/v1/kv/contiv.io?recurse > /dev/null 2>&1
```
mapuri: do you also want to cleanup the skydns related keys here, like what you are doing in `testbedAPI.py`? It would be nice if we could use common cleanup script.

mapuri:
LGTM other than some minor comments.

netctl/netctl.go
DivyaVavili: I think that this should be handled as part of objmodel as well, else the check has to be repeated across all netctl CLI (like create network which uses subnet as well). From what I remember @shaleman was mentioning he was working on sanitizing some of these like making sure the subnets don't overlap across tenants. I'm guessing this will be taken care as part of that.

scripts/cleanup
DivyaVavili: I didn't realize there were two scripts here... This was already handled in scripts/python/testbedApi.py... I will add the fix here as well

netmaster/master/network.go
DivyaVavili: I'm working on more changes to handle attaching/detaching service containers in a more elegant way. This will probably change as part of that. Will make the change accordingly in the next PR.

DivyaVavili:
Looks like the last test case failed because of docker container launch issue...
Is there a way to retrigger the sanity checks without an explicit push

PR #203, opened 2015-12-05
code cleanup and improve sanity tests

shaleman:
- Deprecate unused code
  - dockcontivnet
  - k8scontivnet
  - netdcli
  - crt & crtclient & crtclient/docker
  - examples
  - old system test based on netdcli/old intent

- Add new sanity test infrastructure
- fix bugs found during new sanity tests


erikh:
Why can't we just delete the files? We have source control, let's use it right.

erikh:
NOT LGTM until we get a lot of things resolved here:

* delete the files in `.deprecated` directory, but fix your commits so it doesn't exist so you just delete the files
* the system tests we have a whole framework for. Please use it instead of writing a new one. Write them in go so they can be maintained by the whole team, please.

The rest I will review after this is resolved. This is not in worthy shape to get reviewed.

jainvipin:
The work is commendable, esp random testing, etc., and for that reason this needs to be leveraged across multiple repos, e.g. volplugin
> Add new sanity test infrastructure

jainvipin:
Is there a new README.md, I see the original one purged.

shaleman:
@erikh I'm glad you brought up test framework question.

Using new python based framework is a conscious decision, to use right tool for the job.
Here is my thinking:
- go test is great. We should use go test whenever we are testing go APIs. i.e unit testing individual modules. We should add more unit tests for modules lacking unit tests.
- system tests perform end-to-end tests by emulating user behavior. This mostly involves executing some shell commands to perform a trigger and executing other shell commands to verify things are working as expected across the cluster. I don't think go is the right language for this job. python is more suitable for this job. Python is also used more commonly in automation test environments. So, its easier to find libraries, frameworks and people with python testing experience.

I looked at other projects which have similar system test requirements. Libnetwork and Swarm need to do similar tests of executing trigger commands and verifying results. They both use bats framework based on bash shell scripting. I think python is a better choice than bash/bats.

Here are the requirements for the test framework in my mind:
 1. We should be able to use same testing framework and testcases in Vagrant environment, bare metal server environments and in our scale testbed environments. We should not tie the framework to setting up vagrant environment.
 2. We should be able to use the same test framework in integration tests with physical switches to test (a) ACI integration (b) BGP integration (c) EVPN integration. These tests involve configuring external switches and APIC controllers as well as configuring our plugin.
 3. We should be able to run system tests in both Mac laptops and Linux servers. i.e. consistent developer experience.
 4. Test framework should provide a large library to perform common tasks across the cluster like running containers, verifying traffic, etc so that writing new test cases becomes relatively easy.

We could argue about the best way to achieve those goals. 
But, this is what prompted me to create a new framework instead of using the one we had.

-sukhesh

erikh:
We have a full system test suite that we extracted from netplugin. It is 
used by the tests you intend to replace: 
https://github.com/contiv/systemtests-utils

Please use that as the rest of the project has aligned around it. We put 
man-hours into maintaining it and even extracted a portion of it as a 
standalone utility library for everyone to consume. If you do not like 
it, please put in the time to make it better instead of replacing it and 
forcing us to do this twice.

Additionally, we have strongly said many times that this is a Go 
project, even to the point of removing several chunks of python in the 
past, and the expectation is that it will remain a Go project. It needs 
to be accessible to all maintainers and there are a lot of concerns 
about dynamic language runtimes that I don't really want to get into 
here because they've already been said.

Given the above two things I wish you would have discussed this before 
surprising us. It would be much more fruitful to everyone if you 
contributed back to the systemtests-utils repo with your requirements so 
we can leverage it as a team, instead of a bespoke version of the same 
thing we have to maintain separately, in a language we don't work in as 
a team.

-Erik

core/core.go
```diff
@@ -41,7 +41,7 @@ type ServiceInfo struct {
 // comunicate.
 type Network interface {
 	CreateNetwork(id string) error
-	DeleteNetwork(id string) error
+	DeleteNetwork(id, encap string, pktTag, extPktTag int) error
```
mapuri: This looks a bit inconsistent, `CreateNetwork()` requires just an `id` but `DeleteNetwork()` requires a lot more info. Also by adding encap, pktTag arguments it seems like we are fixing networks to always have these attributes.

The original idea behind having a `Id` based signature was that we have a state store that can be indexed by the `Id` (or some form of it). So Create would just need Id to read the store do necessary updates and Delete would again just need an Id to read the store (this time a create runtime state) with enough info to perform the cleanup. May be that has changed now? In any case it might be a good idea to keep API signatures consistent irrespective of the approach  (i.e. state store based v/s state as API arguments)

netmaster/master/netmaster.go
```diff
@@ -110,33 +110,41 @@ func CreateTenant(stateDriver core.StateDriver, tenant *intent.ConfigTenant) err
...
+		log.Errorf("error updating tenant '%s'.Error: %s", tenant.Name, err)
+		return err
+	}
+
 	return nil
 }
 
 func startServiceContainer(tenantName string) error {
+	// do nothing in test mode
+	if testMode {
```
mapuri: hmm, not sure if this is a good idea i.e. mixing test logic with production logic.

We should explore using mocks or stubs for unit testing. For sanity/system tests we should never need to block any production code.

netmaster/master/network.go
```diff
@@ -321,23 +279,29 @@ func CreateNetwork(network intent.ConfigNetwork, stateDriver core.StateDriver, t
...
+		return err
 	}
 
 	return nil
 }
 
-func attachServiceContainer(tenantName string, networkName string, stateDriver core.StateDriver) error {
+func attachServiceContainer(tenantName, networkName string, stateDriver core.StateDriver) error {
+	// do nothing in test mode
+	if testMode {
```
mapuri: same as before, we shall consider a better way to test.

netmaster/mastercfg/policyState.go
```diff
@@ -0,0 +1,388 @@
...
+		if epg == nil {
+			log.Errorf("Error finding endpoint group %s", epgKey)
+			return nil, core.Errorf("endpoint group not found")
+		}
+
+		remoteEpgID = epg.EndpointGroupID
+	}
+
+	// Set protocol
+	switch rule.Protocol {
```
mapuri: This is more of a nit and we can decide to fix it later or not fix it at all. But I feel we can do better by using Go Interfaces and modelling different policies/rules as implementation of an interface. That way we can possibly avoid single function with a lot of switch-case statements and have simpler, smaller functions.

scripts/python/testcases/tcTrigger.py
```diff
@@ -0,0 +1,253 @@
...
+
+# Check ping between all containers in the same network
+def checkPingContainersInNetworks(containers, networks):
+    cntrIpList = []
+    # Read all IP addresses
+    for cnt in containers:
+        cntrIp = cnt.getIpAddr()
+        cntrIpList.append(cntrIp)
+
+    nidx = random.randint(0, (len(networks) - 1))
```
mapuri: my general concern with random testing is that it can soon become a sanity nightmare if there is no where to reproduce an issue and verify fix. A test environment should be repeatable as much as possible. If I am reading this correct, it doesn't seem possible to repeat a test run with the current approach?

mapuri:
code looks fine overall. 

Regarding deprecated files and system tests, I agree with @erikh. 

We should get rid of the deprecated files altogether.

For tests, the current systemtests library that has been extracted out does a decent job of bringing out a notion of testbed setup, cleanup and comand execution. We had started with vagrant and dind but I don't see a good reason why it can't be extended to other testbed types. In fact if I read it correctly the initial set of python test infra does a lot of same work as systemtest utils.

Regarding python being a better tool for system tests, I think it can be argued both ways. Python is a good tool and if this would have been a project in C or C++ I would have preferred a python based test to say using a gtest library. Go on the other hand provides enough tie in for test infrastructure and libs like go-checks and expressiveness that it is hard to see any benefit of python. Albeit we run the risk of Go programmers (our most likely contributors) forced to write tests in python and making a mess of it.

unclejack:
We really need to use just one single language for the code and for testing. All the testing needs to be done in Go because of the following reasons:
- using Go both for the main code and for testing allows us to use the same language for everything, no need to think in two languages and take into account differences between the two all the time
- Go has tooling which lets us keep the quality of the code high and adhere to generally accepted standards in the community
- potential contributors will discover that it's easier to contribute to the project if there's only one language used across the entire project
- similar to the last point, people who are just starting to code to contribute to this project and to other Go projects will find it easier to pick up just one language, rather than having to pick up two different languages
- I have contributed the initial version of this integration testing framework https://github.com/docker/docker/tree/master/integration-cli ; this was meant to be very close to shell scripts and it has been extremely successful (just look how small it was when it was contributed and how much it has evolved since its inception)
- Using two different languages allows us to play fast and loose with a critical part of the project - its testing; even with the integration-cli testing, there have been numerous problems in the Docker project; imagine having code quality and hidden bugs in the testing code, in addition to kernel and application code bugs

All in all, let's stick to Go.

netmaster/master/netmaster.go
erikh: If this is mocking the server end that may be harder, but this could 
also be accomplished with build tags + some careful refactoring. It 
would work *similarly* to mocks, but for different builds of the 
product.

I can assist with this conversion if that's necessary and desired.

core/core.go
shaleman: We need encap/pktTag to cleanup OVS state. problem here was, when etcd sends delete notification, object is gone. So, only way to get the state back is using an oper state. I want to avoid keeping duplicate info in opera state. So, only way was to pass it down the API. But, thinking about this. Now that I've moved network/endpoint state into mastercfg, better way might be to pass the state itself plugin API calls. I'll file an issue to pick this up in next round of cleanup.

netmaster/master/netmaster.go
shaleman: @mapuri @erikh Adding a mock infra sounds like a good idea. Please help contribute.

netmaster/mastercfg/policyState.go
shaleman: I've gone thru multiple iteration of this. Code here is little long to avoid complexity in other places.
I'm sure there are other ways of doing it :)

scripts/python/testcases/tcTrigger.py
shaleman: It is generally possible to reproduce an issue. Here is the thinking: assume there are `n` triggers, there is `n*(n-1)` different combination of triggers possible. We run thru `K * n*(n-1)` iterations in a random order to make sure every possible combination gets tested multiple times. So, its unlikely we'll hit an issue that is not possible to reproduce. 

If we start seeing unreproducible issues here, we can change the random order to a fixed order which tests all possible combinations.

shaleman:
@jainvipin 

> Is there a new README.md, I see the original one purged.

I removed dockcontivnet's read me since its not used anymore. netplugin's readme and docs have been up to date with dockplugin changes already.

core/core.go
mapuri: sounds good

netmaster/master/netmaster.go
mapuri: sure, if you can point me to the test that depends on this, I might be able to help.

netmaster/mastercfg/policyState.go
mapuri: sounds good

scripts/python/testcases/tcTrigger.py
mapuri: sure sounds good.

mapuri:
LGTM, assuming python tests and deprecated code related comments are addressed with some action or no action ;)

jainvipin:
@mapuri, @erikh - I will convert the test from python to Go, which if okay-ed by the MAINTAINERs woul be merged in soon. In order to not keep the system without sanity-less for now we'll go ahead with python sanity and look for my checkin to move all that to go using system-utils.

mapuri:
SGTM

jainvipin:
LGTM; will move python tests to Go soon after

PR #206, opened 2015-12-11
fix make release

shaleman:
Small change to fix make release target to push release binaries to github


shaleman:
LGTM :)
I'm going to merge this since its a small ENV variable change..

PR #211, opened 2015-12-17
netctl: fix http handlers to not try to print the body of a nil resp

### LINKED ISSUE #209: Deleting a network while it is being used by an endpoint causes http panic
Logs and backtrace:
https://gist.github.com/aa0ffeff1cdb5dc1ee8d

INFO[0000] Deleting network default:contiv-net
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x472689]

goroutine 1 [running]:
github.com/contiv/netplugin/netctl.writeBody(0x0, 0xc82008cfc0)
        /opt/gopath/src/github.com/contiv/netplugin/netctl/http.go:41 +0x119
github.com/contiv/netplugin/netctl.deleteURL(0xc82008cfc0, 0xc820064dc0, 0x36)
        /opt/gopath/src/github.com/contiv/netplugin/netctl/http.go:72 +0x115
github.com/contiv/netplugin/netctl.deleteNetwork(0xc82008cfc0)
        /opt/gopath/src/github.com/contiv/netplugin/netctl/netctl.go:211 +0x4b5
github.com/codegangsta/cli.Command.Run(0x7bc9d8, 0x6, 0x0, 0x0, 0x0, 0x0, 0x0, 0x7f4a00, 0x10, 0x0, ...)
        /opt/gopath/src/github.com/contiv/netplugin/Godeps/_workspace/src/github.com/codegangsta/cli/command.go:131 +0x1055
github.com/codegangsta/cli.(*App).RunAsSubcommand(0xc82008cd80, 0xc82008cc60, 0x0, 0x0)
        /opt/gopath/src/github.com/contiv/netplugin/Godeps/_workspace/src/github.com/codegangsta/cli/app.go:280 +0x1168
github.com/codegangsta/cli.Command.startApp(0x7bd8d8, 0x3, 0x0, 0x0, 0x0, 0x0, 0x0, 0x802fa0, 0x1a, 0x0, ...)
        /opt/gopath/src/github.com/contiv/netplugin/Godeps/_workspace/src/github.com/codegangsta/cli/command.go:209 +0x7cc
github.com/codegangsta/cli.Command.Run(0x7bd8d8, 0x3, 0x0, 0x0, 0x0, 0x0, 0x0, 0x802fa0, 0x1a, 0x0, ...)
        /opt/gopath/src/github.com/contiv/netplugin/Godeps/_workspace/src/github.com/codegangsta/cli/command.go:59 +0x15f1
github.com/codegangsta/cli.(*App).Run(0xc82008cb40, 0xc8200640c0, 0x4, 0x4, 0x0, 0x0)
        /opt/gopath/src/github.com/contiv/netplugin/Godeps/_workspace/src/github.com/codegangsta/cli/app.go:173 +0xfb4
main.main()
        /opt/gopath/src/github.com/contiv/netplugin/netctl/netctl/main.go:15 +0xb4

goroutine 17 [syscall, locked to thread]:
runtime.goexit()
        /usr/local/go/src/runtime/asm_amd64.s:1696 +0x1

erikh:
This fixes an issue with netctl's handling of errors. When an error is received currently where no response is provided (like an invalid read), we still try to print the body of the response in the error.

This fixes this to only print when the response is non-nil. This should resolve the error.

closes #209 

### DIFF
--- a/netctl/http.go
+++ b/netctl/http.go
@@ -13,6 +13,12 @@ import (
 
 var client = &http.Client{}
 
+func handleBasicError(ctx *cli.Context, err error) {
+	if err != nil {
+		errExit(ctx, exitRequest, err.Error(), false)
+	}
+}
+
 func baseURL(ctx *cli.Context) string {
 	return ctx.GlobalString("netmaster")
 }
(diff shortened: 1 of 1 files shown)


erikh:
oh, fwiw, I had to `git clean -fdx` to get things to build properly

jojimt:
LRTM

netctl/http.go
```diff
@@ -54,7 +54,10 @@ func postMap(ctx *cli.Context, url string, jsonMap map[string]interface{}) {
 
 	resp, err := client.Post(url, "application/json", bytes.NewBuffer(content))
 	if err != nil {
-		writeBody(resp, ctx)
+		if resp != nil {
```
mapuri: I am just wondering if there is a real case to even print a `resp` if `err` was not nil? I think most libraries follow this semantic that if they return a non nil error than caller should not interpret the other return values. `err` being not nil here also most likely implies that POST itself failed and we should not get anything back from the server so we can skip trying to write a nil response back.

erikh: I can remove it entirely.

mapuri: yeah, I would much prefer that. Do you agree?

erikh: Well, I thought it would be a useful debug tool. Perhaps we can just have `--debug` cli flag that has this behavior

erikh: But really, I'm having trouble thinking of cases where we have both an error and a body, and I'm struggling. Not quite sure what I was thinking now. Maybe we should just remove it?

mapuri: yeah, let's remove until we encounter a case.

erikh:
@mapuri PTAL. It now is a lot simpler too

mapuri:
yes, this is much cleaner and simpler ! LGTM

thanks @erikh

PR #212, opened 2015-12-19
dev guide documentation

jainvipin:
This PR brings the step guide for developers to make changes to netplugin

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,9 +36,6 @@ deps:
 checks:
 	./scripts/checks "$(PKGS)"
 
-host-build:
-	sudo /bin/bash -c 'source /etc/profile.d/envvar.sh; make run-build'
-
 run-build: deps checks clean
 	godep go install -v $(TO_BUILD)
 
@@ -86,6 +83,22 @@ centos-tests:
 
 sanity-test: start
 	vagrant ssh netplugin-node1 -c 'bash -lc "cd /opt/gopath/src/github.com/contiv/netplugin/scripts/python && PYTHONIOENCODING=utf-8 ./sanity.py -nodes 192.168.2.10,192.168.2.11"'
+
+host-build:
+	@echo "dev: making binaries..."
+	sudo /bin/bash -c 'source /etc/profile.d/envvar.sh; make run-build'
+
+host-unit-test:
+	@echo dev: running unit tests...
+	cd $(GOPATH)/src/github.com/contiv/netplugin && sudo -E PATH=$(PATH) scripts/unittests
+
+host-sanity-test:
+	@echo dev: running sanity tests...
+	cd $(GOPATH)/src/github.com/contiv/netplugin/scripts/python && PYTHONIOENCODING=utf-8 ./sanity.py -nodes ${CLUSTER_NODE_IPS}
+
+host-cleanup:
+	@echo dev: cleaning up, restarting services...
+	cd $(GOPATH)/src/github.com/contiv/netplugin/scripts/python && PYTHONIOENCODING=utf-8 ./cleanup.py -nodes ${CLUSTER_NODE_IPS}
 
 tar: clean-tar build
 	@echo "v0.0.0-`date -u +%m-%d-%Y.%H-%M-%S.UTC`" > $(VERSION_FILE)
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -25,6 +25,8 @@ echo 'export PATH=$PATH:/usr/local/go/bin:$GOBIN' >> /etc/profile.d/envvar.sh
 echo "export http_proxy='$4'" >> /etc/profile.d/envvar.sh
 echo "export https_proxy='$5'" >> /etc/profile.d/envvar.sh
 echo "export no_proxy=192.168.2.10,192.168.2.11,127.0.0.1,localhost,netmaster" >> /etc/profile.d/envvar.sh
+echo "export CLUSTER_NODE_IPS=192.168.2.10,192.168.2.11" >> /etc/profile.d/envvar.sh
+
 
 source /etc/profile.d/envvar.sh
 
--- a/docs/DevEnv.md
+++ b/docs/DevEnv.md
@@ -1,0 +1,62 @@
+## Netplugin Developer's Guide
+This document describes the development environment and typical steps to hack on netplugin.
+While there are alternative ways, the steps listed here are used by a lot of developers while
+making incremental changes and doing frequent compilation for a quicker development cycle.
+
+Happy hacking!
+
+### 1. Check-out a tree. `Estimated time: 2 minutes`
+Notes: Make sure GOPATH is set. This is a one time activity.
+```
+$ cd $GOPATH
+$ mkdir -p src/github.com/contiv
+$ cd src/github.com/contiv
+# it is recommended that you fork the repo if you want to make contributions
+$ git clone https://github.com/<your-github-id>/netplugin.git
+```
+
+### 2. Create development VMs. `Estimated time: 3-4 minutes`
+Note: This is a one time activity
+```
+$ cd $GOPATH/src/github.com/contiv/netplugin
+$ make start
+```
+
+### 3. Make code changes
+Notes: This must be done inside a VM. Note that netplugin repo is mounted from the host,
+therefore any changes changes are saved outside VM and are not lost if VM crashes or dies or any reason
+```
+# ssh into one of the VMs
+$ make ssh
+$ cd $GOPATH/src/github.com/contiv/netplugin
+# make code changes here and add unit/system tests for your changes
+. . .
+$ cd $GOPATH/src/github.com/contiv/netplugin
+# compile the recently made changes. `Estimated time: 1m 20s`
+$ make host-build
+```
+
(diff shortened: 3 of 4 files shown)


jainvipin:
@mapuri, @erikh, @unclejack PTAL
Better yet, try it out if you could and confirm if things look inline with docs, etc.

Vagrantfile
```diff
@@ -25,6 +25,8 @@ echo 'export PATH=$PATH:/usr/local/go/bin:$GOBIN' >> /etc/profile.d/envvar.sh
 echo "export http_proxy='$4'" >> /etc/profile.d/envvar.sh
 echo "export https_proxy='$5'" >> /etc/profile.d/envvar.sh
 echo "export no_proxy=192.168.2.10,192.168.2.11,127.0.0.1,localhost,netmaster" >> /etc/profile.d/envvar.sh
+echo "export CLUSTER_NODE_IPS=192.168.2.10,192.168.2.11" >> /etc/profile.d/envvar.sh
```
erikh: what is this?

erikh: n/m

docs/DevEnv.md
```diff
@@ -0,0 +1,66 @@
+## Netplugin Developer's Guide
+This document describes the development environment and typical steps to hack on netplugin.
+While there are alternative ways, the steps listed here is used by a lot of developers while
+making incremental changes and doing frequent compilation for a quicker development cycle.
+
+Happy hacking!
+
+### 1. Check-out a tree. `Estimated time: 2 minutes`
+Notes: Make sure GOPATH is set. This is a one time activity.
```
mapuri: I am wondering if setting GOPATH on host is really needed.

I remember this was required for running systemtests but since right now the tests are in python this might not be a requirement anymore. Is this requirement coming from anything else?

docs/DevEnv.md
```diff
@@ -0,0 +1,66 @@
...
+```
+$ cd $GOPATH
+$ mkdir -p src/github.com/contiv
+$ cd src/github.com/contiv
+# it is recommended that you fork the repo if you want to make contributions
+$ git clone https://github.com/<your-github-id>/netplugin.git
+
+# you can clone contiv repo, if you like playing with some code before forking it
```
mapuri: since this is a development guide I would avoid adding steps for non developers here i.e. users who are not planning to planning to make code changes. We might want to have a separate demo/trying out guide for that.

docs/DevEnv.md
```diff
@@ -0,0 +1,66 @@
...
+$ git clone https://github.com/contiv/netplugin.git
+```
+
+### 2. Create development VMs. `Estimated time: 3-4 minutes`
+Note: This is a one time activity
+```
+$ cd $GOPATH/src/github.com/contiv/netplugin
+$ make demo
```
mapuri: I think we should be doing `make start` here as `make demo` seems to build and start the netplugin, which doesn't gel well for development environment where the build and test will be done separately.

docs/DevEnv.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+### 3. Make code changes
+Notes: This must be done inside a VM. Note that the GOPATH is mounted from host
```
mapuri: > Note that the GOPATH is mounted from host

I think it should be `Note that the netplugin repo is mounted from the host`

docs/DevEnv.md
```diff
@@ -0,0 +1,66 @@
...
+Note: Again, this is also done from inside the VM. System tests would run across multiple 
+hosts (vm1 and vm2). Therefore it is important to keep both VMs (spun up from make-demo) running 
+otherwise it may not run multi-host networking tests well enough. The time taken to run
+system tests will be higher first time because the tests will download some containers for testing
+```
+$ make host-sanity
+```
+
+### 6. Commit changes to your fork; submit PR
+Note: If this is best done outside the VM assuming you do not want to populate git credentials
```
mapuri: I think it is not clear here if we are suggesting to do commits from inside or out, should this be reworded as?

```
Note: This is best done outside the VM using you git credentials setup on the host.
```

docs/DevEnv.md
jainvipin: @mapuri - I am thinking that developer workflow is that repo is imported from the host. This can only be done if GOPATH is set in Vagrant. If GOPATH is not set it checks out a new tree inside VM, something I'd not prefer a developer to use.

docs/DevEnv.md
jainvipin: Yes this step could be cloning your own fork; I will make the change

docs/DevEnv.md
mapuri: I see. I missed that the current Vagrantfile depends on host's GOPATH. Ideally we just need to mount the current directory to the GOPATH inside the vm for things to work, but the current step as-is works too.

unclejack: Is there a reason behind not wanting to bind mount this into the VM's GOPATH? It seems that would make it less likely to cause troubles.

docs/DevEnv.md
jainvipin: valid point; `make start` is all clean (demo would sound repurposing something else for this). I tried it out and it works.

docs/DevEnv.md
jainvipin: will take your suggestion... thanks for the wording.

docs/DevEnv.md
jainvipin: @mapuri, @unclejack  - current Vagrant doesn't depend on the host's GOPATH. 
However if it is set, it would mount the $GOPATH/src into the VM, otherwise just the netplugin. The reason I prefer setting GOPATH is because it allows for entire GOPATH to be available inside instead of just netplugin.

jainvipin:
@mapuri PTAL - updated with some feedback from you

docs/DevEnv.md
mapuri: @jainvipin 

> The reason I prefer setting GOPATH is because it allows for entire GOPATH to be available inside instead of just netplugin.

I just looked at Vagrantfile (https://github.com/contiv/netplugin/blob/master/Vagrantfile#L122), it doesn't mounts the entire host's GOPATH but just the netplugin repo. But the Vagrantfile still depends on GOPATH being defined in host's environment, which is what I was hoping that we could skip.

@unclejack 

> Is there a reason behind not wanting to bind mount this into the VM's GOPATH?

As I discussed above the netplugin repo is indeed being bind mounted at GOPATH in side the vm. My concern is more around the work needed for having GOPATH to be set in the host itself.

jainvipin: @mapuri - my understanding is different, can we chat offline? and post our findings here... I think entire src is mounted when GOPATH is set, instead when not set only current directory is mounted.

docs/DevEnv.md
```diff
@@ -0,0 +1,62 @@
...
+Note: Again, this is also done from inside the VM. System tests would run across multiple 
+hosts (vm1 and vm2). Therefore it is important to keep both VMs (spun up from make-demo) running 
+otherwise it may not run multi-host networking tests well enough. The time taken to run
+system tests will be higher first time because the tests will download some containers for testing
+```
+$ make host-sanity
+```
+
+### 6. Commit changes to your fork; submit PR
+Note: This is best done outside the VM using you git credentials setup on the host.
```
mapuri: s/ using you git / using your git

docs/DevEnv.md
```diff
@@ -0,0 +1,62 @@
...
+# iterate back to step 3 if tests fails or you need to make more code changes
+```
+
+### 5. Run system tests `Estimated Time: 90 mins`
+Note: Again, this is also done from inside the VM. System tests would run across multiple 
+hosts (vm1 and vm2). Therefore it is important to keep both VMs (spun up from make-demo) running 
+otherwise it may not run multi-host networking tests well enough. The time taken to run
+system tests will be higher first time because the tests will download some containers for testing
+```
+$ make host-sanity
```
mapuri: s / host-sanity / host-sanity-test

Makefile
```diff
@@ -87,6 +84,22 @@ centos-tests:
 sanity-test: start
 	vagrant ssh netplugin-node1 -c 'bash -lc "cd /opt/gopath/src/github.com/contiv/netplugin/scripts/python && PYTHONIOENCODING=utf-8 ./sanity.py -nodes 192.168.2.10,192.168.2.11"'
 
+host-build:
+	@echo "dev: making binaries..."
+	sudo /bin/bash -c 'source /etc/profile.d/envvar.sh; make run-build'
+
+host-unit-test:
+	@echo dev: running unit tests...
+	cd /opt/gopath/src/github.com/contiv/netplugin && sudo -E PATH=$(PATH) scripts/unittests
```
mapuri: should we replace `/opt/gopath` with $GOPATH ?

mapuri:
LGTM other than last two minor comments.

erikh:
LGTM

docs/DevEnv.md
jainvipin: based on the offline discussions, we'll keep the GOPATH settings as is for now. Vagrant file does handle spinning up the environment whether GOPATH is set or not.

Issue #214, opened 2015-12-23
network create should honor the resource ranges specified in the tenant it belongs to.


vvb:
##### Create tenant
```
vagrant@netplugin-node1:~$ netctl tenant create pepsi --subnet-pool 30.1.0.0/24 
--vlans "1200-1500" --vxlans "2001-3000"
INFO[0000] Creating tenant: pepsi
vagrant@netplugin-node1:~$
```

##### Create a network in the above tenant
```
vagrant@netplugin-node1:~$ netctl net create pepsi-net --tenant pepsi 
--subnet=30.1.2.0/24 --gateway=30.1.2.254 --pkt-tag=1300
vagrant@netplugin-node1:~$
```

##### List tenants
```
vagrant@netplugin-node1:~$ netctl tenant list
Name     Subnet Pool    Subnet Len  vlans        vxlans
------   -----------    ----------  -----        ------
default  10.1.1.1/16    24          100-1100     1001-1100
pepsi    30.1.0.0/24    24          1200-1500    2001-3000
vagrant@netplugin-node1:~$
```

##### List networks
```
vagrant@netplugin-node1:~$ netctl net list --tenant pepsi
Tenant  Network    Encap type  Packet tag  Subnet       Gateway
------  -------    ----------  ----------  -------      ------
pepsi   pepsi-net  vxlan       1300        30.1.2.0/24  30.1.2.254
vagrant@netplugin-node1:~$
```

###### The same is true for subnets also.
```
vagrant@netplugin-node1:~$ netctl net create pepsi-net1 --subnet=50.1.2.0/24
--gateway=50.1.2.254 --pkt-tag=200 --tenant pepsi
vagrant@netplugin-node1:~$ netctl net list --tenant pepsi
Tenant  Network     Encap type  Packet tag  Subnet       Gateway
------  -------     ----------  ----------  -------      ------
pepsi   pepsi-net   vxlan       1300        30.1.2.0/24  30.1.2.254
pepsi   pepsi-net1  vxlan       200         50.1.2.0/24  50.1.2.254
vagrant@netplugin-node1:~$
```

shaleman:
Alan/Vxlan resources have been moved to global level
Closing this bug

Issue #216, opened 2015-12-23
Error in network creation when a large VXLAN(>4095) is used


vvb:
#### Error in network creation when a large VXLAN(>4095) is used
```
vagrant@netplugin-node1:~$ netctl tenant create -p 100.1.0.0/16 -vlans 2000-2100
-vxlans 22000-22100 tenant1
INFO[0000] Creating tenant: tenant1
vagrant@netplugin-node1:~$
vagrant@netplugin-node1:~$ netctl tenant list
Name     Subnet Pool     Subnet Len  vlans        vxlans
------   -----------     ----------  -----        ------
default  10.1.1.1/16     24          100-1100     1001-1100
pepsi    30.1.0.0/24     24          1200-1500    2001-3000
tenant1  100.1.0.0/16    24          2000-2100    22000-22100
vagrant@netplugin-node1:~$
vagrant@netplugin-node1:~$
vagrant@netplugin-node1:~$ netctl net create -t tenant1 -p 22001 -s 100.1.1.0/24
-g 100.1.1.254 net1
500 Internal Server Error: Post http:///run/docker/plugins/netplugin.sock/
    IpamDriver.RequestPool: dial unix /run/docker/plugins/netplugin.sock: connection refused

ERRO[0015] Status 500 in request response
vagrant@netplugin-node1:~$
```

##### netplugin log
```
time="2015-12-23T04:43:48Z" level=info msg="Installing Group entry: &{Header:{Version:4 Type:15 Length:8 Xid:948} Command:1 Type:0 pad:0 GroupId:6 Buckets:[{Length:16 Weight:1 WatchPort:4294967295 WatchGroup:4294967295 pad:[0 0 0 0] Actions:[0xc820252270 0xc8205de520 0xc820252240]}]}"
time="2015-12-23T04:43:48Z" level=info msg="Installing vlan flood entry for vlan: 22001"
[...]

PR #217, opened 2015-12-23
fix and unit-tests for issue #214

vijayvikrant:


netmaster/master/netmaster_test.go
```diff
@@ -296,3 +308,33 @@ func TestVxlanConfigWithLateHostBindings(t *testing.T) {
...
+            "Name"              : "orange",
+            "PktTag"            : 1300
+        }]
+    }]}`)
+
+	initFakeStateDriver(t)
+	defer deinitFakeStateDriver()
+
+	// Create the Tenant and Network
+	applyConfig(t, cfgBytes)
```
jainvipin: can you also confirm if this also returns the error? 
Just not establishing the state is not sufficient perhaps for a -ve test case

netmaster/master/netmaster_test.go
```diff
@@ -296,3 +308,33 @@ func TestVxlanConfigWithLateHostBindings(t *testing.T) {
...
+	fakeDriver.DumpState()
+
+	keys := []string{"tenant-one"}
+	verifyKeys(t, keys)
+
+	keys = []string{"nets/orange"}
+	fakeDriver.DumpState()
```
jainvipin: will the above print a lot of gunk during test runs? Can we do that only when test case fails, or better yet remove it all together.

netmaster/master/netmaster_test.go
vijayvikrant: As of today create Network will not return an error for this. So, what you are saying is that I check for err nonetheless, so that when we fix it tomorrow, that also gets verified?

netmaster/master/netmaster_test.go
vijayvikrant: will remove DumpState statements.

netmaster/master/netmaster_test.go
jainvipin: @vijayvikrant - hmm.. if it doesn't return the error, how do we make sure that in reality it returns error as well (say from netctl/contivctl)?
> As of today create Network will not return an error for this. So, what you are saying is that I check for err nonetheless, so that when we fix it tomorrow, that also gets verified?

vijayvikrant: @jainvipin - I meant as of today CreateNetwork does not return an error. I am making changes to CreateNetwork to return error when the specified vxlan/vlan is out of allowed range for the tenant.. And also make the test case check for the error.

jainvipin: @vijayvikrant - got it... it seems like you plan to modify this test case once you fix issue; up to you if you want to combine this with other PR or keep them separate.
otherwise LGTM

vijayvikrant:
@jainvipin so, this issue was also present in our sanity script. I have modified the sanity code to make sure that the vlans/vxlans we allocate are in the tenant's configured ranges. The sanity passes after this change.

utils/netutils/netutils.go
```diff
@@ -213,6 +213,23 @@ func ParseTagRanges(ranges string, tagType string) ([]TagRange, error) {
 	return tagRanges, nil
 }
 
+// IsTagInRange verifies if the passed pktTag adheres to the valueRange
+func IsTagInRange(pktTag int, valueRangeStr string, encap string) bool {
+	// Parse the valueRange to form a []TagRange object
+	valueRanges, err := ParseTagRanges(valueRangeStr, "vlan")
+	if err != nil {
+		log.Errorf("Not able to pass valueRange. Err: %s", err)
```
jainvipin: minor typo: `parse` instead of `pass`

utils/netutils/netutils.go
```diff
@@ -213,6 +213,23 @@ func ParseTagRanges(ranges string, tagType string) ([]TagRange, error) {
...
+		return false
+	}
+
+	for _, valueRange := range valueRanges {
```
jainvipin: would you think it is better to this change in the ParseTagRanges itself, then we can be sure anyplace we parse the tagrange this is validated.

vijayvikrant: @jainvipin not sure I get it. The ranges are created in CreateTenant and so it makes sense to call  ParseTagRange there. CreateNetwork needs to check if the provided vlan/vxlan fall into the configured range. As per your comment CreateNetwork should do ParseTagRanges(Range, encap, vlan/vxlan) - which should validate if PktTag is in range and return multi-values (range, bool)?

vijayvikrant: @jainvipin a lot of places where we would parse tagRange, we won't have the Tag information yet. Like in the earlier example, CreateTenant will parse tagRange, but not have pktTag info yet. That would come later when netctl net create happens.. Am I missing something?

jainvipin: @vijayvikrant - as we talked offline, let's see if the call to ParseRange can ba avoided in IsTagInRange() i.e. pass the range from pre-parsed (tenant) to avoid doing this twice

netmaster/master/network.go
```diff
@@ -236,11 +276,18 @@ func CreateNetwork(network intent.ConfigNetwork, stateDriver core.StateDriver, t
 		nwCfg.ExtPktTag = int(extPktTag)
 		nwCfg.PktTag = int(pktTag)
 	} else if network.PktTagType == "vxlan" {
+		if !IsTagInRange(network.PktTag, "vxlan", gCfg.Auto.VXLANs, stateDriver, tenantName) {
+			return fmt.Errorf("vxlan %d does not adhere to tenant's vxlan range %s", network.PktTag, gCfg.Auto.VXLANs)
+		}
+
 		nwCfg.ExtPktTag = network.PktTag
 		nwCfg.PktTag = network.PktTag
```
jainvipin: @vvb - this line is still a problem.
The changes you made are needed, however not sufficient if we want to allow vxlan range beyond 4k. for example:
```
                } else if nwCfg.PktTagType == "vxlan" {
                        extPktTag, pktTag, err = gCfg.AllocVXLAN(rm)
                        if err != nil {
                                return err
                        }
                }
```
The following code in auto-allocation allows for both vlan-tag and vxlan-tag to be assigned; we need to follow the similar model for the code above to avoid a conflict and accessing beyond the range.

Bottom line: if you try a vxlan id beyond 4k range, it might still not work even with all your changes. In order to fix this you'd have to allocate a vlan-tag for it that is still in the allowed range. If you follow ofnet.AddVlan() call from ovsdriver.go for CreateNetwork() may be it will all make sense.

I understand you had not added all this, but let's fix this all of them.

vijayvikrant:
@jainvipin Yes, the plan is to do those changes as a part of PR for #216. This one is for #214 only.

jainvipin:
Thanks @vijayvikrant 

LGTM

jainvipin:
ping @mapuri 
can you also look at the diffs and confirm the changes?

netmaster/master/network.go
```diff
@@ -175,6 +176,45 @@ func deleteDockNet(tenantName, networkName, serviceName string) error {
 	return nil
 }
 
+// IsTagInRange verifies if the requested vlan/vxlan adheres to tenants configured range
+func IsTagInRange(pktTag int, pktTagType string, pktTagRange string, stateDriver core.StateDriver, tenantName string) bool {
```
mapuri: does this function need to be exported/public? Otherwise we can start it with lower-case.

Also can pktTag be negative? I am wondering if the `pktTag` argument to this function can be a uint instead to avoid the uint type cast below, otherwise we need to deal with testing input for negative numbers as well.

netmaster/master/netmaster_test.go
```diff
@@ -296,3 +308,164 @@ func TestVxlanConfigWithLateHostBindings(t *testing.T) {
...
+
+	tenant := cfg.Tenants[0]
+	err = CreateTenant(fakeDriver, &tenant)
+	if err != nil {
+		t.Fatalf("error '%s' creating tenant\n", err)
+	}
+
+	err = CreateNetworks(fakeDriver, &tenant)
+	if shouldFail {
+		if err == nil {
```
mapuri: it might be good to test that err returned is indeed for `OutOfRange` as `CreateNetworks()` might return other errors as well?

vijayvikrant: @mapuri I have made the changes for this. PTAL

netmaster/master/netmaster_test.go
```diff
@@ -296,3 +309,173 @@ func TestVxlanConfigWithLateHostBindings(t *testing.T) {
...
+
+		var expError string
+		network := tenant.Networks[0]
+		if network.PktTagType == "vlan" {
+			expError = fmt.Sprintf("vlan %d does not adhere to tenant's vlan range %s", network.PktTag, tenant.VLANs)
+		} else {
+			expError = fmt.Sprintf("vxlan %d does not adhere to tenant's vxlan range %s", network.PktTag, tenant.VXLANs)
+		}
+
+		if err.Error() != expError {
```
mapuri: this might panic if err is nil, so may be we should check for `err == nil` as well?

mapuri:
thanks @vijayvikrant , LGTM other than one minor comment.

netmaster/master/netmaster_test.go
vijayvikrant: @mapuri that is correct. I will change it.

vijayvikrant:
@mapuri added the check.

PR #218, opened 2015-12-23
utils/netutils: handle Atoi conversion error

unclejack:
The unit tests passed. The system tests are still running.

### DIFF
--- a/utils/netutils/netutils.go
+++ b/utils/netutils/netutils.go
@@ -221,8 +221,8 @@ func ParseCIDR(cidrStr string) (string, uint, error) {
 	}
 
 	subnetStr := strs[0]
-	subnetLen, _ := strconv.Atoi(strs[1])
-	if subnetLen > 32 {
+	subnetLen, err := strconv.Atoi(strs[1])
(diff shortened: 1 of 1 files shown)


utils/netutils/netutils.go
```diff
@@ -221,8 +221,8 @@ func ParseCIDR(cidrStr string) (string, uint, error) {
 	}
 
 	subnetStr := strs[0]
-	subnetLen, _ := strconv.Atoi(strs[1])
-	if subnetLen > 32 {
+	subnetLen, err := strconv.Atoi(strs[1])
+	if subnetLen > 32 || err != nil {
```
erikh: use ParseInt instead.

erikh: Seems atoi() is nothing like C atoi() so it's not broken. Sorry to confuse you.

unclejack: No worries, it's all good.

jainvipin:
LGTM

erikh:
LGTM

PR #224, opened 2016-01-08
k8s cni integration and related netplugin changes/fixes

jojimt:


netmaster/master/endpointGroup.go
```diff
@@ -17,7 +17,7 @@ package master
 
 import (
 	"errors"
-	"fmt"
+	//"fmt"
```
jojimt: Looks like one commit is missing. Let me fix this.

jainvipin:
@jojimt - can the title of the PR be `k8s cni integration and related netplugin changes/fixes`

jojimt:
Corrected it (what was meant to be description appeared as title…)

jainvipin:
@jojimt: we also need to update the docs with k8 integration, either with this PR or separate one, how to use k8s with contiv, etc.

jojimt:
Yes, that will be part of the next (Vagrant) PR.

contivk8s/clients/network.go
```diff
@@ -0,0 +1,122 @@
+/***
```
shaleman: As i mentioned, can we move contivk8s into mgmtfn/k8splugin hierarchy?

netmaster/master/endpointGroup.go
```diff
@@ -80,13 +80,15 @@ func CreateEndpointGroup(tenantName, networkName, groupName string, epgID int) e
 	}
 
 	// params for docker network
-	subnetCIDR := fmt.Sprintf("%s/%d", nwCfg.SubnetIP, nwCfg.SubnetLen)
+	if GetClusterMode() == "docker" {
```
shaleman: Better way to do this might be to create mgmtfn abstraction here. so that we can move docker/swarm specific things into docker mgmtfn.
We can do this as a subsequent PR too.

contivk8s/clients/network.go
```diff
@@ -0,0 +1,122 @@
...
+	"fmt"
+	log "github.com/Sirupsen/logrus"
+	"github.com/contiv/netplugin/mgmtfn/cniapi"
+	"io/ioutil"
+	"net"
+	"net/http"
+)
+
+const (
+	nwURL = "http://localhost"
```
shaleman: I'm not sure if we need this since we are using unix sockets

contivk8s/k8s_cni.go
```diff
@@ -0,0 +1,129 @@
...
+	ppInfo.IntfName = os.Getenv("CNI_IFNAME")
+	return nil
+}
+
+func addPodToContiv(nc *clients.NWClient, pInfo *cniapi.CNIPodAttr) {
+
+	// Add to contiv network
+	result, err := nc.AddPod(pInfo)
+	if err != nil {
+		log.Fatalf("EP create failed -- %s", err)
```
shaleman: How is error communicated back to kubelet? is it based on exit codes?
does log.Fatalf() set the correct exit code?

mgmtfn/k8splugin/driver.go
```diff
@@ -0,0 +1,370 @@
...
+	if !ok {
+		return -1, fmt.Errorf("Invalid nw name space: %v", ns)
+	}
+
+	elements := strings.Split(ns, "/")
+	return strconv.Atoi(elements[2])
+}
+
+// setIfAttrs sets the required attributes for the container interface
+func setIfAttrs(ifname, netns, cidr, newname string) error {
```
shaleman: Here is how old crtclient code was moving container to namespace without using nester,
https://github.com/contiv/netplugin/blob/k8s-crt-pre-deprecate/crtclient/docker/docker.go#L122

mgmtfn/k8splugin/driver.go
```diff
@@ -0,0 +1,370 @@
...
+	resp := epSpec{}
+
+	// Get labels from the kube api server
+	epg, err := kubeAPIClient.GetPodLabel(pInfo.K8sNameSpace, pInfo.Name, "net-group")
+	if err != nil {
+		log.Errorf("Error getting epg. Err: %v", err)
+		return &resp, err
+	}
+
+	netw, _ := kubeAPIClient.GetPodLabel(pInfo.K8sNameSpace, pInfo.Name, "network")
```
shaleman: error handling missing? :)

mgmtfn/cniapi/api.go
```diff
@@ -0,0 +1,48 @@
...
+
+package cniapi
+
+// cni api definitions shared between the cni executable and netplugin
+
+// PluginPath is the path to the listen socket directory for netplugin
+const PluginPath = "/run/contiv"
+
+// DriverName is the name of the listen socket for netplugin
+const DriverName = "netplugin"
```
shaleman: Looks like this const is not being used?

mgmtfn/k8splugin/cniserver.go
```diff
@@ -0,0 +1,162 @@
...
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/contiv/netplugin/mgmtfn/cniapi"
+	"github.com/contiv/netplugin/netplugin/plugin"
+	"github.com/gorilla/mux"
+)
+
+// contivKubeCfgFile holds credentials to access k8s api server
+const (
+	contivKubeCfgFile = "/opt/contiv/config/contiv.json"
```
shaleman: I'm not too happy about requiring user to create this config file.
But, i don't have an alternative at this point. so, we can keep it for now.

mgmtfn/cniapi/api.go
```diff
@@ -0,0 +1,48 @@
+/***
```
shaleman: as discussed, lets move this under mgmtfn/k8splugin hierarchy too.

shaleman:
LGTM other than the minor comments

contivk8s/clients/network.go
jojimt: Yes, will do that.

netmaster/master/endpointGroup.go
jojimt: I agree. However, I would rather do it as a separate PR.

contivk8s/clients/network.go
jojimt: Actually we do. This is an artifact of the net/http package where it rejects the URL if it did not have a domain name.

contivk8s/k8s_cni.go
jojimt: kubelet just checks for status 0 and treats everything else as an error without interpreting it.

mgmtfn/k8splugin/driver.go
jojimt: Yes I see the netns lib. I am not sure it will do everything we need. But I can look at that separately.

mgmtfn/k8splugin/driver.go
jojimt: Actually, if the first check passes the subsequent ones are guaranteed to pass. I will add a comment to that effect.

mgmtfn/cniapi/api.go
jojimt: Right, will remove.

mgmtfn/k8splugin/cniserver.go
jojimt: For the common scenarios, this is created using ansible; so it's not that bad. But, I agree.

mgmtfn/cniapi/api.go
jojimt: sure.

mgmtfn/k8splugin/cniserver.go
jainvipin: Is this a security problem?

mgmtfn/k8splugin/contivk8s/clients/network.go
```diff
@@ -0,0 +1,122 @@
...
+	c.baseURL = nwURL
+
+	transport := &http.Transport{Dial: unixDial}
+	c.client = &http.Client{Transport: transport}
+
+	return &c
+}
+
+// AddPod adds a pod to contiv using the cni api
+func (c *NWClient) AddPod(podInfo interface{}) (*cniapi.RspAddPod, error) {
```
jainvipin: Unit tests for AddPod and DelPod public functions?

mgmtfn/k8splugin/cniserver.go
jojimt: Well, kube set up uses these certs. Therefore, I don't consider it a security problem specific to contiv. There is also a different way of authenticating that I can look into, separately. However, I don't know that it is any more secure.

mgmtfn/k8splugin/cniserver.go
```diff
@@ -0,0 +1,162 @@
...
+)
+
+// ContivConfig holds information passed via config file during cluster set up
+type ContivConfig struct {
+	K8sAPIServer string `json:"K8S_API_SERVER,omitempty"`
```
jainvipin: Do we specify which version of API that APIServer interaction is on with?

mgmtfn/k8splugin/contivk8s/clients/network.go
jojimt: Because we moved the kubeserver access inside netplugin, these functions will need access to a kube server. I think sys-test is the only option. Will add it after vagrant work is completed.

mgmtfn/k8splugin/cniserver.go
jojimt: Right now it's hard coded to v1 like this:
const (
        nsURL = "/api/v1/namespaces/"
)
We can make it configurable, if necessary.

jainvipin: Ah... as long it is specified; I don't think there is a need for it to be configurable.

mgmtfn/k8splugin/contivk8s/clients/network.go
jainvipin: There should be a way to mock k8s REST handler; for example someone had mentioned an easy way to mock if we made the correct REST calls. Ultimately if you see these functions changing, then making unit tests pays off...

mgmtfn/k8splugin/contivk8s/clients/network.go
```diff
@@ -0,0 +1,122 @@
...
+	body := bytes.NewBuffer(buf)
+	url := c.baseURL + cniapi.EPDelURL
+	r, err := c.client.Post(url, "application/json", body)
+	if err != nil {
+		return err
+	}
+	defer r.Body.Close()
+
+	switch {
+	case r.StatusCode == int(404):
```
jainvipin: This code in switch statement looks similar in AddPod and DelPod; shoudl we combine the error handling in a common place?

jojimt: OK, will do.

mgmtfn/k8splugin/driver.go
```diff
@@ -0,0 +1,371 @@
...
+		if err != nil {
+			log.Errorf("unable to find link %q. Error %q", ifname, err)
+		}
+		return link, err
+	}
+	return link, err
+}
+
+// nsToPID is a utility that extracts the PID from the netns
+func nsToPID(ns string) (int, error) {
```
jainvipin: Perhaps we can use this in mesos plugin too... should this belong in netplugin/utils?

mgmtfn/k8splugin/contivk8s/clients/network.go
jojimt: I guess I can do that. Will add it.

mgmtfn/k8splugin/driver.go
```diff
@@ -0,0 +1,371 @@
...
+	epReq, err := getEPSpec(&pInfo)
+	if err != nil {
+		log.Errorf("Error getting labels. Err: %v", err)
+		return resp, err
+	}
+
+	err = epCleanUp(epReq)
+	resp.EndpointID = pInfo.InfraContainerID
+	return resp, err
+}
```
jainvipin: same comment - no unit tests for any of the functions in this file either...

mgmtfn/k8splugin/kubeClient.go
```diff
@@ -0,0 +1,174 @@
...
+	// initiate a get request to the api server
+	podURL := c.baseURL + ns + "/pods/" + name
+	r, err := c.client.Get(podURL)
+	if err != nil {
+		return err
+	}
+
+	defer r.Body.Close()
+	switch {
+	case r.StatusCode == int(404):
```
jainvipin: repetitive code - was also there in AddPod and DelPod functions above...

mgmtfn/k8splugin/kubeClient.go
```diff
@@ -0,0 +1,174 @@
...
+	// Treat missing metadata as a fatal error
+	if !ok {
+		return fmt.Errorf("metadata not found in podSpec")
+	}
+
+	p := &c.podCache
+	p.setDefaults(ns, name)
+
+	meta := m.(map[string]interface{})
+	l, ok := meta["labels"]
```
jainvipin: minor: if statement below this and this statement can be combined

netmaster/master/netmaster.go
```diff
@@ -36,6 +36,13 @@ const (
 	defaultSkyDNSImage  = "skynetservices/skydns:latest"
 )
 
+// Run Time config of netmaster
+type nmRunTimeConf struct {
+	clusterMode string
+}
+
+var masterRTCfg nmRunTimeConf
```
jainvipin: Should this type of information be added in global runtime structure? This allows for netmaster restartability - @shaleman WDYT?

jainvipin:
@jojimt - great work! thanks!
please identify the changes we need to make before the checkin vs. tests that you want to address as a part of subsequent checkins.

netmaster/master/netmaster.go
jojimt: This is basically environment information. I don't think we should save it in etcd. If the netmaster is restarted, that would be still in the same environment.

jainvipin: Hmm... got it; is this read always after netmaster start? If yes, we are covered.

jojimt: It's a command line argument, expected to come from the service env file.

shaleman: @jainvipin: my thinking is:
(1) Plugin modes (like docker/kubernetes/mesos plugins) that need to be specified before netplugin/netmaster initializes should be command line arguments.
(2) Cluster modes/Fabric modes/control plane modes should be an Intent that can change at runtime. We need little more infra in netplugin/netmaster for this to happen. Expect to work on this in next few PRs.

mgmtfn/k8splugin/contivk8s/clients/network.go
jojimt: After looking at this again, this does not fit as a separate routine as we have different return signatures. It is only 4 lines of code anyway.

mgmtfn/k8splugin/driver.go
jojimt: As we don't know the exact needs of mesos plugin at this point, I will leave this to @unclejack to refactor and move as appropriate.

mgmtfn/k8splugin/driver.go
jojimt: I have added UT for the REST interface. This one, seems like it will need to be covered by system test as this module makes multiple external calls.

mgmtfn/k8splugin/kubeClient.go
jojimt: I have intentionally stayed away from that style because I have seen that it limits the scope of the boolean to just the if. I find that scoping confusing.

jainvipin:
LGTM

PR #228, opened 2016-01-13
dockplugin: refactor to avoid return functions

unclejack:
This PR changes the handlers for dockplugin to not return functions when that's not needed.


mapuri:
LGTM. 

I like the fact that removing the extra return, also reduced the indentation for a whole bunch of code making it more readable :+1:

jainvipin:
+1 for the simplification

PR #229, opened 2016-01-13
Netplugin: L3 integration for Containers

abhinandanpb:
The commit contains changes to Netplugin required for L3 support for containers.

Changes contain : 
1) netctl changes to add,delete and list bgp neighbor object
2) netplugin changes to write,read and clear etcd bgp states
3) Additional netplugin attibutes to take in routerip and forwarding mode (routing vs bridge)
4) Subsequent mechanism to use ofnet library to provision goBgp server and intiate bgp session with the configured bgp peer.
5) Support to instantiate vlrouter datapath mode for routing mode.
6) Both Vlan and Vxlan ofnet agents are created if the forwarding mode is "routing"

Caveats and requirements
1) Currently only one bgp neighbor on a single uplink is supported
2) netplugin forwarding mode needs to be same across all netplugin nodes in the cluster (will need to   bring global object )
3) BGP peering between the Host and uplink switch need to be eBGP (different AS)

Usage : 
//Eg: to add bgp neighbor
netctl bgp add --host="contiv144" --as="500" --neighbor="50.1.1.2"

Eg: to start netplugin:
netplugin  -vlan-if=eth7 -vtep-ip=172.29.205.224 -router-ip=50.1.1.1 -fwd-mode="routing"


abhinandanpb:
https://gist.github.com/abhinandanpb/34c7af2f92fa7909a1ce

shaleman:
@abhinandanpb some comments:

+++ b/drivers/ovsbgpstate.go

Do we really need opera state for BGP? Cant we recover all BGP state from configuration? We should create opera state only if we can't recover some state from config.

+// DeleteBgpNeighbors deletes a bgp neighbor by named identifier
+func (d *OvsDriver) DeleteBgpNeighbors(id string) error {
+	log.Infof("delete router state %s \n", id)
+	//FixME: We are not maintaining oper state for Bgp
+	//Need to Revisit again

Hmm, not sure if thats correct.

+++ b/netmaster/mastercfg/routerState.go
+// CfgBgpState is the router Bgp configuration for the host
+type CfgBgpState struct {

If the object name is "BgpState" lets keep the file name bgpState. Or we can rename the object to "RouterState". Keep the naming consistent.

shaleman:
@abhinandanpb Can you post the complete diff in a gist too?
I'm not sure if all the Godep changes are necessary(for example logrus shouldn't have changed in your patch) I'm wondering if godep is picking older git hashes from your workspace..

abhinandanpb:
@shaleman regarding operstate. Agreed. I am not maintaining operstate for Bgp. Will get rid of the operstate file. 

Regarding the file name I will change it to bgpstate.

shaleman:
LGTM

PR #230, opened 2016-01-13
Makefile: make ssh target cd into directory

unclejack:
This tiny PR makes ssh change directory into the directory where we have the source code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ demo:
 	sleep 10
 
 ssh:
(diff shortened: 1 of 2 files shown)


mapuri:
LGTM 

Just a minor comment. As this appears to save a few commands for developers using vm as development environment. In that regard, should we update the development guide to omit the `cd` step?

unclejack:
@mapuri PTAL, I've made the change to the docs.

mapuri:
thanks @unclejack, still LGTM :)

PR #235, opened 2016-01-17
Vagrant cluster set up with k8s/contiv

jojimt:
Please refer to README (part of the PR) for details of bring up sequence.


Makefile
```diff
@@ -57,6 +57,16 @@ update:
 start: update
 	CONTIV_NODE_OS=${CONTIV_NODE_OS} vagrant up
 
+#kubernetes demo targets
+k8s-cluster:
```
jainvipin: Can we make these targets consistent with docker stack, which has following meanings to them:
- 'make start' = bring up vms, start basic services (does not bring up the netplugin, used by developers typically
- 'make stop' = stop and destroy vagrant vms
- 'make demo' = brings entire demo setup with services running, etc.

It seems like in k8s case:
- 'make kds-cluster' is equivalent to 'make demo'
- 'make k8s-destroy' is equivalent to 'make stop'
- 'make demo' and 'make demo-start' does not have an equivalent one which copies the appropriate docker images, etc. to get the demo going on. I am not very particular if these are not kept in contiv/demo repo; but name conflict is something to be avoided. 
@mapuri, @shaleman, @erikh  - any comments?

mgmtfn/k8splugin/README.md
```diff
@@ -0,0 +1,311 @@
+# Contiv Networking for Kubernetes
+
+Contiv is integrated with Kubernetes via a CNI plugin. This integration enables you
```
jainvipin: minor: 'you' may sound too direct. Suggested text:
"This integration allows using Contiv Networking and Policy for Pod's interconnectivity in a Kubernetes cluster"

mgmtfn/k8splugin/README.md
```diff
@@ -0,0 +1,311 @@
...
+#### Step 1: Clone and build kubernetes
+
+```
+$ mkdir ~/go/src/github.com/k8s
+$ cd ~/go/src/github.com/k8s
+$ git clone https://github.com/kubernetes/kubernetes
+$ cd kubernetes; make
+```
+
+Note: If the build fails, please retry. It will usually build the binaries on 2nd attempt
```
jainvipin: Suggest removing 'It will usually build the binaries on 2nd attempt". If at all something like this is needed "Often intermittent connectivity issues can result into a build failure, which is easily fixed by retrying the build"

mgmtfn/k8splugin/README.md
```diff
@@ -0,0 +1,311 @@
...
+
+#### Step 2: Clone contrib repo
+
+```
+$ cd ~/go/src/github.com/k8s
+$ git clone https://github.com/jojimt/contrib -b contiv
```
jainvipin: I suppose this repo is being upstreamed, and thus "git clone https://github.com/kubernetes/contrib' would be the eventual step here.

mgmtfn/k8splugin/README.md
```diff
@@ -0,0 +1,311 @@
+# Contiv Networking for Kubernetes
+
+Contiv is integrated with Kubernetes via a CNI plugin. This integration enables you
+to set up and manage inter-pod connectivity in a Kubernetes cluster using Contiv 
+policy model.
+
+## Getting Started
+
```
jainvipin: Reading all the steps looks like something developer friendly instructions when someone wants to hack on this. Should avoid:
- making/building binaries in user instructions
- checking out repos
- ansible work to install things (those all should be part of vbox image, for Vagrant)

We should find hard reasons to not follow what we have working with a swarm cluster i.e. following 4 steps for basic interconnectivity, and a few more for policies, etc.
- 'make demo'
- 'create networks'
- 'create containers'
- 'ping them'

ping @mapuri, @shaleman, @erikh, @unclejack

mgmtfn/k8splugin/README.md
```diff
@@ -0,0 +1,311 @@
...
+#### Step 3: Clone and build netplugin
+
+```
+$ cd ~/go/src/github.com/k8s
+$ git clone https://github.com/contiv/netplugin
+
+Note:
+*Until the PR is merged, you need to use the following instead*
+*$ git clone https://github.com/jojimt/netplugin -b k8s-vagrant*
+
```
jainvipin: Which merge is this dependent on? merge into netplugin, or k8s merge? If it is former, can we avoid this `Note` all together?

mgmtfn/k8splugin/README.md
jojimt: Making/building binaries will go away soon -- there is a follow up PR coming.
Checking out repos -- the ansible repo will need to be checked out, unless we want to provide a .tgz.

I did not quite follow this one:
"ansible work to install things (those all should be part of vbox image, for Vagrant)"

The ansible repo is there and the packages are all pre-installed in the vbox.

In terms of the steps, are you suggesting there should be 4 steps instead of the 7 that I have?

mgmtfn/k8splugin/README.md
jojimt: OK, will change.

mgmtfn/k8splugin/contivk8s/vagrant/Vagrantfile
```diff
@@ -0,0 +1,73 @@
...
+        config.vm.provider "virtualbox" do |v|
+          customize_vm(v)
+
+        # configure ip address etc
+        c.vm.hostname = vm_name
+        c.vm.network :private_network, ip: member_info["control-ip"]
+        c.vm.network :private_network, ip: member_info["contiv-network-ip"]
+        c.vm.provision "shell", inline: <<-EOS
+          sudo setenforce 0
+          sudo systemctl stop firewalld
```
jainvipin: seem specific to CENTOS; we use CENTOS flags in the vagrant used for swarm, can we capture under CENTOS flags; it is okay for us to use CENTOS as the default OS for K8s.

mgmtfn/k8splugin/contivk8s/vagrant/Vagrantfile
```diff
@@ -0,0 +1,73 @@
...
+      config.vm.define vm_name = member_info["name"] do |c|
+
+        # Customize VM for better performance
+        config.vm.provider "virtualbox" do |v|
+          customize_vm(v)
+
+        # configure ip address etc
+        c.vm.hostname = vm_name
+        c.vm.network :private_network, ip: member_info["control-ip"]
+        c.vm.network :private_network, ip: member_info["contiv-network-ip"]
```
jainvipin: do we also need one more network for vlan network? Or contiv-network would suffice for that?

mgmtfn/k8splugin/README.md
jojimt: Will remove -- this part will go away when I add support for installing pre-built k8s binaries.

mgmtfn/k8splugin/README.md
jojimt: Correct.

mgmtfn/k8splugin/README.md
jojimt: Yes, I will delete it -- this was just to cover anyone trying this over the weekend e.g. @vijayvikrant

mgmtfn/k8splugin/contivk8s/vagrant/export/nc-busybox/Dockerfile.nc-busybox
```diff
@@ -0,0 +1,5 @@
+FROM progrium/busybox
+RUN opkg-install netcat
```
jainvipin: +1
I was unaware of opkg-install, can we use it to install other utils too? iperf3, etc.?

mgmtfn/k8splugin/contivk8s/vagrant/vagrant_cluster.py
```diff
@@ -0,0 +1,117 @@
...
+    # read cluster config
+    clusterConf = readConfig()
+
+    # get the ssh config info from vagrant
+    res = os.system("vagrant ssh-config > ./.out.vagrant")
+    if res != 0:
+        print "Failed to get vagrant ssh config"
+        sys.exit(-1)
+
+    inFd = open('./.out.vagrant')
```
jainvipin: Is it better to open temp files in /tmp? that way we don't need to do '.gitignore' on them as well?

mgmtfn/k8splugin/contivk8s/vagrant/Vagrantfile
jojimt: If we are sticking to any one mode, one interface is sufficient. If we want to do mixed mode, we might need to look at this part.

mgmtfn/k8splugin/contivk8s/vagrant/Vagrantfile
jojimt: Yes, this Vagrantfile is only for centos for now.

jainvipin:
@jojimt - great work! thanks for the PR.
- There are few commits that can be squeezed (esp if 'nits' show up in github. For example when someone reads 'yet another edit' type of comments
- If you plan to add follow-up some of the larger comments subsequently, I recommend opening Issues for them so we don't lose track fo them.

mgmtfn/k8splugin/contivk8s/vagrant/vagrant_cluster.py
jojimt: IMO, no. If we open in /tmp and you have multiple users sharing a machine we will run into trouble.

mgmtfn/k8splugin/contivk8s/vagrant/Vagrantfile
jainvipin: I was suggesting casing the changes with, for example as in Vagrant file for docker-stack:
`f ENV['CONTIV_NODE_OS'] && ENV['CONTIV_NODE_OS'] == "centos" then`
For this Vagrant file you can set the environment variable automatically up at the top, but it can keep the changes well isolated that are distro specific vs. not.

mgmtfn/k8splugin/contivk8s/vagrant/Vagrantfile
jainvipin: We'd not want mixed mode i.e. simultaneous run of both types of interfaces, however assuming that we intend to run sanity tests for k8s then it is better to add this, because sanity runs can work without restarting the VMs in both modes.

mgmtfn/k8splugin/README.md
jainvipin: > Making/building binaries will go away soon -- there is a follow up PR coming.
> Checking out repos -- the ansible repo will need to be checked out, unless we want to provide a .tgz.

+1 
Let's track it with an issue if you are going to work on it later; that way someone else can also pick it up.

> I did not quite follow this one:
"ansible work to install things (those all should be part of vbox image, for Vagrant)"

> The ansible repo is there and the packages are all pre-installed in the vbox.

Then don't bother with my comment; I was suggesting that ansible should not install packages.

> In terms of the steps, are you suggesting there should be 4 steps instead of the 7 that I have? I think we will automatically get there when the pull repo/build binaries part goes away.

That would be awesome!

mgmtfn/k8splugin/README.md
jainvipin: We should open an issue to track this - assuming upstreaming can take some time (reviews from k8s team, and making in official master, etc.)

mgmtfn/k8splugin/README.md
jainvipin: :-)
@vijayvikrant, @DivyaVavili - please note this workaround (if needed) 
however this instruction would be removed.

mgmtfn/k8splugin/README.md
jojimt: Done.

mgmtfn/k8splugin/contivk8s/vagrant/vagrant_cluster.py
jainvipin: Multiple users can't work on the same machine when we are sharing etcd clustering to be same, netplugin can't be run twice in the same host, and k8s can only interact with one of them.
That said; it may be okay getting a random number to open up the file in /tmp if at all that was a concern. I will let others also chime in and provide their opinion.

mgmtfn/k8splugin/README.md
jojimt: Opened https://github.com/contiv/netplugin/issues/236

mgmtfn/k8splugin/contivk8s/vagrant/Vagrantfile
jojimt: Will cover as part of https://github.com/contiv/netplugin/issues/238.

mgmtfn/k8splugin/contivk8s/vagrant/vagrant_cluster.py
erikh: http://ruby-doc.org/stdlib-2.1.0/libdoc/tmpdir/rdoc/Dir.html might help

erikh:
+1 to squashing, if you're uncomfortable with the git commands required, just lmk and I will help.

mgmtfn/k8splugin/contivk8s/vagrant/Vagrantfile
jojimt: Will do as part of https://github.com/contiv/netplugin/issues/238

mgmtfn/k8splugin/contivk8s/vagrant/vagrant_cluster.py
jojimt: Is adding to .gitignore a big deal? I had some of the other scripts (e.g. ACI) use /tmp for the temp files. What I found is that some people either share machines or use multiple accounts. This causes problems. That's why I decided move away from /tmp. If .gitignore is deemed a problem, then we can look at alternatives.

jojimt:
When I did rebase -i to squash, some other commits are showing up as part of my PR. I hope that is ok.

erikh:
@jojimt I think this might get rid of those commits:


(inside your branch)

```
git fetch origin
git rebase origin/master
```

If that doesn't work, you can delete the commits with `git rebase -i` (just delete the line the commit is on) and that will do it, it's just the more risky method; practice in a clone of the branch if you need to.

jojimt:
No, it did not do it. I think it is best to deal with this during merge. @shaleman let me know if you would like me to do any change.

mgmtfn/k8splugin/contivk8s/vagrant/Vagrantfile
```diff
@@ -0,0 +1,73 @@
...
+# begin execution here
+cluster = read_cluster_config
+create_etc_hosts(cluster)
+
+Vagrant.configure(2) do |config|
+  config.vm.box = "contiv/k8s-centos"
+  config.vm.box_version = "0.0.6"
+
+  config.ssh.password = 'vagrant'
+  config.vm.synced_folder "./export", "/shared", disabled: false
```
shaleman: To actively develop in k8s env, it might be good idea to mount entire go path into the VMs so that you can quickly rebuild netplugin binaries and try the change.

jojimt: The problem is the time taken for sync. May be I can add an option that can be used by dev.
The work flow as of now is make run-build; make k8s-demo-start. The latter step copies the binaries and restarts affected components. It is reasonably quick, but we can refactor.

PR #242, opened 2016-01-19
Removing pre generated files

abhinandanpb:
Removing pre generated files from Godeps directory. The contivModel files will be generated a  part of make build / make demo / make run-build.


shaleman:
LGTM

PR #243, opened 2016-01-20
add versioning to netplugin binaries

jainvipin:
first cut changes; let's review, this...

Few things to note:
- I don't see much of a reason to do unit testing for two little functions in version/version.go - please tell me if I am off
- The more important unit-test is really the one that tests the REST endpoint to netmaster fetching the version info. I will work on doing unit test after @shaleman is done with checking in changes for doing netmaster unit tests using rest apis 
- I omitted showing netplugin version in `netctl version`, then what if the netplugin versions are different on different hosts? Besides, `netplugin -version` command can show that on individual hosts.
- `netplugin -version` shows some junk upfront, @unclejack has opened an issue for this, let's track it there.


version/generate_version
```diff
@@ -0,0 +1,40 @@
...
+  UseVersionStr="// +build !use_version" 
+  useDefaultVersionString="False"
+else
+  file_name=version_release.go
+  UseVersionStr="// +build use_version" 
+  useDefaultVersionString="True"
+fi
+
+GitCommitStr=`git show | grep commit | cut -f 2 -d ' ' | cut -b 1-10`
+VersionStr=`cat CURRENT_VERSION`
```
jojimt: Would be nice if we can also incorporate the output of "git status" with a flag such as "dirty" or "(pvt-patch)"

version/version.go
```diff
@@ -0,0 +1,44 @@
...
+
+	return &ver
+}
+
+// Print prints the versioning details
+func Print(ver *Info) {
+	fmt.Printf("Version: %s\n", ver.Version)
+	fmt.Printf("GitCommit: %s\n", ver.GitCommit)
+	fmt.Printf("BuildTime: %s\n", ver.BuildTime)
+}
```
jojimt: LRTM.

version/generate_version
jainvipin: Hmm.. I guess you are trying to find out if there were changes on top of  git-commit that might makes the image not exactly same as the git-commit level. That would be easy to do, let's get opinion from others too...
ping @unclejack @shaleman @erikh @mapuri

erikh:
```
git rev-parse HEAD
```

Will get you the latest commit of the current branch. (replace `HEAD` with `master` if you prefer)

version/version.go
```diff
@@ -0,0 +1,44 @@
...
+	ver := Info{}
+	ver.GitCommit = gitCommit
+	ver.Version = versionStr
+	ver.BuildTime = buildTime
+
+	return &ver
+}
+
+// Print prints the versioning details
+func Print(ver *Info) {
```
erikh: If you replace this with:

```
func (i *Info) String() {
   // returns a string instead
}
```

any print of `i` (e.g., `fmt.Println` or `fmt.Printf("%#v")`) will yield this data. I would prefer this as it's more flexible (and not a huge change from what we have here)

version/generate_version
jainvipin: thanks for the hint @erikh - will incorporate
WDYT about @jojimt's suggestion about including "+pvt" string to the commit string if there are some local changes on top of HEAD?

unclejack: We should add something like `-dirty` or `-unsupported` if there are any local changes on top of what's found in git.

version/version.go
jainvipin: @unclejack - makes sense; will change

version/generate_version
erikh: I wasn't sure of the appropriate status command, so I looked here, which has what we need:

https://github.com/docker/docker/blob/master/hack/make.sh#L70-L78

version/version.go
erikh: I'm @erikh :P

version/generate_version
jainvipin: @unclejack - I also think we should, let me add '-unsupported' to it...

jainvipin: @erikh - even better... will just use that

shaleman: adding -unsupported automatically sounds nice.
Since version is a string, we can always add custom data to private builds we give to customers.

shaleman:
LGTM

erikh:
Before merging:

* Needs a rebase
* I would like it if we use the much shorter/more correct code in the docker repository than the shell script that we're using now.
* `String()` support would be very nice, but not necessary. We could also always make `Print()` use the string method to automatically print things, if that's an additional selling point.

Makefile
```diff
@@ -37,6 +37,7 @@ checks:
 	./scripts/checks "$(PKGS)"
 
 run-build: deps checks clean
+	cd ${GOPATH}/src/github.com/contiv/netplugin && version/generate_version  && \
```
mapuri: this is minor, feel free to ignore.

Not sure if you already explored this, but this looks like a good candidate for `go generate` (https://blog.golang.org/generate)

so we can add something like `//go:generate <script similar to version/generate_version>` in say an existing the `netd.go` and `netmaster.go` (and even docker and k8s plugins) files to generate their respective version logic. It's kind of nice in a way that this will happen part of compilation without extra Makefile instrumentation and keep version logic close to each daemon we produce as part of the build.

netctl/netctl.go
```diff
@@ -401,3 +402,14 @@ func dumpList(ctx *cli.Context, list []map[string]interface{}) {
 	os.Stdout.Write(content)
 	os.Stdout.WriteString("\n")
 }
+
+func showVersion(ctx *cli.Context) {
```
mapuri: looks like netctl will only return netmaster's version and not netplugin's version or the docker/k8s plugin's version. 

I am thinking about the usefulness of being able to print partial info here. Should we just return netctl's version instead? Sorry if I am missing a usecase this will solve.

version/generate_version
```diff
@@ -0,0 +1,40 @@
...
+cat > $file_name << EOF
+$UseVersionStr
+// NOTE: this file is auto-generated by PR commits
+
+package version
+
+const usedDefaultVersion = "$useDefaultVersionString"
+
+// sample values are specified here
+const (
```
mapuri: minor: could we instantiate a `Info` struct here instead of independent consts?

Makefile
jainvipin: @mapuri - was going over the unaddressed comments; very interesting, I didn't know about it.
For now I will keep the scripting because I can see what got generated in a file and was comfortable with that.

netctl/netctl.go
jainvipin: @mapuri 

> looks like netctl will only return netmaster's version and not netplugin's version or the docker/k8s plugin's version.

true; as in my comment in the high level comments for the PR; I am not addressing netplugin version for the reason that if I should reconcile and make it common across all and produce one such version.

How about if we do following:
- generate a version for contivk8s and show its version alongside `netplugin -v` output
- generate a version for netctl and show it alongside `netctl version`
- I am not sure if we should show versions of all netplugins in `netctl version` output; note that incompatible versions can work alongside as long as API versions match. Should we continue discussing the versioning mismatch between netmaster and netplugin as we discuss how to version the APIs. 

WDYT @shaleman, @erikh, @unclejack @jojimt

version/generate_version
jainvipin: @mapuri 

> minor: could we instantiate a Info struct here instead of independent consts?

I wanted to do that, unfortunately there is no const structure in Go from what I understand. Please let me know if you know about a way to do this.
@erikh, @shaleman, @unclejack - would you know?

netctl/netctl.go
jojimt: I think each executable should include (only) its own version because that is really the only source of truth.

version/generate_version
mapuri: oh I just meant define a var, no need for consts. Defining this var eliminates need for `Get()` as anyways we are just generating one version, we don;t need a way to instantiate it multiple times.

Also note that we really don't need fields in Info struct to be public so we can make them lower case and avoid anyone trying to access them directly. 

```
var MyVersion = Info {
  gitCommit:  "$GitCommitStr"
  version: "$VersionStr"
  buildTime: "$BuildTimeStr"
}
```

netctl/netctl.go
jainvipin: agree; we could add the `k8contiv -version` command separately, and same for `netctl version` however the main reason for someone to issue `netctl version` was to show the netmaster/netplugin version, so perhaps keep it under the heading `netmaster's verison` and then show netmaster version

version/generate_version
```diff
@@ -0,0 +1,40 @@
...
+# generates the version of file based on current git version
+
+cd $GOPATH/src/github.com/contiv/netplugin/version
+
+USE_RELEASE=$1
+
+if [ -z "${USE_RELEASE}" ]
+then
+  file_name=version_default.go
+  UseVersionStr="// +build !use_version"
```
mapuri: I don't see the `use_version` build tag being used in the patch. Did you want to add them to the `make build` and/or `make release` targets in some way?

version/generate_version
jainvipin: @mapuri - if generated str is not in the same file as version.go then the field names would be capitals so I can access them from a Print routine. In which case someone can modify the variables.
I am not sure if you are suggesting that the script does sed 's/\$GitCommitStr/<something>' on an existing file... but that results into the file that is not in .gitignore getting modified.

version/version.go
jainvipin: yes @erikh - I incorporated these changes; can you PTAL?

> I'm @erikh :P

version/generate_version
jainvipin: I guess we can still access them in the same package... so ignore my comment. The only catch is that some other file in the package doesn't end up modifying it.

jainvipin:
PTAL @jojimt, @mapuri, @erikh , here is the how various outputs look like:
```
$ netctl -version
netctl version 
Version: 0.1
GitCommit: 7b331b8
BuildTime: 2016-01-22T03:53:46.991821091+00:00

$ netplugin -version
INFO[0000] Calling consul init                          
INFO[0000] Calling skydns2 init                         
INFO[0000] Got host addr: 192.168.2.10                  
Version: 0.1
GitCommit: 7b331b8
BuildTime: 2016-01-22T03:53:46.991821091+00:00

$ netmaster -version
Version: 0.1
GitCommit: 7b331b8
BuildTime: 2016-01-22T03:53:46.991821091+00:00

$ netctl version
Client Version:
Version: 0.1
GitCommit: 7b331b8
BuildTime: 2016-01-22T03:53:46.991821091+00:00

Server Version:
Version: 0.1
GitCommit: 7b331b8
BuildTime: 2016-01-22T03:53:46.991821091+00:00
```

netctl/netctl.go
jainvipin: @jojimt, @mapuri - added versions for k8splugin, and netctl independently; also the output of netctl version shows the client/server(netmaster) versions together.

the pending item from above discussion is really netplugin version; I suggest we skip that for this PR until we start putting versions for APIs

mapuri: SGTM

I am still not sure if showing server version is very helpful and needed. Unlike docker we have different binaries altogether so each binary having their own --version flag shall suffice. But this might be just me.

version/generate_version
mapuri: sure, sgtm. the current changes.are just fine.

netctl/netctl.go
jainvipin: > I am still not sure if showing server version is very helpful and needed. Unlike docker we have different binaries altogether so each binary having their own --version flag shall suffice. But this might be just me.

`netctl version` exercises the REST endpoint to query the version information.

erikh:
LGTM

netctl/netctl.go
mapuri: I know that, but I am not sure if we need this added feature right now :)

I would rather add version flag to each binary first and then figure/discuss usecases for having versions visible at one place for contiv-network service as a whole than just netmaster.

jainvipin: > I know that, but I am not sure if we need this added feature right now :)

IMHO it is needed because without API versioning this is your best way to compare the netctl/netmaster compatibility.

mapuri: but we have not added API version yet (or to be specific objmodel version), right? 

Also showing the version here doesn't ensures compatibility. Those compat checks would need to happen in the client/server protocol itself and need to be designed.

Anyways I am not arguing we don't need versioning or something similar. I am just saying whatever we are adding right now (on top of a basic version flag) is not sufficient/complete as is since we have not figured the things like compat checks etc yet.

jainvipin: No we haven't addedd the API version yet.

Btw, I wasn't suggesting that showing version would ensure compatibility, rather a way to compare if things are compatible. I think it will come down to issuing two commands and comparing them, vs netctl version.

If folks feel strongly that 'netctl version' is an overkill, I can resubmit the PR removing the code for `netctl version`. If we don't need `netctl version` I can also simplify the Get and Print functions in version.go as one i.e. do not expose Print at all, Get will can give a formatted string. Please let me know.

mapuri: +1 for starting simple

erikh: Please call it `String()`. Go automatically uses this method when handling `%v` and all sorts of stuff.

Other than that, I am not opposed to moving forward with this patch as-is. Handling API version would require a lot more code, and I think is out of scope at this point.

jainvipin: @erikh - introduced String(); thanks for the hint
@mapuri - simplified some code - PTAL; however given that we'd eventually need a version REST endpoint, I am not inclined to taking that code out.

Let's discuss API versioning separately...

mapuri: We could discuss the goals of versioning binaries, APIs, object-model in more detail and separately. And I am completely ok if we wish to go ahead and merge this as well.

But just for completeness, let me summarize how I see the features current patch is trying to bring in; value they add and potential missing things that we are all agreeing to fix somehow in future. Please correct me if I am missing something here:
- bring in per binary `--version`
  - this is immediately useful as right now we don't have a way to know versions for our binaries once they are downloaded and unzipped from github release.
- bring in an extra REST endpoint in netmaster for getting it's binaries version
  - this possibly brings in a value to dump netmaster's version from single place
  - but it is not complete, in many ways:
    - it doesn't address something similar for netplugin and k8s plugins.
    - it doesn't deal with multiple instances of netmaster itself
    - it is not useful for compatability as well since it doesn't talks about what API versions are being used by the client and server
    - we are adding a new endpoint now (ironically for bringing in versions), without addressing what happens when we decide to change contents of `version.Info` structure itself.

version/generate_version
mapuri: @jainvipin , may be this comment got overlooked. Can you kindly address how you plan to use `use_version` build label and if it needs to go in this patch?

netctl/netctl.go
mapuri: btw please don't take it personally if any of the above came out too strong or aggressive. I just wish that we consider all the sides and make sure that we introduce new features in a maintainable fashion.

jainvipin: @mapuri - no, I am going to take it personally :-)

Jokes apart, I appreciate your comments! Let me see if we can conclude on the right balance that we need now.

> it doesn't address something similar for netplugin and k8s plugins.

If I am a user I'd use this command to find out things that I'd care about. Typically, user would care about the API version (to know what version an API is), Netmaster version (to know if certain features are available, etc.), and a client version (which can tell if the API that we document is supported by a given client version). Netplugin and k8s version are really for debugging, because they have no user implication. This is to say that if we have incompatible netplugin and netmaster version then we should go to lowest common denominator of API version, or disallow netplugin from them coming up and form some form of fully functional adjacency. What I thought I was doing was the user aspect of things - of course without introducing API version, because it requires a different set of changes. 

Having said that, showing netplugin version here, require `host` object that I have to wait for @abhinandanpb's changes for.

> it doesn't deal with multiple instances of netmaster itself

Could we do this as a part of doing active-active redundancy model for netmaster? 

> It is not useful for compatability as well since it doesn't talks about what API versions are being used by the client and server

Versions are to know what features are being used, compatibility is with API versioning. I felt we could address the compability with API versioning, while addressing the feature-support with this versioning

> we are adding a new endpoint now (ironically for bringing in versions), without addressing what happens when we decide to change contents of version.Info structure itself.

Ideally the first endpoint to be added is version, however this is needed before official 0.1 release, hence this PR. About changing the Version Info structure, we deal with the same way we modify the rest. Provide with API version in the version object so a client can know what language to talk to with here.

This discussion and thoughts will become handy regardless of whether we make the changes now or later. 

Thoughts?

mapuri: @jainvipin 

Thanks for your elaborate reply. From your response, I think we are on same page wrt what this PR doesn't take cares of and issues that will still need to be tracked/addressed. And I am fine leaving them open for now.

Last question, what would we miss if for 0.1 release we were to let user issue `netmaster --version` command on the host where netmaster is running? Is it just the ease of issuing the command from any host?

jainvipin: @mapuri - no worries; appreciate your diligence...

> Last question, what would we miss if for 0.1 release we were to let user issue netmaster --version command on the host where netmaster is running? Is it just the ease of issuing the command from any host?

It is the assurance to the user that even if he uses 0.1 version of netmaster/netctl, we can flag the mismatch regardless of the versions of binary used here. I am of course assuming that there is no guarantee that person has root access to the host where he could run `netmaster -version` to know this, rather the user only has REST access to the netmaster to manipulate/query it.

version/generate_version
jainvipin: @mapuri - missed this one.
`use_version` would need to be triggered when doing `make release`; ideally when we start making fixed release tags. For now I will make the release tag consistent with `-version` option output. Will make this fix now.

Issue #236, opened 2016-01-17
Contiv/K8s -- build cluster using pre-built K8s and Contiv binaries


jojimt:
The current ansible uses locally built binaries for k8s and contiv -- the latter mainly because changes were needed to the binaries themselves. Need to modify this so the pre-built binaries are used.

PR #246, opened 2016-01-21
Build cluster using k8s and contiv binaries from release repos

jojimt:
This PR mainly addresses https://github.com/contiv/netplugin/issues/236. Additional changes are 1) Enable -v option for unit tests in order to be able to debug sporadic failures 2) Add locks in the endpoint delete path 3) Allow AddPod without an epg.

### DIFF
--- a/mgmtfn/k8splugin/README.md
+++ b/mgmtfn/k8splugin/README.md
@@ -8,6 +8,10 @@ Networking and Policy can be used for Pod inter-connectivity in a Kubernetes clu
 This step-by-step procedure will guide you through a minimal experience of creating
 a Kubernetes cluster with Contiv networking and applying policy between pods.
 
+Before starting, please be sure to set http/https proxies if your network requires it.
+*(Note that https_proxy should be set to point to a http:// URL (not https://).
+This is an ansible requirement.)*
+
 #### Step 1: Clone contrib and netplugin repos
 
 ```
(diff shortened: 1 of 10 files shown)


mgmtfn/k8splugin/README.md
```diff
@@ -8,35 +8,18 @@ Networking and Policy can be used for Pod inter-connectivity in a Kubernetes clu
...
-```
 $ cd ~/go/src/github.com/k8s
 $ git clone https://github.com/contiv/netplugin
-
-$ cd netplugin
-$ make run-build
 ```
 
-#### Step 4: Set proxies if your network requires http/https proxy.
+#### Step 2: Set proxies if your network requires http/https proxy.
```
shaleman: IMHO, this step can be just a note instead of a step. Not everyone will have a http proxy

shaleman:
LGTM

mgmtfn/k8splugin/kubeClient.go
```diff
@@ -86,7 +86,7 @@ func (p *podInfo) setDefaults(ns, name string) {
 	p.name = name
 	p.labels["tenant"] = "default"
```
jainvipin: Do we want to name the labels as 'io.contiv.tenant', and similarly 'io.contiv.network', etc.?
This could avoid potential conflicts...

jojimt: Right, will do that.

jainvipin:
LGTM

PR #247, opened 2016-01-22
Vlan policy and resource changes

shaleman:
This change adds support for policy enforcement in Vlan mode and moves Vlan/Vxlan resource management to global level. There are also misc cleanups bundled with this change:

Summary of changes:
- Stop generating contivmodel after importing it
- Stop assuming eth2 is Vlan uplink interface
- Create of net agent for both Vlan mode and Vxlan mode(Actual policy changes are in ofnet repo. See PR on ofnet repo)
- Delete Bunch of unused code 
- Move gstate and resource under netmaster hierarchy since they are used by net master only
- Add netctl commands to set global object
- Manage vlan and vxlan Ids as global resources. Decouple local Vlan-id used by vxlan OVS from user configured Vlan-ids since they are on different switches.
- Eliminate Auto subnet pool management. User has to specify subnets with each network.
- Make --pkt-tag an optional parameter to network. Auto allocate from global pool when not specified, use specific value when specified.
- Disable installing Vlan policies on hosts in ACI mode
- Perform all netplugin initialization after parsing the command line flags.
- Perform policy system tests in both Vxlan mode and Vlan mode

Diff is too big to see on github. Here is the diff excluding Godep changes: https://gist.github.com/shaleman/91c9302681c9a54b243d


jojimt:
Diffs can't be viewed because of the godeps :(

shaleman:
I posted a link to gist which contains non-godep changes to the PR.

-Sukhesh


On Jan 21, 2016, at 10:24 PM, jojimt <notifications@github.com<mailto:notifications@github.com>> wrote:


Diffs can't be viewed because of the godeps :(

-

jojimt:
+	// Wait for a while for OVS switch to disconnect/connect to ofnet agent
+	time.Sleep(time.Second)
+	sw.ofnetAgent.WaitForSwitchConnection()

Nit. I think this does not belong in AddUplinkPort, rather some where up in the calling path.

+func (d *OvsDriver) getIntfName() (string, error) {
+	// get the next available port number
+	for {

I think we need a reasonable upper bound for this loop. Especially as we are relying on the netlink lib here. I also think we need a better way to allocate the port name/number. However, I can't think of one now.

+		// Pick next port number
+		d.oper.CurrPortNum++
+		intfName := fmt.Sprintf("vport%d", d.oper.CurrPortNum)
+
+		// check if the port name is already in use
+		_, err := netlink.LinkByName(intfName)
+		if err != nil {

Need to check for "not-found" here. 

+			// save the new state
+			err = d.oper.Write()
+			if err != nil {
+				return "", err
+			}
+			return intfName, nil
+		}
+	}
 }

 
+	// Add master to vlan and vxlan datapaths
+	err = d.switchDb["vlan"].AddMaster(node)
+	if err != nil {
+		return err
+	}

Why is this needed in switchDb?

shaleman:
Thanks Madhav.

Please see inline

jojimt:
When Joji comments, why does Sukhesh thank Madhav? ;)

``
sukhesh>> Switchdb has two OVS switch instances one for Vlan and one for Vxlan. We are just sending AddMaster() call to both OVS instances. There is a bit of long story here which is probably best discussed in person. We are basically creating two instances of ofnet agent and informing both of them about the master.
``
I also see the master being added to ofnet agent while we explicitly add to "vlan" in the DB. Anyway, we can discuss in person.

mapuri:
I have earned a bad reputation I guess, the mail client defaults all comments to be from me instead of checking up github :smile:

shaleman:
Folks, if there are no further comments, i'd like to merge this today.

jojimt:
Sounds good.

jainvipin:
@shaleman - finally found some time to go over the changes; some comments (may be unrelated to the core of your changes, but putting them here as I see the diffs).  
- `useVethPair      = true`: could we set this to false for K8s? If this improves efficiency? @jojimt 
- some commented out code in ovsdriver.go
- REST hierarchy is not in a common place:
```
+	baseGlobal          = "/contiv.io/"
+	cfgGlobalPrefix     = baseGlobal + "config/global/"
```
- gstate has a version field, perhaps we can take it out and think about introducing it when we do API versioning
- question: is `vlan` or `vxlan` id allocation automatic when user doesn't specify it?
- are we still using old style `desired cfg json` in the testing?
- once checked in I will test out the code to see if errors are clean/crisp

Good work!

jojimt:
```
useVethPair = true: could we set this to false for K8s? If this improves efficiency?
```
I don't think there is anything k8s specific here. Regardless, I doubt there is any difference efficiency-wise between veth pair and the internal interface. Something to check out I guess.

jainvipin:
@jojimt - we can't use this in docker because docker renames the interface to eth0, if k8s does that too then we can't reuse it, however my take was that it wouldn't, therefore we can gain back some performance improvement for those, not as much though from what @shaleman had measured earlier. 

Again a separate PR but discussing it here.

jojimt:
Yes, k8s also renames to eth0.

shaleman:
> some commented out code in ovsdriver.go

I removed those commented out code in ovsdriver.go. I'll post the updated diff once i rebase.

> REST hierarchy is not in a common place:
> +   baseGlobal          = "/contiv.io/"
> +   cfgGlobalPrefix     = baseGlobal + "config/global/"

Hmm. i'll make it same as other config state.

> gstate has a version field, perhaps we can take it out and think about introducing it when we do API versioning

Sure, i'll remove the version.

> question: is vlan or vxlan id allocation automatic when user doesn't specify it?

Yes, we allocate one from the global pool if its not specified.

> are we still using old style desired cfg json in the testing?

Yes, net master tests are still using old style config. These will be moved to REST based testing in next PR.

shaleman:
I've incorporated all the review comments. Updated diffs are at: https://gist.github.com/shaleman/91c9302681c9a54b243d
I'm going to merge this PR once sanity passes

Issue #249, opened 2016-01-23
Sporadic UT failure on contivk8s


jojimt:
Unit test fails on contivk8s occasionally in sanity.

To reproduce, run the unit-test repeatedly, e.g. from a script.

This is from the error log:
```
INFRA_CONTAINER_ID=8ec72deca647bfa60a4b815aa735c87de859b47e872828586749b9d852af1f49  NETNS
=98765
FATA[0000] EP create failed -- Post http://localhost/ContivCNI.AddPod: dial unix /run/con
tiv/contiv-cni.sock: connect: no such file or directory  NETNS=98765
```

jainvipin:
@jojimt - can you please also add some details about the failure? This if someone were to pick up the issue we have some hints on how to reproduce it? Say just run unit-tests, or system-tests make target?

jojimt:
Yes, PR is on the way :)

jainvipin:
:+1:

PR #251, opened 2016-01-23
Fix spurious UT failure

jojimt:
This fixes https://github.com/contiv/netplugin/issues/249

The failure was caused by a timing issue where the server was not ready by the time posts were issued.

### DIFF
--- a/mgmtfn/k8splugin/contivk8s/k8s_cni_test.go
+++ b/mgmtfn/k8splugin/contivk8s/k8s_cni_test.go
@@ -25,9 +25,11 @@ import (
 	"net"
 	"net/http"
 	"os"
+	osexec "os/exec"
 	"strconv"
 	"strings"
 	"testing"
+	"time"
 )
 
 const (
@@ -161,6 +163,17 @@ func setupTestServer() {
 		l.Close()
 		logger.Infof("k8s test plugin closing %s", driverPath)
 	}()
+
+	// make sure the listener is ready before returning
+	for count := 0; count < 5; count++ {
+		_, err := osexec.Command("ls", driverPath).CombinedOutput()
+		if err == nil {
+			return
+		}
+		time.Sleep(time.Second)
+	}
+
+	logger.Fatalf("Listener not ready after 5 sec")
 
 }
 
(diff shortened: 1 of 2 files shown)


jainvipin:
LGTM (except some minor questions)
some of the other sanity tests are failing because of this; need to merge in sooner.

mgmtfn/k8splugin/contivk8s/k8s_cni_test.go
```diff
@@ -162,6 +164,17 @@ func setupTestServer() {
 		logger.Infof("k8s test plugin closing %s", driverPath)
 	}()
 
+	// make sure the listener is ready before returning
+	for count := 0; count < 5; count++ {
```
jainvipin: very minor: feel free to go ahead without any of this change
Should we use `go chan` instead of active waiting; however, this is a test case so not a concern

scripts/unittests
```diff
@@ -58,6 +58,10 @@ if ${run_in_vagrant}; then
     (CONTIV_NODES=1 vagrant ssh netplugin-node1 -c 'sudo -E PATH=$PATH $GOSRC/github.com/contiv/netplugin/'${0})
     ret=$?
     if [ ${ret} -ne 0 ]; then
+        if [ -f /var/log/contivk8s.log ]; then
```
jainvipin: question: is this log produced only upon an error?

mgmtfn/k8splugin/contivk8s/k8s_cni_test.go
jojimt: Because this is a one time thing and we are unlikely to hit the "sleep" part of it, I am not sure of the value it brings. But, I agree that using chan will help the wait to be optimized.

scripts/unittests
jojimt: It is also produced for a PASS.

mgmtfn/k8splugin/contivk8s/k8s_cni_test.go
shaleman: Other way of fixing this might be to do ListenUnix inline in setupTestServer() and http.Serve() in a go routine.
That guarantees socket i.e. ready before you execute the tests

scripts/unittests
shaleman: Philosophically, I'm not sure if unit tests should write the log files.
Generally, setting up log files should happen outside the scope of code being unit tested.

shaleman:
LGTM other than the minor comment on log file..

Issue #256, opened 2016-01-29
VTEP missing on some nodes


jojimt:
In an 8 node cluster, one node reports "Could not find the VTEP for endpoint" error.
From the logs, it is clear that the corresponding peer did not get added to that node. Seems to be a boot up timing issue. Will get a PR out with a fix soon.

PR #258, opened 2016-01-29
Fix https://github.com/contiv/netplugin/issues/256

jojimt:
When multiple nodes start up almost at the same time, it is possible for the netplugin to miss some peer registrations in the real-time window between a service state read and the watch which occurs in a different thread. This fix forces the WatchService routine to inject the current state of the service to the listening channel and utilizes that for the initial state sync, instead of separately reading the state.


shaleman:
LGTM

PR #262, opened 2016-02-02
Netmaster changes

shaleman:
Various net master changes:

 - Separate out docker network code into separate package
    - Store docker network to UUID mapping as docknet state and use that to derive network/service names during lib network API calls(So that we don't need to query docker for this info). 
    - Add unit test for docker network APIs
 - Change rules to take from/to endpoint-groups instead of just end-point-group for clarity.
    - Make corresponding contiv model and netctl CLI changes
    - Change 'accept' action to 'allow' action
 - Change netctl to use auto-generated contiv-model go-client
 - Add unit test for net master REST apis.
 - Fix minor bugs found while adding unit test for netmaster REST apis


shaleman:
Folks, can someone review this change?

shaleman:
I'm going to rebase and merge this PR if there are no further comments

jojimt:
LGTM

jainvipin:
LGTM

PR #263, opened 2016-02-02
Misc fixes to k8s demo files.

jojimt:
1) Edited README to add pre-requisites
2) Invoke scripts with ./ instead of relying on PATH
3) Modified policy.sh script to handle new netctl syntax.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,11 +62,11 @@ start: update
 
 #kubernetes demo targets
 k8s-cluster:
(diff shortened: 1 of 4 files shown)


jojimt:
@jainvipin Please take a look at the updated diff. If looks ok, I will merge.

jainvipin:
LGTM @jojimt

jojimt:
@vijayvikrant I guess you commented on the previous version of the diff. I had kept it for backwards compatibility, but removed it now based on feedback from @jainvipin.

PR #264, opened 2016-02-03
steps to add/update dependencies in netplugin using godep

vijayvikrant:
godep documentation for netplugin - based on the recent experience of adding and updating dependencies in netplugin


jainvipin:
:+1: 
Good doc; highly needed...

shaleman:
LGTM
Good doc. Thanks @vijayvikrant.

PR #265, opened 2016-02-03
netctl cli to take in router ip 

abhinandanpb:
The changes contain:
1) Changes to netctl cli to take in router ip and as as additional parameters
2) This would mean there is only 1 bgp config cli to be used to configure L3 mode.
3) decoupled bgp server start from netplugin start and tied it to the cli.
4) This wil ensure bgp config can be changed dynamically even after containers come up . 

new cli  netctl bgp create <hostname> -router-ip=<> -as=<> -neighbor-ip=<> neighbor-as=<>

### DIFF
--- a/drivers/fakenetepdriver.go
+++ b/drivers/fakenetepdriver.go
@@ -60,12 +60,12 @@ func (d *FakeNetEpDriver) DeleteMaster(node core.ServiceInfo) error {
 	return core.Errorf("Not implemented")
 }
 
-// AddBgpNeighbors is not implemented.
-func (d *FakeNetEpDriver) AddBgpNeighbors(id string) (err error) {
+// AddBgp is not implemented.
+func (d *FakeNetEpDriver) AddBgp(id string) (err error) {
 	return core.Errorf("Not implemented")
 }
 
-// DeleteBgpNeighbors is not implemented.
-func (d *FakeNetEpDriver) DeleteBgpNeighbors(id string) (err error) {
+// DeleteBgp is not implemented.
+func (d *FakeNetEpDriver) DeleteBgp(id string) (err error) {
 	return core.Errorf("Not implemented")
 }
--- a/drivers/ovsdriver.go
+++ b/drivers/ovsdriver.go
@@ -145,7 +145,7 @@ func (d *OvsDriver) Init(config *core.Config, info *core.InstanceInfo) error {
 	}
 	// Create Vlan switch
 	d.switchDb["vlan"], err = NewOvsSwitch(vlanBridgeName, "vlan", info.VtepIP,
-		info.FwdMode, info.RouterIP, info.VlanIntf)
+		info.FwdMode, info.VlanIntf)
 	if err != nil {
 		log.Fatalf("Error creating vlan switch. Err: %v", err)
 	}
@@ -433,7 +433,7 @@ func (d *OvsDriver) DeleteMaster(node core.ServiceInfo) error {
 }
 
 // AddBgpNeighbors adds bgp neighbor by named identifier
-func (d *OvsDriver) AddBgpNeighbors(id string) error {
+func (d *OvsDriver) AddBgp(id string) error {
 	var sw *OvsSwitch
 
 	cfg := mastercfg.CfgBgpState{}
@@ -449,17 +449,17 @@ func (d *OvsDriver) AddBgpNeighbors(id string) error {
 	// Find the switch based on network type
 	sw = d.switchDb["vlan"]
 
-	return sw.AddBgpNeighbors(cfg.Hostname, cfg.As, cfg.Neighbor)
+	return sw.AddBgp(cfg.Hostname, cfg.RouterIP, cfg.As, cfg.NeighborAs, cfg.Neighbor)
 }
 
 // DeleteBgpNeighbors deletes a bgp neighbor by named identifier
-func (d *OvsDriver) DeleteBgpNeighbors(id string) error {
+func (d *OvsDriver) DeleteBgp(id string) error {
 	log.Infof("delete Bgp Neighbor %s \n", id)
 	//FixME: We are not maintaining oper state for Bgp
 	//Need to Revisit again
 	// Find the switch based on network type
 	var sw *OvsSwitch
 	sw = d.switchDb["vlan"]
-	return sw.DeleteBgpNeighbors()
+	return sw.DeleteBgp()
 
 }
--- a/drivers/ovsdriver.go
+++ b/drivers/ovsdriver.go
@@ -432,7 +432,7 @@ func (d *OvsDriver) DeleteMaster(node core.ServiceInfo) error {
 	return nil
 }
 
-// AddBgpNeighbors adds bgp neighbor by named identifier
+// AddBgp adds bgp config by named identifier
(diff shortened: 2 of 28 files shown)


erikh:
"BGP" should be capitalized in help and error text.

abhinandanpb:
@erikh . sure will keep the description consistent for add,rm,list.

abhinandanpb:
@vijayvikrant . as discussed have added the flag. Thanks for taking a note of it.

drivers/fakenetepdriver.go
```diff
@@ -60,12 +60,12 @@ func (d *FakeNetEpDriver) DeleteMaster(node core.ServiceInfo) error {
 	return core.Errorf("Not implemented")
 }
 
-// AddBgpNeighbors is not implemented.
-func (d *FakeNetEpDriver) AddBgpNeighbors(id string) (err error) {
+// AddBgp is not implemented.
+func (d *FakeNetEpDriver) AddBgp(id string) (err error) {
 	return core.Errorf("Not implemented")
```
jainvipin: reminds me that the unit tests for all this code is missing; more importantly system tests are missing too...
let's talk offline on how to incorporate testing of all this.

drivers/ovsdriver.go
```diff
@@ -432,8 +432,8 @@ func (d *OvsDriver) DeleteMaster(node core.ServiceInfo) error {
 	return nil
 }
 
-// AddBgpNeighbors adds bgp neighbor by named identifier
-func (d *OvsDriver) AddBgpNeighbors(id string) error {
+// AddBgp adds bgp config by named identifier
+func (d *OvsDriver) AddBgp(id string) error {
```
jainvipin: curious why is this code in ovsdriver.go
I mean we should avoid keeping non driver logic in ovsdriver.go; this would help other drivers to leverage this code e.g. @danehans request to do ipvlan driver

netctl/commands.go
```diff
@@ -292,40 +292,48 @@ var Commands = []cli.Command{
...
+				Aliases:   []string{"delete"},
+				Usage:     "Delete BGP configuration",
+				ArgsUsage: "[hostname]",
+				Flags:     []cli.Flag{},
+				Action:    deleteBgp,
+			},
+			{
+				Name:      "create",
+				Usage:     "Add BGP configuration.",
+				ArgsUsage: "[hostname]",
```
jainvipin: will need to create bash auto-completion for some of these commands as well.

drivers/ovsdriver.go
abhinandanpb: @jainvipin AddBgp would instantiate a protocol server in ofnet only on vlan mode.So we have to pipe it to the right ofnetAgent for this to happen.

netmaster/objApi/apiController.go
```diff
@@ -880,6 +882,6 @@ func (ac *APIController) BgpDelete(bgpNeighborCfg *contivModel.Bgp) error {
 }
 
 //BgpUpdate updates bgp config
-func (ac *APIController) BgpUpdate(oldbgpNeighborCfg *contivModel.Bgp, NewbgpNeighborCfg *contivModel.Bgp) error {
+func (ac *APIController) BgpUpdate(oldbgpCfg *contivModel.Bgp, NewbgpCfg *contivModel.Bgp) error {
 	return nil
```
jainvipin: why is 'update' returning 'nil'? Do we need to retrigger the updates and retrigger grat-arp, etc. to update this?

drivers/ovsdriver.go
```diff
@@ -449,17 +449,17 @@ func (d *OvsDriver) AddBgpNeighbors(id string) error {
...
-	return sw.AddBgpNeighbors(cfg.Hostname, cfg.As, cfg.Neighbor)
+	return sw.AddBgp(cfg.Hostname, cfg.RouterIP, cfg.As, cfg.NeighborAs, cfg.Neighbor)
 }
 
-// DeleteBgpNeighbors deletes a bgp neighbor by named identifier
-func (d *OvsDriver) DeleteBgpNeighbors(id string) error {
+// DeleteBgp deletes bgp config by named identifier
+func (d *OvsDriver) DeleteBgp(id string) error {
 	log.Infof("delete Bgp Neighbor %s \n", id)
 	//FixME: We are not maintaining oper state for Bgp
```
jainvipin: Can you elaborate why is oper state required for this?

jainvipin:
@abhinandanpb - good set of changes; taking bgp configuration from intent is the right direction; also like the consistency for various CLIs (rm/ls/create/etc.)
please identify the changes needed for this PR and work on the remaining ones independently.

netctl/commands.go
abhinandanpb: @jainvipin  yes spoke to @vvb about this . He is going to take care of it in the next commit.

netmaster/objApi/apiController.go
abhinandanpb: This is a filler function. Currently not being used. I believe this was added to keep compiler from not complaining about non implementation of the function

drivers/ovsdriver.go
abhinandanpb: Well it is not needed with the current support hence not maintained. I made a note to see if would need it in cases of bgp restartbility scenarios without restarting netplugin. Havent gotten a chance to walk through the cases.

jainvipin: ok, 'FIXME' suggests something known is not coded.

> Well it is not needed with the current support hence not maintained. I made a note to see if would need it in cases of bgp restartbility scenarios without restarting netplugin. Havent gotten a chance to walk through the cases.

drivers/ovsdriver.go
jainvipin: I see; then the question really is if someone uses non ovs (i.e. non ofnet), would we need any part of the protocol server outside (not in ofnet), we can take this discussion offline perhaps as we discuss the changes needed for other drivers.

> @jainvipin AddBgp would instantiate a protocol server in ofnet only on vlan mode.So we have to pipe it to the right ofnetAgent for this to happen. In addition to this it will create loopback interface, add bgp endpoints etc.

netctl/commands.go
jainvipin: sounds good @abhinandanpb

netmaster/objApi/apiController.go
jainvipin: got it

> This is a filler function. Currently not being used. I believe this was added to keep compiler from not complaining about non implementation of the function

jainvipin:
LGTM @abhinandanpb - let us discuss the sanity tests for this to move to VM environment offline...

abhinandanpb:
@jainvipin  sounds good. Thanks for reviewing.

netmaster/objApi/apiController.go
shaleman: Well, its not exactly filler function :)
Its there to handle update operation on an existing REST object.
We need to start handling update on all objects. 
You can take this up in next PR..

shaleman:
LGTM
Lets discuss unit/system tests and update handling offline..

abhinandanpb:
@shaleman sure will incorporate the changes needed in the next PR.

Issue #255, opened 2016-01-29
allow runtime flag to avoid running skydns


jainvipin:
Allow using external (or pre-built dns) from libnetwork/k8s

`netmaster -dns-service=false . . .`

Desired behavior could be:
- if `dns-service` is specified, either `true` or `false` it is used as a hint to mandate or bypass dns service by netplugin.
- Otherwise, apply the behavior as exists now i.e. when dockplugin is use dns-service, otherwise (e.g. for k8s) skip

shaleman:
Closing.
fixed by #268

PR #268, opened 2016-02-03
Fix DNS attach error message

DivyaVavili:
This fixes the error message thrown when attaching DNS to network fails and makes it more intuitive to users on what is failing.

### DIFF
--- a/netmaster/master/network.go
+++ b/netmaster/master/network.go
@@ -301,7 +301,9 @@ func attachServiceContainer(tenantName, networkName string, stateDriver core.Sta
 	if err != nil {
 		log.Errorf("Could not attach container(%s) to network %s. Error: %s",
 			contName, dnetName, err)
-		return err
+		return fmt.Errorf("Could not attach container(%s) to network %s."+
+			"Please make sure %s container is up.",
+			contName, dnetName, contName)
 	}
 
 	// inspect the container


jainvipin:
@DivyaVavili - how does the new error message look like?

DivyaVavili:
After the change, it looks like this. 

vagrant@netplugin-node1:~$ netctl net create nw1 -p 1002 -s 10.1.2.0/24 -g 10.1.2.254
Could not attach container(defaultdns) to network nw1.
ERRO[0000] Status 500 in request response

Error message before the change:
vagrant@netplugin-node1:~$ netctl net create nw1 -p 1002 -s 10.1.2.0/24 -g 10.1.2.254
Error: 500 Internal Server Error: Container 0f837a8b0b044d16370ad64f68ff3d167a16f563354fab1ef4f2a39d7299e877 is not running

jainvipin:
@DivyaVavili - minor, how about:
vagrant@netplugin-node1:~$ netctl net create nw1 -p 1002 -s 10.1.2.0/24 -g 10.1.2.254
Could not attach container(defaultdns) to network nw1. Please make sure defaultdns container is up.

ERRO[0000] Status 500 in request response  <== if we can suppress this, even better.

DivyaVavili:
@jainvipin: The Status 500 message seems to be returned from docker which we might not have control over. I have made rest of the change(s) as suggested.

shaleman:
@DivyaVavili Changes itself looks fine.
but, instead of reporting error when connect fails, wouldn't it be better if we checked the status of the container and started it if its not running?
Since netmaster manages these containers, its not fair to ask user to make sure its up.
I'm ok if this is coming in a separate PR.

jainvipin:
@shaleman - this would go back to `what if the image doesn't exist`;
How about we just move on without dns/service-discovery in that case, when there is a problem with starting dns container...

shaleman:
@jainvipin sounds fine to disable service discovery if skydns image doesn't exist or we are not able to pull skydns image

DivyaVavili:
@shaleman @jainvipin: Sure, SGTM. I will handle these comments as part of resolving issue #255.

PR #269, opened 2016-02-05
Skydns refactor

DivyaVavili:
- Including an option in netmaster to deploy DNS container (default is enabled)
- Added additional checks during tenant create and network create to deal with failure cases
- To ensure HA for DNS containers, introduced a restart policy so that it is always restarted 

**Explanation of changes with output**
- When started with -dns-enable=false, creation of tenant does not start skydns container
```
vagrant@netplugin-node1:~$ netctl tenant list
Name
------
tenant1
default

vagrant@netplugin-node1:~$ docker ps
CONTAINER ID        IMAGE                          COMMAND             CREATED             STATUS              PORTS               NAMES
```
- Starting netmaster without any options, starts off skydns containers per tenant by default
```
vagrant@netplugin-node1:~$ netctl tenant list
Name
------
tenant1
default

vagrant@netplugin-node1:~$ docker ps
CONTAINER ID        IMAGE                          COMMAND             CREATED             STATUS              PORTS               NAMES
ff729f5a4130        skynetservices/skydns:latest   "/skydns"           9 minutes ago       Up 9 minutes        53/tcp, 53/udp      defaultdns
2afb234c7f92        skynetservices/skydns:latest   "/skydns"           8 minutes ago       Up 8 minutes        53/tcp, 53/udp      tenant1dns
```
- During network attach:
   - If the tenant-dns is missing
     - DNS container is started off and the attach is tried again 
       - If this fails, the network creation continues after a warning is issued
   - If the tenant-dns is in stopped state
     - A restart of the dns container is attempted
       - If this fails, network creation continues after a warning is issued
   
- When starting a skydns container:
   - If the skydns image is missing on the node 
     - A pull of skydns image is attempted
       - If the skydns pull fails, a warning is issued and tenant creation proceeds
     
- To ensure that the HA of DNS provider, the skydns container is started off with "always" restart-policy. With this policy, docker tries to bring back the skydns container whenever it fails/docker restarts.

### DIFF
--- a/netmaster/master/netmaster.go
+++ b/netmaster/master/netmaster.go
@@ -264,6 +264,7 @@ func startServiceContainer(tenantName string) error {
 	containerID, err := docker.CreateContainer(containerConfig, getDNSName(tenantName), nil)
 	if err != nil {
 		log.Errorf("Error creating DNS container for tenant: %s. Error: %s", tenantName, err)
+		return err
 	}
 
 	hostConfig := &dockerclient.HostConfig{
--- a/netmaster/master/netmaster.go
+++ b/netmaster/master/netmaster.go
@@ -38,6 +38,7 @@ const (
 // Run Time config of netmaster
 type nmRunTimeConf struct {
 	clusterMode string
+	dnsEnabled  bool
 }
 
 var masterRTCfg nmRunTimeConf
@@ -53,12 +54,29 @@ func SetClusterMode(cm string) error {
 	}
 
 	masterRTCfg.clusterMode = cm
+	SetDNSEnabled(false)
 	return nil
 }
 
 // GetClusterMode gets the cluster mode of the contiv plugin
 func GetClusterMode() string {
 	return masterRTCfg.clusterMode
+}
+
+// IsDNSEnabled gets the status of whether DNS is enabled
+func IsDNSEnabled() bool {
+	return masterRTCfg.dnsEnabled
+}
+
+// SetDNSEnabled sets the status of DNS Enable
+func SetDNSEnabled(dnsEnableFlag bool) error {
+	log.Infof("Setting dns flag to %s", dnsEnableFlag)
+	masterRTCfg.dnsEnabled = dnsEnableFlag
+	return nil
+}
+
+func getDNSName(tenantName string) string {
+	return tenantName + "dns"
 }
 
 func getEpName(networkName string, ep *intent.ConfigEP) string {
@@ -196,12 +214,12 @@ func CreateTenant(stateDriver core.StateDriver, tenant *intent.ConfigTenant) err
 		return err
 	}
 
-	if GetClusterMode() == "docker" {
+	if IsDNSEnabled() {
 		// start skydns container
 		err = startServiceContainer(tenant.Name)
 		if err != nil {
-			log.Errorf("Error starting service container. Err: %v", err)
-			return err
+			log.Errorf("Error starting service container. Err: %v. Disabling DNS option.", err)
+			SetDNSEnabled(false)
 		}
 	}
 
@@ -243,7 +261,7 @@ func startServiceContainer(tenantName string) error {
 			"SKYDNS_ADDR=0.0.0.0:53",
 			"SKYDNS_DOMAIN=" + tenantName}}
 
-	containerID, err := docker.CreateContainer(containerConfig, tenantName+"dns", nil)
+	containerID, err := docker.CreateContainer(containerConfig, getDNSName(tenantName), nil)
 	if err != nil {
 		log.Errorf("Error creating DNS container for tenant: %s. Error: %s", tenantName, err)
 	}
@@ -268,7 +286,7 @@ func stopAndRemoveServiceContainer(tenantName string) error {
 		return err
 	}
 
-	dnsContName := tenantName + "dns"
+	dnsContName := getDNSName(tenantName)
 	// Stop the container
 	err = docker.StopContainer(dnsContName, 10)
 	if err != nil {
@@ -286,15 +304,17 @@ func stopAndRemoveServiceContainer(tenantName string) error {
 
 // DeleteTenantID deletes a tenant from the state store, by ID.
 func DeleteTenantID(stateDriver core.StateDriver, tenantID string) error {
-	if GetClusterMode() == "docker" {
-		err := stopAndRemoveServiceContainer(tenantID)
+	var err error
+
+	if IsDNSEnabled() {
+		err = stopAndRemoveServiceContainer(tenantID)
 		if err != nil {
 			log.Errorf("Error in stopping service container for tenant: %+v", tenantID)
 			return err
 		}
 	}
 
-	return nil
+	return err
 }
 
 // DeleteTenant deletes a tenant from the state store based on its ConfigTenant.
--- a/netmaster/master/netmaster.go
+++ b/netmaster/master/netmaster.go
@@ -54,7 +54,6 @@ func SetClusterMode(cm string) error {
 	}
 
 	masterRTCfg.clusterMode = cm
-	SetDNSEnabled(false)
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


netmaster/master/netmaster.go
```diff
@@ -53,6 +54,7 @@ func SetClusterMode(cm string) error {
 	}
 
 	masterRTCfg.clusterMode = cm
+	SetDNSEnabled(false)
```
shaleman: Hmm, This is probably not correct. If i run netmaster -cluster-mode docker, this will disable DNS.
Plus, It'll be good to keep all flag parsing/setting logic in parseOpts() or execOpts() functions..

netmaster/master/netmaster.go
```diff
@@ -196,12 +214,12 @@ func CreateTenant(stateDriver core.StateDriver, tenant *intent.ConfigTenant) err
...
 
-	if GetClusterMode() == "docker" {
+	if IsDNSEnabled() {
 		// start skydns container
 		err = startServiceContainer(tenant.Name)
 		if err != nil {
-			log.Errorf("Error starting service container. Err: %v", err)
-			return err
+			log.Errorf("Error starting service container. Err: %v. Disabling DNS option.", err)
+			SetDNSEnabled(false)
```
shaleman: I know we discussed this. But, just realized that we should not globally disable DNS mode when start container fails for one tenant. We should probably keep track of this at per tenant level..

netmaster/master/netmaster.go
```diff
@@ -283,15 +305,17 @@ func stopAndRemoveServiceContainer(tenantName string) error {
...
+	if IsDNSEnabled() {
+		err = stopAndRemoveServiceContainer(tenantID)
 		if err != nil {
 			log.Errorf("Error in stopping service container for tenant: %+v", tenantID)
 			return err
 		}
 	}
 
-	return nil
+	return err
```
shaleman: I don't think err needs to be a function scope here. It can be just block scope.

netmaster/master/network.go
```diff
@@ -162,34 +163,55 @@ func CreateNetwork(network intent.ConfigNetwork, stateDriver core.StateDriver, t
...
+				return nil
+			}
+		}
+	}
+
+	// If it's not in running state, restart the container.
+	// This case can occur if the host is reloaded
+	if !cinfo.State.Running {
+		log.Debugf("Container %s not running. Restarting the conttainer", contName)
+		err = docker.RestartContainer(contName, 0)
```
shaleman: You might want to test this scenario. I'm not sure if RestartContainer is the right call to make.
If container is in stopped state, you might need to do just StartContainer()

netmaster/master/network.go
```diff
@@ -162,34 +163,55 @@ func CreateNetwork(network intent.ConfigNetwork, stateDriver core.StateDriver, t
...
+		if err != nil {
+			log.Warnf("Error restarting service container %s. "+
+				"Continuing without DNS provider. Error: %v",
+				contName, err)
+			return nil
+		}
+
+		// Refetch container info after restart
+		cinfo, err = docker.InspectContainer(contName)
```
shaleman: Can you check for error here?

shaleman:
Changes look fine. 
Lets sync up about globally disabling dns mode when skydns on single tenant fails..

shaleman:
@DivyaVavili Another note before i forget. Can you look into DNS server implementation in latest libnetwork? I'm wondering if we should embed lightweight DNS server into netplugin/netmaster similar to libnetwork. This might avoid lot of dependency issues with skydns.. 
We should also look at kubernetes/mesos requirements for service discovery and see how we can unify our code across all three..

jainvipin:
> I'm wondering if we should embed lightweight DNS server into netplugin/netmaster similar to libnetwork. This might avoid lot of dependency issues with skydns.. 

I am wondering if we need dns-server at all now that libnetwork/k8s/mesos have built in dns for container workloads.

netmaster/master/netmaster.go
DivyaVavili: I had put in this check so that it is explicitly disabled for kubernetes case. The DNS Enabled flag is set after SetClusterMode from execOpts. I'll remove this for consistent handling of setting of options from execOpts. 

/cc @jojimt: You might want to consider updating the flags when running in kubernetes mode

netmaster/master/network.go
DivyaVavili: I tested this with both start and restart option. It seems to be doing the same thing in this case. Are there any differences in the handling that we should watch out for?

netmaster/master/netmaster.go
DivyaVavili: Makes sense. I will add tenant level state for this.

netmaster/master/network.go
```diff
@@ -338,18 +363,20 @@ func DeleteNetworkID(stateDriver core.StateDriver, netID string) error {
...
+	if IsDNSEnabled() {
 		// detach Dns container
 		err = detachServiceContainer(nwCfg.Tenant, nwCfg.NetworkName)
 		if err != nil {
 			log.Errorf("Error detaching service container. Err: %v", err)
 		}
+	}
 
+	// Check if there are any active endpoints
+	if hasActiveEndpoints(nwCfg) {
```
DivyaVavili: /cc @vvb: With the DNS option being explicit now, we should ideally be detaching the container from the network. After this we should check if there are no active endpoints whether dns is enabled or not. Made the corresponding changes in hasActiveEndpoints function as well. Please check it out and let me know if you there are any scenarios that might break your changes.

DivyaVavili:
@jainvipin: I'm trying out the embedded DNS option with docker 1.10. Running into a couple of issues with that. Will try it out and we could move to that option as required. These changes should help till then to sort out any issues.

netmaster/master/network.go
```diff
@@ -162,34 +163,58 @@ func CreateNetwork(network intent.ConfigNetwork, stateDriver core.StateDriver, t
...
+				log.Warnf("Error starting service container. "+
+					"Continuing without DNS provider. Error: %v", err)
+				return nil
+			}
+		}
+	}
+
+	// If it's not in running state, restart the container.
+	// This case can occur if the host is reloaded
+	if !cinfo.State.Running {
```
vvb: would cinfo be what we expect if we hit "no such id" error case above and restarted the container? Do we need to inspect again?

vvb:
LGTM, other than the minor comments

netmaster/master/network.go
vvb: @DivyaVavili One slight view we might want to consider here. If active endpoints are present and we fail network delete, we might still want to have dns container lying around. as per the previous changes - in case of a failure Nothing would have changed. Now we will end up with partial cleanup..It might be better to have the ActiveEp check first. and fail it to start with.

netmaster/master/network.go
```diff
@@ -499,9 +526,5 @@ func networkReleaseAddress(nwCfg *mastercfg.CfgNetworkState, ipAddress string) e
 }
 
 func hasActiveEndpoints(nwCfg *mastercfg.CfgNetworkState) bool {
-	// Uncomment the below after https://github.com/contiv/netplugin/pull/269 is merged
-	// We spin a dns container if IsDNSEnabled() == true
-	// We need to exlude that from Active EPs check.
-	//return (IsDNSEnabled() && nwCfg.EpCount > 1) || ((!IsDNSEnabled()) && nwCfg.EpCount > 0)
-	return nwCfg.EpCount > 1
+	return nwCfg.EpCount > 0
```
vvb: @DivyaVavili might need to checked based on my earlier comment. Good to go back to the commented code.

netmaster/master/network.go
DivyaVavili: @vvb: Good point, made the changes accordingly

netmaster/master/network.go
DivyaVavili: @vvb: Yes, we need to check for the status again. Made the change

netmaster/master/network.go
```diff
@@ -162,34 +163,62 @@ func CreateNetwork(network intent.ConfigNetwork, stateDriver core.StateDriver, t
...
+		if strings.Contains(err.Error(), "no such id") {
+			// DNS container not started for this tenant. Start skydns container
+			err = startServiceContainer(tenantName)
+			if err != nil {
+				log.Warnf("Error starting service container. "+
+					"Continuing without DNS provider. Error: %v", err)
+				return nil
+			}
+			cinfo, err = docker.InspectContainer(contName)
+			if err != nil {
```
vvb: @DivyaVavili the change looks good. we might want to "return nil" in case this inspect gives an error. Not sure If we can depend on the cinfo value in case of errors.

shaleman:
LGTM
We can merge this once sanity passes..

Issue #271, opened 2016-02-05
Add endpoint failure in ACI mode


jojimt:
In ACI mode, add ep fails with the following error log:
```Jan 29 09:27:13 contiv-aci-server1 netplugin[8308]: time="Jan 29 09:27:13.728212508" level=error msg="VNI for vlan 3 is not known"
Jan 29 09:27:13 contiv-aci-server1 netplugin[8308]: time="Jan 29 09:27:13.728241311" level=error msg="Error adding local port vvport2 to ofnet. Err: Unknown Vlan"```

PR #272, opened 2016-02-06
Fix https://github.com/contiv/netplugin/issues/271

jojimt:
Picked up the fix from @shaleman. Added a sanity test case to cover ACI.


shaleman:
LGTM
Thanks for fixing unit test and adding system test

PR #273, opened 2016-02-08
Netmaster HA changes

shaleman:
This adds netmaster hot standby support.
- Netmaster tries to take an etcd lock when it starts up. The one acquiring the lock is the leader and others are followers. 
- There can be any number of netmasters in the cluster. When current leader fails, one of the followers acquire the lock and become new leader.
- Leader is responsible for making all state changes. Followers simply proxy all REST calls to current leader.
- Netplugins detect the current leader using etcd and send all REST calls to current leader.
- External clients(like netctl or web ui) can use either virtual-IP or loadbalancer to access netmaster's REST API.
- All netmasters participate in ofnet route distribution in active-active manner.

Sumary of changes:
- Restructure netmaster main package to introduce a master FSM(This includes leader lock mgmt).
- Change netplugin to send REST requests to current leader..
- Minor bug fixes in policy/rule handling.
- Add test cases to test netmaster HA and switchover.
- Add missing test cases on some of the policy features like match on IP address, match on network, ICMP etc..


netmaster/daemon.go
```diff
@@ -0,0 +1,339 @@
...
+	// Setup the router to serve the web UI
+	goPath := os.Getenv("GOPATH")
+	if goPath != "" {
+		webPath := goPath + "/src/github.com/contiv/contivmodel/www/"
+
+		// Make sure we have the web UI files
+		_, err := os.Stat(webPath)
+		if err != nil {
+			webPath = goPath + "/src/github.com/contiv/netplugin/" +
+				"Godeps/_workspace/src/github.com/contiv/contivmodel/www/"
```
jojimt: I guess this is ok for now. But, we should use a install directory such as /opt/contiv/...

shaleman: Yes. Thanks. its on my todo list to make this a command line argument instead of hard coding.
Next PR may be..

netmaster/daemon.go
```diff
@@ -0,0 +1,339 @@
...
+	d.ofnetMaster = ofnet.NewOfnetMaster(ofnet.OFNET_MASTER_PORT)
+	if d.ofnetMaster == nil {
+		log.Fatalf("Error creating ofnet master")
+	}
+
+	// Create an objdb client
+	d.objdbClient = objdb.NewClient("")
+
+	// Get the address to be used for local communication
+	localIP, err := d.objdbClient.GetLocalAddr()
```
jojimt: This means that the netmasters can run only on the nodes where etcd run in non-proxy mode. If we want to avoid this restriction, we should take a control-ip command line arg like netplugin.

shaleman: Good point.
Let me file an issue for making both of these command line options.

netmaster/daemon.go
```diff
@@ -0,0 +1,339 @@
...
+// becomeLeader changes daemon FSM state to master
+func (d *daemon) becomeLeader() {
+	// ask listener to stop
+	d.stopFollowerChan <- true
+
+	// set current state
+	d.currState = "leader"
+
+	// Run the HTTP listener
+	go d.runLeader()
```
jojimt: There is a small window during which both leader and follower are executing on the same daemon. Will that cause any catastrophic issue? Just checking.

shaleman: both leader and follower go routines try to acquire d.listenerMutex
This should protect them from running at the same time..

netplugin/cluster/cluster.go
```diff
@@ -100,8 +106,45 @@ func httpPost(url string, req interface{}, resp interface{}) error {
...
+		return "", errors.New("No leader node")
+	}
+
+	return masterNode, nil
+}
+
 // MasterPostReq makes a POST request to master node
 func MasterPostReq(path string, req interface{}, resp interface{}) error {
+	// first find the holder of master lock
+	masterNode, err := getMasterLockHolder()
```
jojimt: Don't we need a similar change to the rpc call from ofnet agent as well? Did I miss it in the diff?

netmaster/daemon.go
jojimt: It can happen while transitioning from follower to leader. We issue a stop follower before becoming leader, but the actual stop happens in another thread.

netplugin/cluster/cluster.go
shaleman: ofnet agent calls are made to all ofnet masters. i.e. ofnet master are in active-active mode.
Think ofnet masters acting as two route reflectors and all ofnet agent peering with all ofnet masters.. github.com/contiv/ofnet page has a nicer picture for it..

netmaster/daemon.go
shaleman: But, since follower does not release the listenerMutex till it closes all connections and exits the go routine, leader go routine will be just waiting for the listenerMutex. That should be okay right?

jojimt: Yes, missed that part.

shaleman:
Can i get an LGTM on this PR?

PR #278, opened 2016-02-09
updating the accept to allow change in netctl autocompletion

vvb:
action property value "accept" was renamed to "allow".


shaleman:
LGTM
Thanks!

PR #279, opened 2016-02-11
Fix a bug in deleting ovs interfaces

shaleman:
Delete veth interfaces after they are removed from OVS


drivers/ovsSwitch.go
```diff
@@ -400,6 +390,16 @@ func (sw *OvsSwitch) DeletePort(epOper *OvsOperEndpointState) error {
 		return err
```
jojimt: I think this return needs to change to an error log in order to let the clean up continue.

shaleman: Yeah. fair point. i'll change that.

jojimt: LGTM, other than that minor issue.

PR #280, opened 2016-02-11
Changes for complying with docker 1.10

DivyaVavili:
Changes include:
- Changing Accept header version
- Adding GetCapabilities API for IPAM

The actual diff(excluding the Godep changes) as such is not huge and can be found here: 
https://gist.github.com/DivyaVavili/47b89c14e2792ae4a6f7


shaleman:
@DivyaVavili Changes look good Thanks.

    router := mux.NewRouter()
-   s := router.Headers("Accept", "application/vnd.docker.plugins.v1.1+json").
+   s := router.Headers("Accept", "application/vnd.docker.plugins.v1.2+json").
        Methods("POST").Subrouter()

You might want to change the above to something like:
s := router.Methods("POST").Subrouter()

Otherwise, netplugin will not work with docker 1.9..
I'm guessing sanity failed for the same reason..

PR #281, opened 2016-02-11
scripts: use overlay graphdriver in docker startup scripts

erikh:
Watch the test times go *DOWN*


unclejack:
LGTM

shaleman:
LGTM

unclejack:
I'm merging this now. We can always revert if we run into issues.

PR #282, opened 2016-02-11
Vagrantfile: use KVM paravirt

unclejack:


shaleman:
@unclejack Change looks good to me.
Do you know if this works on MacOS?
LGTM otherwise..

PR #284, opened 2016-02-12
System tests

erikh:
This implements a number of performance enhancements for the system tests (formerly called the sanity tests).

* Written in Go instead of Python
* Parallel execution of tests (ping, connect)
* Parallel startup and teardown of containers
* Parallel listener startup
* Uses alpine image now for faster download and container startup/teardown
* Overlayfs for maximum filesystem performance

The result of this work is that you get down to 800s on a lab box, which is better than the 1h30m that was  done in the systemtests before.

The systemtests themselves are all in one commit, please click on that to review independently. I took care to re-implement in spirit the python code, so it was easier to maintain for those already familiar with the code. I also abstracted most of the parallelism into a framework that should be easy-to-use for future tests that need to be created.

I also exposed a few bugs: one @shaleman has already fixed in master, but there's a netplugin panic that seems to be happening occasionally due to ovsdb driver code. I'll get more details in an issue soon.


jojimt:
@erikh There was a TC added to cover ACI via https://github.com/contiv/netplugin/pull/272. I didn't see it covered here.

erikh:
@jojimt right now, the tests only run in the VMs, not on metal. Is this something I can do in the VMs? I will be happy to port it if so.

The limitations on bare-metal work are something I plan to fix, but I don't think it's necessary for the first candidate; I have not removed the sanity tests because of this.

jojimt:
Yes, this is run in the vagrant environment to cover the ACI mode -- it is not an e2e and does not require any physical infra.

erikh:
@jojimt ok, I will convert them later today then. PTAL at the rest if you have time, that'll save us trouble later.

shaleman:
Overall structure looks good.
Some high level comments.
 I'll add more specific comments on the code.

1. Looks like tests are run in vxlan mode only. We currently run most tests on both Vlan and Vxlan modes because code paths for the two are very different. Control plane, data plane and OVS features we use in two modes are very different. It'll be good to do the same here.
2. Random trigger test seem to be empty. It'll be good to add netplugin/netmaster restart trigger tests. Its perfectly fine if this comes in a subsequent PR.
3. As previously discussed, ACI, switchover and other recently added tests are missing. It'll be good to add those.
4. If you are planning to add these tests in a subsequent PRs, it'll be good to keep both old sanity and new system tests in jenkins runs so that we don't loose coverage during the period.
5. Since golang tests don't stop or error, it might become hard to debug failures. It might be good idea to add a stop-on-error option to tests.
6. Current tests look for errors in netplugin/netmaster logs after each iteration and fail if there are any errors. It'll be good to keep this practice.
7. It'll be good to add a test suite cleanup so that everything is cleaned up after the test suite is complete..

erikh:
* the test suite cleans up itself on boot, there's no point in cleaning it up after it finishes.
* I already implemented the code to fail on netplugin log errors -- then quickly triggered it over and over again on no actual errors.
* I'm not sure what you mean by the random trigger test seems empty, can you elaborate?
* I'll do the additional tests that are in master *now* in this test. I will be expecting you guys to port the rest. I'm not sisyphus. :)

SOE options are going to be something you guys need to investigate, I looked at a few potential options in volplugin and found no good ones. Regardless, I didn't find this to be an issue while working with these tests, since you can run them independently by passing the `-check.f` flag.

I'll fix anything not mentioned later tonight.

shaleman:
> I'm not sure what you mean by the random trigger test seems empty, can you elaborate?

TestTriggers() test isn't applying any triggers as far as i can tell.. 
python tests trigger (i) netplugin restart (2) netmaster restart OR (3) container restart randomly and test forwarding works after each trigger..

erikh:
Ah ok. I will look into what I overlooked then. Thanks for reviewing it all!

erikh:
so, everything but the vlan tests are here now. PTAL, I will have the vlan changes in later tonight, btu they will be very small.

erikh:
btw, the changes I made are in the last commit, so if you've already read the original changes you can just look there. I will squash those on top of the originals before merge.

erikh:
`systemtestSuite.TestNetpluginDisconnect` is failing randomly and I'm not sure why yet, so please do not merge this yet even if it passes.

erikh:
https://github.com/erikh/netplugin/commit/6c5a2f744076e2331b83ac5bf032d65f7bc1a48a has the latest updates, and completes the patch. PTAL.

shaleman:
@erikh It largely looks good to me other than the minor comments..
See my comments on error checking and trigger tests..

erikh:
@shaleman ok, I'll get this sorted out this evening. Thanks for the review!

erikh:
I think I've resolved everything. PTAL and merge if ready.

shaleman:
Thanks for taking care of the comments @erikh 
Pardon my nagging here.. I'm perfectly fine merging this PR and fixing these minor comments in subsequent PR..

1. I think this comment was missed in TestTriggerNetmasterSwitchover()
 https://github.com/erikh/netplugin/commit/e685740b7a8bc23926af7caf699220981f2da273#commitcomment-16163186
 ```
 We should be failing the test if new node is not elected as leader within a timeout period. 
 Currently, this might loop forever instead of failing the test..
 ```
 
2.  also, it'll be good to rotate netmaster logs in TestTriggerNetmasterSwitchover()

3.  finally, can we check for netplugin/netmaster log errors in individual tests than at the end of the test?
If we call node.checkForNetpluginErrors() during test teardown, Its very hard to say which test actually caused the error

erikh:
> 3.  finally, can we check for netplugin/netmaster log errors in 
> individual tests than at the end of the test?
> If we call node.checkForNetpluginErrors() during test teardown, Its 
> very hard to say which test actually caused the error

Pretty sure all the tests will fail if the teardown does for one of 
them, making it pretty obvious.

-Erik

shaleman:
never mind. Didn't realize TearDownTest() gets called after each test.. thats neat..

erikh:
This should be ready after CI passes, the tests pass here.

This incorporates the fix for the netmaster leader election test.

shaleman:
LGTM

erikh:
these patches are ready; please merge when appropriate.

PR #286, opened 2016-02-16
Add default gateway into pod

jojimt:


shaleman:
LGTM
I believe you can do the same using nets and netlink libraries..
See libnetwork implementation here: https://github.com/docker/libnetwork/blob/master/osl/route_linux.go#L80

There might be value in building a good linux namespace manipulation library since we'll need to do similar things for both kubernetes and mesos.. We might also need to manipulate more attributes like static routes and DNS options..

jojimt:
Agreed on the namespace lib. Will do that via a separate PR.

erikh:
you mean like https://github.com/docker/libcontainer? ;)

jojimt:
Perhaps. Ideally, we should be using netlink lib for all of these. However, I found some issues with netlink lib earlier. Rather than rat hole on that, I chose to go with nsenter for now (incidentally, nsenter is what kubelet also uses).

erikh:
I might be able to help if you've got time for it, but @unclejack was just recently using it in the mesos work I believe, so he might be able to help too.

unclejack:
@jojimt @erikh Is right, I've used that and I can take care of making this change.

jainvipin:
@jojimt - if there is only one interface in the container, this would be okay; do we need to do this only if the network this interface belongs to is the default network in the tenant? I mean we have that built in and we make the default network available to this code, why not leverage that variable? This would work for multiple-interfaces as well (as docker starts adding the support for it), besides we might be able to sue this in k8s regardless.

jojimt:
@jainvipin as discussed, let's address that when we add support for multiple interfaces.

PR #294, opened 2016-02-26
use netmaster service name as default for netctl

mapuri:
this allows netctl commands to run from any where in cluster setup by contiv ansible.

Also removed one unused variable

/cc @shaleman

### DIFF
--- a/netctl/netctl.go
+++ b/netctl/netctl.go
@@ -16,7 +16,7 @@ import (
 )
 
 // DefaultMaster is the master to use when none is provided.
-const DefaultMaster = "http://localhost:9999"
+const DefaultMaster = "http://netmaster:9999"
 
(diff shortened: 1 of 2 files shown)


shaleman:
@mapuri changes look good.
Just wondering if we should check if we can resolve the hostname `netmaster` and fall back to localhost if we can't resolve it.. Its not a big concern since ansible always sets up `netmaster` host alias..

mapuri:
@shaleman 

> Just wondering if we should check if we can resolve the hostname netmaster and fall back to localhost if we can't resolve it

yes, I thought about it. But then reverting to localhost is not a good fallback always as netmaster might not be available at localhost as well (this is what we actually hit today in clustered setups). I thought `netmaster` instead is a better value for default as atleast it covers the setups brought by ansible. In other setups users will be required to override the default.

shaleman:
LGTM

PR #295, opened 2016-02-29
Makefile,scripts/python: rm sanity.py & target

unclejack:
This PR removes the old sanity tests written in Python.


shaleman:
@unclejack 
Please don't bother with this change. 
These files will be removed soon enough.
Its abundantly clear to everyone that new test cases will be added in system-tests directory..
If you have bandwidth, we can always use your help :)

PR #297, opened 2016-03-03
MPLS additions and VLAN mask

karamsivia:
Added code for MPLS and VLAN mask


abhinandanpb:
@karamsivia I guess this PR needs to be submitted to shaleman/libopenflow repo ?

PR #298, opened 2016-03-03
avoid race using a wrapper over net.Listener 

vvb:
listener.close and server happening close to each other can cause situations where the port is still in use.
This happens even after http.Close() has been executed. Added a wrapper, which takes care of number of connections in realtime.

### DIFF
--- a/netmaster/netwrapper.go
+++ b/netmaster/netwrapper.go
@@ -1,0 +1,72 @@
+/***
+Copyright 2014 Cisco Systems Inc. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


netmaster/netwrapper.go
```diff
@@ -0,0 +1,72 @@
+/***
```
shaleman: @vvb Thanks for making this change.
Would it be possible to move this file to a common package like /utils/netutils so that we can reuse it in other places?

shaleman:
LGTM other than the minor comment

shaleman:
@vvb 
#307 and #299 fixed most of the known sanity failures. Can you rebase and resubmit this PR?

PR #299, opened 2016-03-04
Add service proxy support for k8s client

jojimt:
Brought in the ofnet changes and added a kube watch client to support service proxy.


core/core.go
```diff
@@ -21,6 +21,10 @@ limitations under the License.
 // hardware/kernel/device specific programming implementation, if any.
 package core
 
+import (
+	"github.com/contiv/ofnet"
+)
```
shaleman: I'm not sure netplugin core depending on ofnet is a good idea.
Core plugin API is supposed to be driver independent.
So that we can replace OVS/openflow driver with other drivers like opflex, ipvlan etc in future.
We should create driver independent structure definitions for plugin layer to use.

core/core.go
```diff
@@ -101,6 +105,12 @@ type NetworkDriver interface {
 	DeleteMaster(node ServiceInfo) error
 	AddBgp(id string) error
 	DeleteBgp(id string) error
+	// Add a service spec to proxy
+	AddSvcSpec(svcName string, spec *ofnet.ServiceSpec) error
```
shaleman: We should define a ServiceSpec in netplugin/core so that its driver independent

drivers/fakenetepdriver.go
```diff
@@ -1,6 +1,7 @@
 package drivers
 
 import "github.com/contiv/netplugin/core"
+import "github.com/contiv/ofnet"
```
shaleman: Same comment as above. Fake driver should not depend on OVS driver

mgmtfn/k8splugin/types.go
```diff
@@ -0,0 +1,451 @@
+package k8splugin
+
+import "time"
+
+// ListMeta describes metadata that synthetic resources must have, including lists and
+// various status objects.
+type ListMeta struct {
```
shaleman: Shouldn't we be using these definitions from k8s instead of redefining it here?
Was there a problem importing k8s structure definitions?

core/core.go
jojimt: Is it ok if ofnet includes core?

mgmtfn/k8splugin/types.go
jojimt: We can't include the package as is. There are just too many dependencies. Cloning just what we need seems to be the standard practice around this, judging from other similar packages.

mgmtfn/k8splugin/kubeClient.go
```diff
@@ -172,3 +208,140 @@ func (c *APIClient) GetPodLabel(ns, name, label string) (string, error) {
...
+				for _, addr := range subset.Addresses {
+					resp.providers = append(resp.providers, addr.IP)
+				}
+			}
+
+			log.Infof("kube ep watch: %v", resp)
+			respCh <- resp
+		}
+	}()
+}
```
shaleman: Should we be adding some form of unit test for newly added code?

core/core.go
shaleman: it should be ok.
So far we have tried to keep ofnet as an independent library thats not dependent on netplugin.
It'll be nice to keep it that way..

jojimt:
@shaleman please take a look and let me know if this is ok to merge.

jojimt:
I am merging this now as there are more things stuck in the pipe line behind this. This also fixes the sanity failure that has been blocking PRs the last few weeks. If there are any other comments, I can address them separately.

PR #306, opened 2016-03-15
Support aci app-profile creation from contiv cli

jojimt:
Brought in changes from contivmodel for aci/app-profile. Added cli to allopw create/delete app-profile.


shaleman:
LGTM

PR #315, opened 2016-04-05
Changes to support network range

DivyaVavili:
Includes changes to accept network range. 

Additional fixes include:
Fixing Min and Max address set in InitSubnet
Including check to validate network params


shaleman:
LGTM

PR #317, opened 2016-04-06
Multi Tenancy support with over lapping Subnet IP

abhinandanpb:
Multi tenancy will be supported with docker 1.10 onwards
The changes contains the following:
1) Enabled overlapping subnets across tenants
2) tenant isolation 
3) Policy support
4) Supported in routed and bridged vxlan , bridge vlan modes only. 
5) Docker 1.10+ is required


shaleman:
LGTM

PR #319, opened 2016-04-06
Add aci-opflex option to netctl CLI

saramach:
Add aci-opflex fabric mode option to the netctl CLI. This will be used when opflex is used for ACI communication in the ACI fabric mode.

### DIFF
--- a/netctl/commands.go
+++ b/netctl/commands.go
@@ -269,7 +269,7 @@ var Commands = []cli.Command{
 				Flags: []cli.Flag{
 					cli.StringFlag{
 						Name:  "fabric-mode, f",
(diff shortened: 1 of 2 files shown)


jainvipin:
ping @jojimt 
@saramach I wonder if we should also add a new API to check if we are in ACI or opflex mode?

saramach:
@jainvipin , Yes. we'll need that. Thought would add that when we start integrating. Sure. Will ping @jojimt .

shaleman:
@saramach I'll merge this PR for now. Can you take care of @jainvipin's comment during opflex changes?

saramach:
Sure Sukhesh.

Thanks
-Satish.

PR #320, opened 2016-04-08
Vagrantfile: use linked clones if supported

unclejack:
This PR enables linked clones for VirtualBox with vagrant 1.8 or newer.

This change makes vagrant skip the VM import and provisions the VMs in one second, instead of importing them.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -85,6 +85,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     else
         config.vm.box = "contiv/ubuntu1504-netplugin"
     end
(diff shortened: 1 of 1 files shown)


shaleman:
@unclejack 
Will there be performance impact to using linked clones?
for example, will system-test take longer to run on linked clones?

unclejack:
@shaleman There wouldn't be a negative performance impact. This would speed things up because we're only going to store one copy of the full VM and write to disk when something changes for that particular VM. Once the source VM gets created for that particular box, all subsequent runs using that box will just use that source VM. This will improve performance.

shaleman:
LGTM

PR #321, opened 2016-04-10
arp fix godep update

abhinandanpb:


shaleman:
LGTM.
We can merge this once sanity passes..

PR #322, opened 2016-04-11
correct return error for docker-1.11 compatibility

prateekgogia:
Seeing following error when running docker container in Docker version 1.11-rc3, with netplugin as remote driver:

root@ip-10-10-128-130:~# docker run -itd --net=busybox:latest.test busybox
25b95b93da52aa2ef32630a0c4e9a5473561d3b0ce3ca520830b45d34469b10c
docker: Error response from daemon: driver failed programming external connectivity on endpoint amazing_lumiere 

Fix includes correct response error from netplugin.

PR #289, opened 2016-02-18
Wait for OVS switch connect - fix netplugin restart crash

DivyaVavili:

PR #325, opened 2016-04-14
state store and other fixes

shaleman:
- Add a single -store-url option to specify the statestore for netplugin/netmaster
- migrate to new etc client
- add full support consul in objdb
- simplify state store API and eliminate netplugin config file
- Move away from relying on etcd's client URL for local ip address
- separate out VTEP interface from control interface
- cleanup unused godep files
- reconstruct endpoint db when netmaster restarts.
- pull in fix from PR #289 & #322.
- misc fixes in system test


shaleman:
I'm going to merge this since sanity passed

PR #332, opened 2016-04-21
Makefile,Vagrantfile: make centos the default

unclejack:
This PR makes centos the default when no OS is specified. This is the first step towards improving development and testing parity.


shaleman:
LGTM
We can merge once sanity passes

PR #338, opened 2016-04-27
Fixing Proxy ARP for ACI-gw

gaurav-dalvi:
For a version : v0.1-04-11-2016.17-31-22.UTC , proxy ARP was broken for ACI-gw. We were not able to ping default gateway because of this. With this fix, issue has been resolved and we are now able to ping gateway.


shaleman:
LGTM

PR #340, opened 2016-04-29
infra nw support

pkeerthana:


Godeps/_workspace/src/github.com/contiv/contivmodel/network.json
```diff
@@ -17,6 +17,13 @@
 					"title": "Tenant Name",
 					"length": 64
 				},
+				"nwType": {
+					"type": "string",
+					"format": "^(infra|data)$",
+					"default": "\"data\"",
```
shaleman: Didn't notice this while reviewing the contivmodel change doesn't `"default": "data",` work? do you need the `\"`?

jojimt:
Looks like this needs  a fresh rebase? It is showing some other commits as part of this changeset.
Can you please provide a brief description of the functionality you're adding as well as the code changes? That will make it easier to review the changeset.

Godeps/_workspace/src/github.com/contiv/contivmodel/network.json
pkeerthana: Yes, need the \". Just "default": "data" results in nwType = data - i.e without quotes as if data is a number.

drivers/ovsdriver.go
```diff
@@ -204,6 +209,23 @@ func (d *OvsDriver) DeleteNetwork(id, encap string, pktTag, extPktTag int, gatew
 		sw = d.switchDb["vlan"]
 	}
 
+	// Delete infra nw endpoint if present
+	if nwType == "infra" {
+		hostName, _ := os.Hostname()
```
shaleman: for my understanding? Why is this required? Wouldn't we receive endpoint delete from need?

Godeps/_workspace/src/github.com/contiv/contivmodel/network.json
shaleman: Ohh ok. Looks like a bug in modelgen compiler.. I'll take a look

shaleman:
Changes largely looks good to me.
Please rebase to latest master, godep update ofnet and resubmit the PR.

drivers/ovsdriver.go
pkeerthana: Both create and delete endpoint is explicitly triggered by the mgmt plugins - they trigger endpoint create/delete on both master and local-netplugin. Similarly infra-nw create/delete triggers endpoint create/delete on both master and local-netplugin.
When non-master nodes receive infra network delete, network is already deleted on master nodes. Endpoint delete requires network-specific information (encap and nwtype), so we will not have sufficient context to clean the endpoint after network delete processing is complete on a node.

pkeerthana:
@jojimt , @shaleman : yes, will rebase and resubmit PR. @jojimt - will add description of changes as well.

pkeerthana:
Details of changes added:
a) Networks can be of "infra" or "data" type. "data" nw type is for container endpoints. "infra" nw type is for running ceph storage or other such host-node traffic such that it shares the OVS uplink. This is needed in cases where a single 10G port is available at host, and both infra & data traffic need to share the ovs uplink.
b) In case of infra nw, the interface name (or internal OVS port name) is same as network name. Veth pair creation is not done for infra nw.
c) For "infra" nw type, an endpoint is automatically created when per node when the infra nw is created. Infra nw EP creation/deletion is triggered from netplugin on corresponding nw create/delete.

drivers/ovsSwitch.go
```diff
@@ -222,11 +222,12 @@ func setLinkMtu(name string, mtu int) error {
 	return netlink.LinkSetMTU(iface, mtu)
 }
 
-// getOvsPostName returns OVS port name depending on if we use Veth pairs
-func getOvsPostName(intfName string) string {
+// getOvsPortName returns OVS port name depending on if we use Veth pairs
+// For infra nw, dont use Veth pair
+func getOvsPortName(intfName, nwType string) string {
```
jojimt: I think it would be better if we pass a useVethPair argument to this function rather than nwType. Would be nice to avoid propagating the nw type where it is not necessary.

drivers/ovsdriver.go
```diff
@@ -189,12 +190,16 @@ func (d *OvsDriver) CreateNetwork(id string) error {
 		sw = d.switchDb["vlan"]
 	}
 
-	return sw.CreateNetwork(uint16(cfgNw.PktTag), uint32(cfgNw.ExtPktTag), cfgNw.Gateway, cfgNw.Tenant)
+	err = sw.CreateNetwork(uint16(cfgNw.PktTag), uint32(cfgNw.ExtPktTag), cfgNw.Gateway, cfgNw.Tenant)
```
jojimt: I don't see the point of this change. Better to retain the original, more concise code.

drivers/ovsSwitch.go
```diff
@@ -236,14 +237,14 @@ func getOvsPostName(intfName string) string {
 }
 
 // CreatePort creates a port in ovs switch
-func (sw *OvsSwitch) CreatePort(intfName string, cfgEp *mastercfg.CfgEndpointState, pktTag, nwPktTag int) error {
+func (sw *OvsSwitch) CreatePort(intfName string, cfgEp *mastercfg.CfgEndpointState, pktTag, nwPktTag int, nwType string) error {
 	var ovsIntfType string
```
jojimt: Same comment here - better to pass useVethPair instead of nwType.

drivers/ovsSwitch.go
```diff
@@ -326,9 +327,10 @@ func (sw *OvsSwitch) CreatePort(intfName string, cfgEp *mastercfg.CfgEndpointSta
 }
 
 // UpdatePort updates an OVS port without creating it
-func (sw *OvsSwitch) UpdatePort(intfName string, cfgEp *mastercfg.CfgEndpointState, pktTag int) error {
+func (sw *OvsSwitch) UpdatePort(intfName string, cfgEp *mastercfg.CfgEndpointState, pktTag int, nwType string) error {
+
```
jojimt: Should pass the ovsPortName here instead of deriving within this function.

drivers/ovsSwitch.go
```diff
@@ -363,17 +365,14 @@ func (sw *OvsSwitch) UpdatePort(intfName string, cfgEp *mastercfg.CfgEndpointSta
 }
 
 // DeletePort removes a port from OVS
-func (sw *OvsSwitch) DeletePort(epOper *OvsOperEndpointState) error {
+func (sw *OvsSwitch) DeletePort(epOper *OvsOperEndpointState, nwType string) error {
```
jojimt: Avoid passing the nwType if possible. Instead, use a more context relevant argument.

drivers/ovsSwitch.go
pkeerthana: This API is exposed to (and invoked by) ovsdriver which is aware only about intfName and not OVS Port name. So to be consistent with CreatePort, retained this API to take intfName and passing a skipVethPair flag instead of nwType to this API. Hope this sounds fine.

drivers/ovsSwitch.go
```diff
@@ -399,7 +398,7 @@ func (sw *OvsSwitch) DeletePort(epOper *OvsOperEndpointState) error {
 	}
 
 	// Delete the Veth pairs if required
-	if useVethPair {
+	if useVethPair && !skipVethPair {
```
jojimt: @shaleman Do we need to retain this "useVethPair" constant? If not, we could just get rid of it and that will make this cleaner.

shaleman: Yes, using Veth pairs is a temporary workaround. We should go back to using OVS internal interfaces.
Using Veth pairs vs. OVS internal interfaces takes 15-20% performance hit. So, eventually we'd like to go back to OVS internal interfaces. But, there is a bug in OVS where renaming the OVS interface to "eth0" inside the container namespace messes up ovsdb. Plan is, once that bug is fixed(or we contribute a fix to OVS), we'll go back to using OVS internal interfaces.

jojimt:
LGTM.

PR #356, opened 2016-05-10
Service load balancer support

abhinandanpb:
This PR contains the following : 
1) Minor fix for Bgp
2) Fix for netmaster restart
3) Service Load balancer netctl cli
4) service object creation
5) Attaching containers behind service based on labels
7) Initial usage of go dockerclient
6) Systemtests for vrouter mode with services
7) unit tests

DIFF: https://gist.github.com/abhinandanpb/d6fd3ec30fa751dc85307db18ada1fdc


shaleman:
@abhinandanpb Sorry, it took a while to review this code.

-var dnsBridge = &bridge.Bridge{}
+
+//DNSBridge is svc plugin bridge
+var DNSBridge = &bridge.Bridge{}

I think this part of the code has changed. You might want to rebase to latest

shaleman:
+//createServiceLB is a netctl interface routine to delete

Typo. should be 'routine to create'

shaleman:
+	// Register netmaster service
+	d.registerService()
+

This should be after all initialization(like restore cache) is done

shaleman:
+//ConfigServiceLB keeps servicelb specific configs
+type ConfigServiceLB struct {
+	ServiceName string
+	Tenant      string
+	Selectors   map[string]string
+	Network     string
+	Ports       []string
+	IPAddress   string
+}

I'm not sure if these intent struct definitions are needed anymore. If you can reuse the contivmodel service struct, its fine

shaleman:
+type SvcProvUpdateResponse struct {
+	IPAddress string // provider IP
+}

just for my education, why do we need to send the IP address of the provider back?

shaleman:
+		//update provider db
+		mastercfg.ProviderDb[providerDbID] = provider
+		for serviceID, service := range mastercfg.ServiceLBDb {
+			count := 0
+			if service.Tenant == svcProvUpdReq.Tenant {
+				for key, value := range svcProvUpdReq.Labels {
+					if val := service.Selectors[key]; val == value {
+						count++
+					}
+
+					if count == len(service.Selectors) {
+						//Container corresponds to the service since it
+						//matches all service Selectors
+						mastercfg.ProviderDb[providerDbID].Services =
+							append(mastercfg.ProviderDb[providerDbID].Services, serviceID)
+							//Update ServiceDB
+						mastercfg.ServiceLBDb[serviceID].Providers[providerID] = provider
+						stateDriver, err := utils.GetStateDriver()
+						if err != nil {
+							return nil, err
+						}
+						serviceLbState := &mastercfg.CfgServiceLBState{}
+						serviceLbState.StateDriver = stateDriver
+						err = serviceLbState.Read(serviceID)
+						if err != nil {
+							return nil, err
+						}
+						serviceLbState.Providers[providerID] = provider
+						serviceLbState.Write()
+						SvcProviderUpdate(serviceID, false)
+					}
+				}
+			}
+		}

Can we create a utility function for provider add/delete handling?
We might need to call it from multiple places

shaleman:
+//Provider has providers info
+type Provider struct {
+	IPAddress   string            // provider IP
+	ContainerID string            // container id
+	Labels      map[string]string // lables
+	Tenant      string
+	Network     string
+	Services    []string
+}

Hmm, I'm starting to wonder if we need a separate provider state or can we reuse the endpoint state..

shaleman:
+func (s *systemtestSuite) TestServiceAddDeleteService(c *C) {

Can we make these tests run in both vlan and vxlan modes?

abhinandanpb:
@shaleman : thanks for reviewing the PR. Since the gist file is old some of the comments are kind of addressed already. As discussed offline i will be address the comments and resubmit

Issue #344, opened 2016-05-03
Veth pair exists errors while moving from etcd cluster store to consul


shaleman:
Saw these errors when netplugin/netmaster were moved from etcd cluster store to consul cluster store without removing the containers.

time="May  2 21:04:57.995997833" level=error msg="error creating veth pair: file exists" 
time="May  2 21:04:57.996019066" level=error msg="Error creating veth pairs. Err: file exists" 
time="May  2 21:04:57.996038940" level=error msg="Error creating port vport9. Err: file exists" 
time="May  2 21:04:57.996058163" level=error msg="Endpoint creation failed. Error: file exists" 
time="May  2 21:04:57.996103758" level=error msg="Returning HTTP error handling plugin negotiation: Could not create endpoint file exists" 
time="May  2 21:11:44.742472546" level=error msg="error creating veth pair: file exists" 
time="May  2 21:11:44.742498243" level=error msg="Error creating veth pairs. Err: file exists" 
time="May  2 21:11:44.742521834" level=error msg="Error creating port vport10. Err: file exists" 
time="May  2 21:11:44.742544319" level=error msg="Endpoint creation failed. Error: file exists" 
time="May  2 21:11:44.742574956" level=error msg="Returning HTTP error handling plugin negotiation: Could not create endpoint file exists" 
time="May  2 21:14:32.180486718" level=error msg="error creating veth pair: file exists" 
time="May  2 21:14:32.180518575" level=error msg="Error creating veth pairs. Err: file exists" 
time="May  2 21:14:32.180547160" level=error msg="Error creating port vport11. Err: file exists" 
time="May  2 21:14:32.180574881" level=error msg="Endpoint creation failed. Error: file exists" 
time="May  2 21:14:32.180613143" level=error msg="Returning HTTP error handling plugin negotiation: Could not create endpoint file exists"

shaleman:
fixed by #357

PR #357, opened 2016-05-11
svcplugin and other bugfixes

### LINKED ISSUE #345: Repeated skydns2 errors while using consul cluster store
When Netplugin is using consul cluster store and etcd is not running in the system, netplugin repeatedly prints this error:

time="May  2 21:04:38.854514705" level=error msg="skydns2: failed to register service: 501: All the given peers are not reachable (failed to propose on members [http://127.0.0.1:4001] twice [last error: Put http://127.0.0.1:4001/v2/keys/skydns/default/contiv-net/app/0e73eeeac56e: dial tcp 127.0.0.1:4001: getsockopt: connection refused]) [0]" 
time="May  2 21:04:38.854601941" level=error msg="Service refresh failed for appcontiv-net. Error: 501: All the given peers are not reachable (failed to propose on members [http://127.0.0.1:4001] twice [last error: Put http://127.0.0.1:4001/v2/keys/skydns/default/contiv-net/app/0e73eeeac56e: dial tcp 127.0.0.1:4001: getsockopt: connection refused]) [0]"

shaleman:
- Fix #345 and #344 seen at a customer. Adds tests for both scenarios.
- Vagrantfile changes to run a cluster in etcd or consul modes so that we can run consul tests without etcd in the system
- Add checks to see Veth interfaces and OVS configuration already exists
- Rearrange svcplugin code to make it work in consul cluster.
-  Move to svcplugin to new etcd client
- Remove the delay in creating default tenant
- Now that CI is stable, re-enables skydns in CI runs
- Adds a system test to verify skydns service discovery works
- Print out all netplugin errors logs at the end of the system-test so that we can debug CI failures

### DIFF
--- a/drivers/ovsSwitch.go
+++ b/drivers/ovsSwitch.go
@@ -265,6 +265,17 @@ func (sw *OvsSwitch) CreatePort(intfName string, cfgEp *mastercfg.CfgEndpointSta
 		ovsIntfType = "internal"
 	}
 
+	// If the port already exists in OVS, remove it first
+	if sw.ovsdbDriver.IsPortNamePresent(ovsPortName) {
+		log.Debugf("Removing existing interface entry %s from OVS", ovsPortName)
+
+		// Delete it from ovsdb
+		err := sw.ovsdbDriver.DeletePort(ovsPortName)
+		if err != nil {
+			log.Errorf("Error deleting port %s from OVS. Err: %v", ovsPortName, err)
+		}
+	}
+
 	// Ask OVSDB driver to add the port
 	err := sw.ovsdbDriver.CreatePort(ovsPortName, ovsIntfType, cfgEp.ID, pktTag)
 	if err != nil {
--- a/drivers/ovsdriver.go
+++ b/drivers/ovsdriver.go
@@ -88,10 +88,13 @@ func (d *OvsDriver) getIntfName() (string, error) {
 		// Pick next port number
 		d.oper.CurrPortNum++
 		intfName := fmt.Sprintf("vport%d", d.oper.CurrPortNum)
+		ovsIntfName := getOvsPortName(intfName, false)
 
 		// check if the port name is already in use
 		_, err := netlink.LinkByName(intfName)
-		if err != nil && strings.Contains(err.Error(), "not found") {
+		_, err2 := netlink.LinkByName(ovsIntfName)
+		if err != nil && strings.Contains(err.Error(), "not found") &&
+			err2 != nil && strings.Contains(err2.Error(), "not found") {
 			// save the new state
 			err = d.oper.Write()
 			if err != nil {
--- a/netplugin/netd.go
+++ b/netplugin/netd.go
@@ -242,6 +242,34 @@ func processEpState(netPlugin *plugin.NetPlugin, opts cliOpts, epID string) erro
 	return err
 }
 
+//processBgpEvent processes Bgp neighbor add/delete events
+func processBgpEvent(netPlugin *plugin.NetPlugin, opts cliOpts, hostID string, isDelete bool) error {
+	var err error
+
+	if opts.hostLabel != hostID {
+		log.Errorf("Ignoring Bgp Event on this host")
+		return err
+	}
+	netPlugin.Lock()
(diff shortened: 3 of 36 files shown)


drivers/ovsSwitch.go
```diff
@@ -265,6 +265,15 @@ func (sw *OvsSwitch) CreatePort(intfName string, cfgEp *mastercfg.CfgEndpointSta
 		ovsIntfType = "internal"
 	}
 
+	// If the port already exists in OVS, remove it first
+	if sw.ovsdbDriver.IsPortNamePresent(ovsPortName) {
+		// Delete it from ovsdb
+		err := sw.ovsdbDriver.DeletePort(ovsPortName)
+		if err != nil {
+			log.Errorf("Error deleting port %s from OVS. Err: %v", ovsPortName, err)
+		}
```
jojimt: I think a Infof or Debugf would be useful here.

drivers/ovsdriver.go
```diff
@@ -88,10 +88,13 @@ func (d *OvsDriver) getIntfName() (string, error) {
 		// Pick next port number
 		d.oper.CurrPortNum++
 		intfName := fmt.Sprintf("vport%d", d.oper.CurrPortNum)
+		intfName2 := fmt.Sprintf("vvport%d", d.oper.CurrPortNum)
```
jojimt: It seems like we should be returning two interface names from this function?

netplugin/netd.go
```diff
@@ -242,6 +242,33 @@ func processEpState(netPlugin *plugin.NetPlugin, opts cliOpts, epID string) erro
 	return err
 }
 
+//processBgpEvent processes Bgp neighbor add/delete events
+func processBgpEvent(netPlugin *plugin.NetPlugin, opts cliOpts, hostID string,
+	isDelete bool) (err error) {
+
+	if opts.hostLabel != hostID {
+		log.Errorf("Ignoring Bgp Event on this host")
+		return
```
jojimt: We should have either a return value here or make it a void function.

netplugin/netd.go
```diff
@@ -242,6 +242,33 @@ func processEpState(netPlugin *plugin.NetPlugin, opts cliOpts, epID string) erro
...
+	} else {
+		err = netPlugin.AddBgp(hostID)
+		operStr = "create"
+	}
+	if err != nil {
+		log.Errorf("Bgp operation %s failed. Error: %s", operStr, err)
+	} else {
+		log.Infof("Bgp operation %s succeeded", operStr)
+	}
+	return
```
jojimt: Here too. Should have resulted in a compile error...

erikh: https://tour.golang.org/basics/7

jojimt: Interesting. One of the golang "features" I would rather stay away from...

erikh: I agree.

drivers/ovsSwitch.go
shaleman: ok. i'll add a Debugf

drivers/ovsdriver.go
shaleman: Since the name of veth pair is passed between different layers of the driver, we pass only one and derive the other one as needed..

netplugin/netd.go
shaleman: Btw, this i didn't add this code. I just moved this code from bottom of the file to middle where all the other event handlers are there. I'll be happy to make the return values more explicit.

jojimt: Your call. I just thought it was odd.

drivers/ovsdriver.go
shaleman: Actually, let me use getOvsPortName() wrapper to make this clearer..

netplugin/netd.go
shaleman: I agree with the comment.. I'm not a fan of named return values either..

abhinandanpb: I will fix it in my next commit.

shaleman: Already fixed.. Missed your opportunity @abhinandanpb  :)

PR #363, opened 2016-05-12
Fixing Issue#360

gaurav-dalvi:
This is to fix Issue#360.
Preventing Add function to add Link and Linkset into etcd before operation is completely successful.
Ran System Tests and unit test they are fine.

### DIFF
--- a/netmaster/objApi/apiController.go
+++ b/netmaster/objApi/apiController.go
@@ -364,18 +364,8 @@ func (ac *APIController) EndpointGroupCreate(endpointGroup *contivModel.Endpoint
 		return core.Errorf("Tenant not found")
 	}
 
-	// Setup links
-	modeldb.AddLink(&endpointGroup.Links.Tenant, tenant)
-	modeldb.AddLinkSet(&tenant.LinkSets.EndpointGroups, endpointGroup)
-
-	// Save the tenant too since we added the links
-	err := tenant.Write()
-	if err != nil {
-		return err
-	}
-
 	// create the endpoint group state
-	err = master.CreateEndpointGroup(endpointGroup.TenantName, endpointGroup.NetworkName,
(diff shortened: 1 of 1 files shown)


netmaster/objApi/apiController.go
```diff
@@ -382,6 +378,10 @@ func (ac *APIController) EndpointGroupCreate(endpointGroup *contivModel.Endpoint
 		return err
 	}
 
+	// Setup links
+	modeldb.AddLink(&endpointGroup.Links.Tenant, tenant)
+	modeldb.AddLinkSet(&tenant.LinkSets.EndpointGroups, endpointGroup)
+
```
shaleman: Two comments:
1. Please move tenant.Write() after "modeldb.AddLinkSet(&tenant.LinkSets.EndpointGroups, endpointGroup)" since it modifies the tenant's linkset.
2. This change still leaves stale link sets if master.PolicyAttach() fails. Can we handle that too

netmaster/objApi/apiController.go
```diff
@@ -584,6 +580,10 @@ func (ac *APIController) NetworkCreate(network *contivModel.Network) error {
 		return err
 	}
 
+	// Setup links
+	modeldb.AddLink(&network.Links.Tenant, tenant)
+	modeldb.AddLinkSet(&tenant.LinkSets.Networks, network)
+
```
shaleman: Same comment as above, please move the tenant.Write() after this

shaleman:
Can you please add unit test cases for the two scenarios?

netmaster/objApi/apiController.go
gaurav-dalvi: Done.

netmaster/objApi/apiController.go
gaurav-dalvi: Added in newer version of PR

PR #364, opened 2016-05-12
L3 sanity stability fixes

abhinandanpb:
Removing sleeps between for  bgp route distribution and adding explicit checks


systemtests/bgp_test.go
```diff
@@ -928,3 +928,20 @@ func (s *systemtestSuite) CheckBgpNoConnectionForaNode(c *C, node vagrantssh.Tes
 		}
 	}
 }
+
+func (s *systemtestSuite) CheckBgpRouteDistribution(c *C, node vagrantssh.TestbedNode, containers []*container) error {
+	count := 0
+	for {
```
shaleman: Can we break the infinite loop here with a really high timeout?
Else, this can loop forever when there is a route distribution bug

shaleman:
LGTM other than minor comment

systemtests/bgp_test.go
abhinandanpb: cool let me fix that

PR #367, opened 2016-05-16
Support for one Touch provisioning of Contiv as a Container

rnataraja:
I have to catch a flight in a few hours. CLIs seem to work after executing 

./netContain/BuildContainer.sh

So raising a pull request. 

One needs to modify contivInit.sh with right values and run 

./netContain/BuildContainer.sh 

Once that is done on each node, things should be up and running w.r.to cluster.


Dockerfile
```diff
@@ -20,11 +20,17 @@
 # docker run --net=host <image> -host-label=<label>
 ##
 
-FROM golang:1.4
+FROM golang:1.5.1
 MAINTAINER Madhav Puri <mapuri@cisco.com> (@mapuri)
 
+
+ENV http_proxy "http://proxy.esl.cisco.com:8080"
```
shaleman: We probably shouldn't hardcode Cisco internal http proxies in dockerfile

Makefile
```diff
@@ -43,7 +43,7 @@ run-build: deps checks clean
 	cd ${GOPATH}/src/github.com/contiv/netplugin && version/generate_version ${USE_RELEASE} && \
 	cd $(GOPATH)/src/github.com/contiv/netplugin && \
 	GOGC=1500 godep go install -v $(TO_BUILD) && \
-	sudo cp scripts/contrib/completion/bash/netctl /etc/bash_completion.d/netctl
+	$(SUDO)cp scripts/contrib/completion/bash/netctl /etc/bash_completion.d/netctl
```
shaleman: This will fail in non-docker builds since $SUDO will be undefined

Makefile
```diff
@@ -58,10 +58,15 @@ clean: deps
 update:
 	vagrant box update
 
+
 # setting CONTIV_NODES=<number> while calling 'make demo' can be used to bring
 # up a cluster of <number> nodes. By default <number> = 1
+ifeq ($(VAGRANT_SUPPORT), 1)
```
shaleman: Jenkins builds are failing because VAGRANT_SUPPORT is undefined. Please see if there is better way to do this

netContain/BuildContainer.sh
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/bash
```
shaleman: Can we please move netContain under scripts directory?

Dockerfile
shaleman: I'm not sure if we need to change this file. 
Can you elaborate what was the intent of these changes?

netContain/contivInit.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+#Initialize contiv Net Plugin and Net Master as required
+# Values below need to be tailored as per needs
+
+source contivRc
+
+if [ $IS_NETMASTER == 1 ]
+then
+   /contiv/bin/netmaster 2> /var/log/contiv/netmaster.errlog 1> /var/log/contiv/netmaster.log &
```
shaleman: Can we add a cluster-store option to netmaster so that we can point to either etcd or consul?

netContain/contivInit.sh
```diff
@@ -0,0 +1,22 @@
...
+echo "$NETMASTER_IP  netmaster" > /etc/hosts
+echo "0.0.0.0 localhost" >> /etc/hosts
+export no_proxy="0.0.0.0, $NETMASTER_IP" 
+
+if [ not $CONTIV_MODE == "routing" ] 
+then
+   CONTIV_MODE="bridged"
+fi
+
+/contiv/bin/netplugin -vtep-ip $VTEP_IP -vlan-if $VLAN_IF -fwd-mode $CONTIV_FWD_MODE 2> /var/log/contiv/netplugin.errlog 1> /var/log/contiv/netplugin.log &
```
shaleman: Can we add -cluster-store option to netplugin too?

netContain/contivNet.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+#Initialize complete contiv container. Start OVS and Net Plugin
+
+if [ $1 == "reinit" ]; then
+    ovs-vsctl del-br contivVlanBridge
+    ovs-vsctl del-br contivVxlanBridge
+fi
+
+/contiv/scripts/ovsInit.sh
+/contiv/scripts/contivInit.sh
```
shaleman: Should we look into using supervisors(http://supervisord.org) instead of just running these processes in the background?
Problem is, if netplugin or netmaster die because of a bug, there is no way to restart it

netContain/BuildContainer.sh
```diff
@@ -0,0 +1,61 @@
...
+    echo "Remove contivbase (docker rmi contivbase) and retry"
+    exit
+fi
+
+if [ $1 == "reinit" ]; then
+   etcdctl rm -recursive /contiv.io
+   etcdctl rm -recursive /docker/network
+fi
+
+sudo modprobe openvswitch
```
shaleman: Is there a way to do this from the container itself?

shaleman:
@rnataraja Thanks for the changes.
overall changes look OK to me. 
I raised some open questions so that we can discuss it. We don't need to hold this commit for all of them to be addressed.
If you can fix some of the build issues, we can move forward with this PR

Dockerfile
rnataraja: I had to make Proxy additions for the following reason. I will remove cisco proxy values and have some other values.

make start
make[1]: Entering directory '/go/src/github.com/contiv/netplugin'
make[1]: Nothing to be done for 'start'.
make[1]: Leaving directory '/go/src/github.com/contiv/netplugin'
make ssh-build
make[1]: Entering directory '/go/src/github.com/contiv/netplugin'
cd /go/src/github.com/contiv/netplugin && make run-build
make[2]: Entering directory '/go/src/github.com/contiv/netplugin'
./scripts/deps
# cd .; git clone https://github.com/tools/godep /go/src/github.com/tools/godep
Cloning into '/go/src/github.com/tools/godep'...
fatal: unable to access 'https://github.com/tools/godep/': Failed to connect to github.com port 443: Connection timed out
package github.com/tools/godep: exit status 128

Makefile
rnataraja: This is fixed in the latest commit.

netContain/BuildContainer.sh
rnataraja: I have moved it under scripts in the latest commit.

Makefile
rnataraja: fixed in the latest commit.

netContain/contivNet.sh
rnataraja: Yes, I agree. I was thinking something similar as well.

netContain/BuildContainer.sh
rnataraja: This is regarding modprobe right? 
I found it tricky as just mounting the modules lib does not seem to help, it looks for more things and was throwing errors. 
May help make ubuntu version much more leaner if we get rid of some tools like this. We can modprobe from the build script it self.

netContain/contivInit.sh
rnataraja: Can we do this in a follow up commit?

netContain/contivInit.sh
rnataraja: Can we do this in a follow up commit?

PR #371, opened 2016-05-17
vlan network support in k8s vagrant

pkeerthana:
Changed vagrant file to add support for vlan network with k8s VM: 
Ethernet interface used as contivVlanBridge uplink should have ip 0, and have promisc/virtio settings enabled.

### DIFF
--- a/vagrant/k8s/Vagrantfile
+++ b/vagrant/k8s/Vagrantfile
@@ -11,6 +11,18 @@ def customize_vm(v)
   v.customize ["modifyvm", :id, "--cpus", cpus]
 end
 
+# method to set up network attributes
+def customize_vm_network(v)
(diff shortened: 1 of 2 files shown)


jojimt:
@pkeerthana why do we need vlan/k8s in vagrant?

pkeerthana:
@jojimt - sky poc needs vlan network instead of vxlan network. contiv vlan network doesnt work even in bridge mode without specific port settings in vagrant.

jojimt:
I understand vlan is not supported in vagrant now. I don't think vagrant is the right set up for this. Let's chat offline.

pkeerthana:
We will have only vxlan network in vagrant setup for demo, since vlan network cannot be properly validated in vagrant setup. So closing this.

PR #386, opened 2016-05-25
Service load balancer support

abhinandanpb:
This PR contains the following : 
1) Minor fix for Bgp
2) Fix for netmaster restart
3) Service Load balancer netctl cli
4) service object creation
5) Attaching containers behind service based on labels
7) Initial usage of go dockerclient
6) Systemtests for vrouter mode with services
7) unit tests
8) L3 sanity fixes
9) updated godeps


abhinandanpb:
L3 sanity passed as well
http://contiv.ngrok.io/view/Netplugin/job/Netplugin_OnDemand_L3_CI/22/

netmaster/daemon.go
```diff
@@ -405,3 +415,31 @@ func (d *daemon) runMasterFsm() {
...
+		return nil, err
+	}
+
+	if id == "all" {
+		return svc.ReadAll()
+	} else if err := svc.Read(id); err == nil {
+		return []core.State{core.State(svc)}, nil
+	}
+
+	return nil, core.Errorf("Unexpected code path")
```
jojimt: Let's print and return the actual error here instead of "Unexpected code path" :)

netmaster/master/api.go
```diff
@@ -72,6 +71,22 @@ type DeleteEndpointRequest struct {
...
+	IPAddress   string            // provider IP
+	ContainerID string            // container id
+	Labels      map[string]string // lables
+	Tenant      string
+	Network     string
+	Event       string
+	Container   string
+}
+
+//SvcProvUpdateResponse is service provider update request from netplugin
```
jojimt: Please fix the comment.

netmaster/master/api.go
```diff
@@ -296,3 +311,147 @@ func DeleteEndpointHandler(w http.ResponseWriter, r *http.Request, vars map[stri
...
+		}
+
+		providerID := getProviderID(provider)
+		providerDbID := getProviderDbID(provider)
+		if providerID == "" || providerDbID == "" {
+			return nil, fmt.Errorf("Invalid ProviderID from providerInfo:{%v}", provider)
+		}
+
+		//update provider db
+		mastercfg.ProviderDb[providerDbID] = provider
```
jojimt: Would be useful to check if an entry already exists.

netmaster/daemon.go
abhinandanpb: @jojimt sure will make the change.

netmaster/master/api.go
abhinandanpb: agreed. copy paste error :)

netmaster/master/api.go
abhinandanpb: ah providerDbID is the container id. and it we wouldnt expect consecutive start events from the same container. And even if we do - i guess it would make sense to just the updated providerinfo.

netmaster/master/api.go
```diff
@@ -296,3 +311,147 @@ func DeleteEndpointHandler(w http.ResponseWriter, r *http.Request, vars map[stri
...
+		for serviceID, service := range mastercfg.ServiceLBDb {
+			count := 0
+			if service.Tenant == svcProvUpdReq.Tenant {
+				for key, value := range svcProvUpdReq.Labels {
+					if val := service.Selectors[key]; val == value {
+						count++
+					}
+
+					if count == len(service.Selectors) {
```
jojimt: Ideally, the matching logic should be moved to a function. At the very least, we should break out of the inner loop once a match has been found.

netmaster/master/api.go
```diff
@@ -296,3 +311,147 @@ func DeleteEndpointHandler(w http.ResponseWriter, r *http.Request, vars map[stri
...
+						//Container corresponds to the service since it
+						//matches all service Selectors
+						mastercfg.ProviderDb[providerDbID].Services =
+							append(mastercfg.ProviderDb[providerDbID].Services, serviceID)
+							//Update ServiceDB
```
jojimt: Would this result in duplicate entries if there are multiple providers of the same service?

netmaster/master/api.go
```diff
@@ -296,3 +311,147 @@ func DeleteEndpointHandler(w http.ResponseWriter, r *http.Request, vars map[stri
...
+		}
+
+		for k, v := range svcProvUpdReq.Labels {
+			provider.Labels[k] = v
+			epCfg.Labels[k] = v
+		}
+		//maintain the containerId in endpointstat for recovery
+		epCfg.ContainerID = svcProvUpdReq.ContainerID
+
+		err = epCfg.Write()
```
jojimt: Do we have any protection for this write? Would we be ok if the container is removed while we are in the middle of this?

netmaster/master/api.go
```diff
@@ -296,3 +311,147 @@ func DeleteEndpointHandler(w http.ResponseWriter, r *http.Request, vars map[stri
...
+			return nil, nil
+		}
+
+		for _, serviceID := range provider.Services {
+			service := mastercfg.ServiceLBDb[serviceID]
+			providerID := getProviderID(provider)
+			if providerID == "" {
+				return nil, fmt.Errorf("Invalid ProviderID from providerInfo:{%v}", provider)
+			}
+			if service.Providers[providerID] != nil {
```
jojimt: I think we want to make sure service exists in the dict before accessing it.

netmaster/master/api.go
```diff
@@ -296,3 +311,147 @@ func DeleteEndpointHandler(w http.ResponseWriter, r *http.Request, vars map[stri
...
+						//matches all service Selectors
+						mastercfg.ProviderDb[providerDbID].Services =
+							append(mastercfg.ProviderDb[providerDbID].Services, serviceID)
+							//Update ServiceDB
+						mastercfg.ServiceLBDb[serviceID].Providers[providerID] = provider
+						stateDriver, err := utils.GetStateDriver()
+						if err != nil {
+							return nil, err
+						}
+						serviceLbState := &mastercfg.CfgServiceLBState{}
```
jojimt: Please add locking as discussed.

netmaster/master/provider.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	for _, provider := range mastercfg.ServiceLBDb[serviceID].Providers {
+		providerList = append(providerList, provider.IPAddress)
+	}
+
+	//empty the current provider list
+	svcProvider.Providers = nil
+	//update to the latest provider list
+	svcProvider.Providers = append(svcProvider.Providers, providerList...)
```
jojimt: Probably better to use direct assignment than append here.

PR #398, opened 2016-06-07
adding inspect support

jainvipin:
- add support for network/global inspect objects
- vendor in contivmodel to take get the oper state to be show in inspect cli
- add unit test cases for various oper state operations

Not done:
- adding oper state of wider set of objects (i.e. all)
- system test cases to verify inspect operation (could be done, pls point me to which is the best place to add them)


shaleman:
LGTM

PR #399, opened 2016-06-08
aci-sys-infra commit 1

gaurav-dalvi:
aci-sys-infra commit - Adding automation for ACI Swarm systemtests.

### DIFF
--- a/systemtests/aci_test.go
+++ b/systemtests/aci_test.go
@@ -59,3 +59,48 @@ func (s *systemtestSuite) TestACIMode(c *C) {
 	c.Assert(s.cli.EndpointGroupDelete("default", "epgB"), IsNil)
 	c.Assert(s.cli.NetworkDelete("default", "aciNet"), IsNil)
 }
+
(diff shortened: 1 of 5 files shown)


systemtests/aci_test.go
```diff
@@ -59,3 +59,48 @@ func (s *systemtestSuite) TestACIMode(c *C) {
 	c.Assert(s.cli.EndpointGroupDelete("default", "epgB"), IsNil)
 	c.Assert(s.cli.NetworkDelete("default", "aciNet"), IsNil)
 }
+
```
jojimt: Would be good to move this to a separate file which can hold all tests that require ACI physical fabric.

jojimt:
LGTM other than the minor comment.

systemtests/aci_test.go
gaurav-dalvi: All ACI specific test cases would come under this file as of now. TestACIMode was already present and I think we should add ACI specific test cases here in the same file.

PR #416, opened 2016-06-17
fixing issues in ACIMode system-test case

gaurav-dalvi:
Instead of default, adding custom tenant.

### DIFF
--- a/systemtests/aci_test.go
+++ b/systemtests/aci_test.go
@@ -15,11 +15,14 @@ func (s *systemtestSuite) TestACIMode(c *C) {
 	c.Assert(s.cli.GlobalPost(&client.Global{
 		Name:             "global",
 		NetworkInfraType: "aci",
-		Vlans:            "1-4094",
+		Vlans:            "1140-1150",
 		Vxlans:           "1-10000",
 	}), IsNil)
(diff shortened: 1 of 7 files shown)


systemtests/aci_test.go
```diff
@@ -15,39 +15,47 @@ func (s *systemtestSuite) TestACIMode(c *C) {
 	c.Assert(s.cli.GlobalPost(&client.Global{
 		Name:             "global",
 		NetworkInfraType: "aci",
-		Vlans:            "1-4094",
+		Vlans:            "1140-1150",
```
shaleman: Why is vlan range significant for this test?
Tests become less portable if we hard code plan ranges based on a specific setup..

shaleman:
LGTM other than minor comments

systemtests/aci_test.go
jojimt: This is a hw independent test that gets run as part of sanity. The vlan range should not matter for this test. However, the vlan range will matter for the other tests that use real ACI fabric. @gaurav-dalvi can we create a shared config file that will contain all the external fabric parameters: vlan range, tenant, physdom, external contracts etc. This can be a json/yml dict, keyed by the IP address of the setup. That way, each setup will gets its on vlan.

PR #424, opened 2016-06-23
Added functionality to take custom parameters for ACI testing.

apoorva11029:
The ACI tests use parameters from a JSON file- systemtests/cfg.json.
Added another test- TestACIGW_Restart to validate the functionality after restarting the ACI gateway.


systemtests/aci_test.go
```diff
@@ -386,11 +393,188 @@ func (s *systemtestSuite) TestACIProfile(c *C) {
...
+	cB2, err := s.nodes[0].runContainer(containerSpec{networkName: netstrB})
+	c.Assert(err, IsNil)
+
+	// Verify cA1 can ping cA2
+	c.Assert(cA1.checkPingWithCount(cA2.eth0.ip, 5), IsNil)
+	// Verify cB1 can ping cB2
+	c.Assert(cB1.checkPingWithCount(cB2.eth0.ip, 5), IsNil)
+
+	//s.nodes[0].restartContainer("contiv-aci-gw")
+	//time.Sleep(100 * time.Millisecond)
```
shaleman: please cleanup unused code

systemtests/aci_test.go
```diff
@@ -386,11 +393,188 @@ func (s *systemtestSuite) TestACIProfile(c *C) {
...
+		PolicyName:        "policyAB",
+		TenantName:        s.acinfo.Tenant,
+		FromEndpointGroup: "epgA",
+		Direction:         "in",
+		Protocol:          "tcp",
+		Port:              8001,
+		Action:            "allow",
+	}), IsNil)
+	//cA1.checkPing("20.1.1.254")
+	//cB1.checkPing("20.1.1.254")
```
shaleman: cleanup unused code

systemtests/aci_test.go
```diff
@@ -386,11 +393,188 @@ func (s *systemtestSuite) TestACIProfile(c *C) {
...
+		EndpointGroups: []string{"epgA", "epgB"},
+		AppProfileName: "profile1",
+	}), IsNil)
+	netstrA := "epgA/" + s.acinfo.Tenant
+	netstrB := "epgB/" + s.acinfo.Tenant
+	time.Sleep(5 * time.Second)
+	cA1, err := s.nodes[0].runContainer(containerSpec{networkName: netstrA})
+	c.Assert(err, IsNil)
+	// Verify cA1 can ping default gateway
+	c.Assert(cA1.checkPingWithCount("20.1.1.254", 5), IsNil)
```
shaleman: Shouldn't the gateway ip address be taken from JSON config?

systemtests/aci_test.go
```diff
@@ -386,11 +393,188 @@ func (s *systemtestSuite) TestACIProfile(c *C) {
...
+		c.Assert(s.cli.EndpointGroupDelete(s.acinfo.Tenant, "epgA"), IsNil)
+		c.Assert(s.cli.EndpointGroupDelete(s.acinfo.Tenant, "epgB"), IsNil)
+		c.Assert(s.cli.RuleDelete(s.acinfo.Tenant, "policyAB", "2"), IsNil)
+		c.Assert(s.cli.RuleDelete(s.acinfo.Tenant, "policyAB", "3"), IsNil)
+		c.Assert(s.cli.PolicyDelete(s.acinfo.Tenant, "policyAB"), IsNil)
+		c.Assert(s.cli.NetworkDelete(s.acinfo.Tenant, s.acinfo.Network), IsNil)
+	}
+}
+
+func (s *systemtestSuite) TestACIGW_Restart(c *C) {
```
shaleman: Couple of high level comments on this test:

1. Its good to write tests such that it can run on any number of nodes. Not just a single node. This way we can reuse the tests on different kinds of setups. Please see non-ACI tests for an example
2. We should take number of iteration as an argument and run the tests multiple times. please see other tests for an example.
3. We should also take number of containers as an argument too.
4. If there are different things being tested, like connectivity and policy. Its generally best to break that up into two separate tests.
5. Its best to separate the initialization (i.e, part where you setup network/epg/policy/rule ) and test run (i.e. where you are running the containers and checking ping)

systemtests/init_test.go
```diff
@@ -34,6 +34,21 @@ type systemtestSuite struct {
 	// user       string
 	// password   string
 	// nodes      []string
+	acinfo ACInfo
+}
+type ACInfo struct {
+	IP                string `json:"IP"`
+	HostIPs           string `json:"HOSTIPS"`
+	HostUsernames     string `json:"HOSTUSERNAMES"`
```
shaleman: Convention for JSON field names is to use lower case letters(not caps).
Can you please change it to lower case?

systemtests/init_test.go
```diff
@@ -34,6 +34,21 @@ type systemtestSuite struct {
...
+	HostDataInterface string `json:"HOSTDATA"`
+	Vlan              string `json:"VLAN"`
```
shaleman: This struct seems to be mixing global info and host specific info.
Would it be better to split it into two separate structs?

systemtests/util_test.go
```diff
@@ -794,3 +795,43 @@ func (s *systemtestSuite) startListenersOnProviders(containers []*container, por
...
+	json.Unmarshal(raw, &c)
+	return c
+}
+
+//Function to get the master node for ACI mode
+func getMaster() ACInfo {
+	infos := getInfo()
+	var mast ACInfo
+	for _, p := range infos {
+		if p.Master == 1 {
```
shaleman: please use boolean field if its just true/false

systemtests/node_test.go
```diff
@@ -160,6 +160,10 @@ func (n *node) runCommand(cmd string) (string, error) {
 
 	return str, err
 }
+func (n *node) restartContainers(str string) error {
+
+	return n.tbnode.RunCommand("docker restart `docker ps | grep " + str + " | awk '{print $1}' `")
```
shaleman: you should be able to do just `docker restart <container-name>`

apoorva11029: This is to restart containers with ambiguous names, using their IDs.

PR #425, opened 2016-06-23
issue-423- commit-1

gaurav-dalvi:
This is to fix netplugin issue 423.
stale entries of errored app-profile in etct store.

### DIFF
--- a/netmaster/objApi/apiController.go
+++ b/netmaster/objApi/apiController.go
@@ -246,15 +246,15 @@ func (ac *APIController) AppProfileCreate(prof *contivModel.AppProfile) error {
 		}
 	}
 
+	// Setup links
+	modeldb.AddLink(&prof.Links.Tenant, tenant)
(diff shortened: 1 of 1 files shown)


netmaster/objApi/apiController.go
```diff
@@ -256,6 +252,10 @@ func (ac *APIController) AppProfileCreate(prof *contivModel.AppProfile) error {
 		return err
 	}
 
+	// Setup links
+	modeldb.AddLink(&prof.Links.Tenant, tenant)
+	modeldb.AddLinkSet(&tenant.LinkSets.AppProfiles, prof)
```
shaleman: This needs to be before tenant.Write()

shaleman:
LGTM other than small comment

netmaster/objApi/apiController.go
gaurav-dalvi: Thanks for catching that. Fixed and resubmitted it !

Issue #435, opened 2016-06-27
After interface shut/no shut on leaf , traffic between containers is stopped


santosh76:
Created an app-profile for 2 epg belonging to a single bd. bd subnet 20.1.1.254/24. 
Created contracts between 2 epgs allowing icmp traffic. 
Spawned 50 containers in each epg and verified one-2-all icmp traffic . All the container mac is learnt on the epg.
Shut/No shut the interface connected to leaf. All the learnt entries are flushed. 
Now the traffic does not flow between the containers unless I ping the bd gateway from the container which should not be the case

gaurav-dalvi:
@santosh76 : Looks like this is fixed. Could you please verify and close this issue ?

PR #438, opened 2016-06-29
Update aci-gw interface to use the new epg model and pick up ofnet 

jojimt:
1) Addresses contiv/netplugin#435
2) Update aci-gw interface to use the new model and use contiv network name as aci BD name
3) Propagate app-profile errors back.


shaleman:
LGTM

PR #439, opened 2016-06-29
enhance ip range config

shaleman:
- change ip add range to <from-addr> - <to-addr> format
- add a available ip address field to network inspect.
- add tests around ip addr range


jojimt:
LGTM

Issue #411, opened 2016-06-16
Remove additional network from the list in ACI mode


gaurav-dalvi:
When we run netplugin in ACI mode and then create Tenant, network and EOG we are seeing these networks under docker.

docker network ls
NETWORK ID          NAME                DRIVER
38bad00aa2e8        localhost/host      host                
181d493b220d        localhost/bridge    bridge              
a23bf605e9b8        localhost/none      null                
849e8fee1709        localhost/host      host                
84f15b0755d4        localhost/bridge    bridge              
5324bc67f00e        n1/t1               netplugin     ---------- This should not be shown.       
80eb975fc33a        g1/t1               netplugin           
a74196672486        localhost/none      null

gaurav-dalvi:
Even we see it in Non ACI mode:

docker network ls
NETWORK ID          NAME                                     DRIVER
3b5ad16829c5        contiv-demo-swarm-l2-6/host              host                
b43c4cce400f        contiv-demo-swarm-l2-4/host              host                
1a9f0aaa2853        contiv-demo-swarm-l2-4/docker_gwbridge   bridge              
f56b6a91bd0d        g1/t1                                    netplugin           
d9f07187fdc6        contiv-demo-swarm-l2-6/none              null                
73509f411200        g2/t1                                    netplugin           
8f7feb8719bc        g4/t1                                    netplugin           
c19f67bbb2c2        contiv-demo-swarm-l2-5/bridge            bridge              
466b451a7c88        contiv-demo-swarm-l2-5/none              null                
14b62b0d7d1f        contiv-demo-swarm-l2-5/host              host                
68084c7596af        contiv-demo-swarm-l2-6/bridge            bridge              
7d5cee5b1e6a        n1/t1                                    netplugin           
b308caaec075        n2/t1                                    netplugin           
f065851f4661        contiv-demo-swarm-l2-5/docker_gwbridge   bridge              
43ff9905bca8        contiv-demo-swarm-l2-4/bridge            bridge              
53993541ba7c        contiv-demo-swarm-l2-4/none              null                
b99e7ddf033e        g3/t1                                    netplugin

abhinandanpb:
@gaurav-dalvi this should be epg networks that we create in docker ?

gaurav-dalvi:
@abhinandanpb yes. This issue is only in case ACI. We dont docker network ls to show networks of these types : <networkName>/<Tenant Name>. This is fine in Non ACI Mode though

[...]

PR #446, opened 2016-07-09
disabling network creation in ACI mode

gaurav-dalvi:
This is regarding : https://github.com/contiv/netplugin/issues/411

### DIFF
--- a/netmaster/master/network.go
+++ b/netmaster/master/network.go
@@ -177,6 +177,12 @@ func CreateNetwork(network intent.ConfigNetwork, stateDriver core.StateDriver, t
 		return nil
 	}
 
+	aci, _ := IsAciConfigured()
+	if aci {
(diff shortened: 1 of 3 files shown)


netmaster/master/network.go
```diff
@@ -177,6 +177,12 @@ func CreateNetwork(network intent.ConfigNetwork, stateDriver core.StateDriver, t
 		return nil
 	}
 
+	aci, _ := IsAciConfigured()
```
jojimt: The code change looks good. However, we need to add a system test case to cover this. You could this to TestACIMode and verify "docker network ls" shows only epgs.

jojimt:
LGTM - once sanity passes, we can merge.

gaurav-dalvi:
@jojimt / @shaleman : Could you please merge this one ?

PR #447, opened 2016-07-10
Host acc

jojimt:
Added a host bridge to enable access from the containers to the host network.


shaleman:
LGTM

netmaster/utils.go
```diff
@@ -57,6 +57,10 @@ func slaveProxyHandler(w http.ResponseWriter, r *http.Request) {
 
 	// get current holder of master lock
 	masterNode := leaderLock.GetHolder()
+	if masterNode == "" {
+		http.Error(w, "Self proxying error", http.StatusInternalServerError)
```
shaleman: minor: Can we change this error message to "Leader not found"?

PR #451, opened 2016-07-11
Resolving Issue#450

apoorva11029:


shaleman:
@apoorva11029 
Thanks for making this change. 
These tests are not used anymore and we need to deprecate them.
Can you remove the entire scripts/python/testcases/ directory along with following files in scripts/python/ directory?

runTestCase.py
sanity.py
setupProxy.py

shaleman:
LGTM

PR #454, opened 2016-07-13
Resolving issue#450

apoorva11029:
Removing deprecated files


scripts/python/startPlugin.py
```diff
@@ -5,7 +5,6 @@
 import time
 import sys
 import os
-import setupProxy
```
shaleman: Can you also remove these two lines in the same file?

```
# Setup proxy
# setupProxy.setupProxy()
```

PR #463, opened 2016-07-15
Global forwarding mode knob

abhinandanpb:
The changes include 
1) capability to change the forwarding mode between routing and bridging.
2) Ovs cleanups
3) netctl changes to update global config (stateful)
4) Minor cleanups and fixes.


netctl/commands.go
```diff
@@ -360,12 +360,14 @@ var Commands = []cli.Command{
 					cli.StringFlag{
 						Name:  "vlan-range, v",
 						Usage: "Allowed Vlan id range",
-						Value: "1-4094",
```
jojimt: why is this taken out?

abhinandanpb: from now we can set individual parameters of the global settings. If we allow this then we will have defaults set. For example if we already had vlans 1-2000 as the global setting and i want to only change fwd mode and retain the vlans and vxlans - i should be able to do it without vlans being defaulted.

netplugin/plugin/netplugin.go
```diff
@@ -220,3 +196,43 @@ func (p *NetPlugin) GetEndpointStats() ([]byte, error) {
...
+
+//GetFwdMode returns the fabric forwarding mode
+func GetFwdMode(stateDriver core.StateDriver) string {
+
+	gCfg := mastercfg.GlobConfig{}
+	gCfg.StateDriver = stateDriver
+	err := gCfg.Read("")
+	if err != nil {
+		core.Errorf("Error reading forwarding mode from cluster store")
+		return ""
```
jojimt: Should we handle the error here?

netplugin/plugin/netplugin.go
```diff
@@ -220,3 +196,43 @@ func (p *NetPlugin) GetEndpointStats() ([]byte, error) {
 func (p *NetPlugin) InspectState() ([]byte, error) {
 	return p.NetworkDriver.InspectState()
 }
+
+//GlobalFwdModeUpdate update the forwarding mode
+func (p *NetPlugin) GlobalFwdModeUpdate(cfg Config) {
+	var err error
+
+	if p.NetworkDriver != nil {
```
jojimt: If we had any endpoints on the OVS at this point, would that result in an error?

abhinandanpb: we dont allow the fwd mode change if we have any networks created - so endpoints will not be there

jojimt: Agreed that we have that check. But that does not necessarily guarantee everything will be ok at the netplugin. If we somehow had an endpoint on the OVS/ofnet, what would be the behavior?

abhinandanpb: we will delete the ovsSwitch instance and corresponding ofnetagent. The endpoint will be removed from ofnet.

jojimt:
LGTM other than the etcd error handling comment.

netplugin/plugin/netplugin.go
abhinandanpb: @jojimt this will handle the case where netplugin starts first followed by netmaster. In this case global config will not be there during netplugin init. So we default  to bridge mode.

shaleman:
@abhinandanpb Looks like this needs a rebase

abhinandanpb:
@shaleman yes. I also need to update godeps from contivmodel and ofnet. . I am testing ansible changes now for l3 mode.

PR #465, opened 2016-07-19
list all option for policy and ext-contracts

reshmagiliyal:


netctl/netctl.go
```diff
@@ -33,12 +33,12 @@ func createPolicy(ctx *cli.Context) {
...
 	policy := ctx.Args()[0]
 
-	logrus.Infof("Creating policy %s:%s", tenant, policy)
-
 	errCheck(ctx, getClient(ctx).PolicyPost(&contivClient.Policy{
 		PolicyName: policy,
 		TenantName: tenant,
 	}))
+
+	fmt.Printf("Creating policy %s:%s\n", tenant, policy)
```
gaurav-dalvi: Can you change this to 
logrus.Infof("Creating policy %s:%s", tenant, policy)

reshmagiliyal: The reason i kept it fmt.Printf is that it does not say[INFO] like logrus.Infof, it just displays the message of the console. Since its better to just display the message on the console for the user, beautified it.

gaurav-dalvi: In other functions also we are using log.Infof.
Check in createTenant and deleteTenant etc.

reshmagiliyal: updating all the netctl info messages to fmt.Prinftf

gaurav-dalvi:
LGTM

netctl/netctl.go
```diff
@@ -943,7 +951,7 @@ func inspectServiceLb(ctx *cli.Context) {
 	tenant := ctx.String("tenant")
 	service := ctx.Args()[0]
 
-	logrus.Infof("Inspeting service: %s tenant: %s", service, tenant)
+	fmt.Printf("Inspeting service: %s tenant: %s\n", service, tenant)
```
jojimt: Typo - "Inspecting"

reshmagiliyal: Thank you for pointing that out. sorry i missed that. Corrected.

PR #466, opened 2016-07-19
Includes tenant/network/servicelb checks and list -a for serviceLb

reshmagiliyal:


netctl/netctl.go
```diff
@@ -823,13 +823,12 @@ func deleteServiceLB(ctx *cli.Context) {
 	errCheck(ctx, getClient(ctx).ServiceLBDelete(tenantName, serviceName))
 }
 
-//listServiceLB is a netctl interface routine to delete
+//listServiceLB is a netctl interface routine to list
 //service object
 func listServiceLB(ctx *cli.Context) {
-	argCheck(1, ctx)
+	argCheck(0, ctx)
```
abhinandanpb: I think we can remove this check.

netmaster/objApi/apiController.go
```diff
@@ -1215,6 +1215,25 @@ func (ac *APIController) ServiceLBCreate(serviceCfg *contivModel.ServiceLB) erro
...
+	network := contivModel.FindNetwork(nwObjKey)
+	if network == nil {
+		return core.Errorf("Network %s not found", serviceCfg.NetworkName)
+	}
+
+	svcKey := serviceCfg.TenantName + ":" + serviceCfg.ServiceName
+	svc := contivModel.FindServiceLB(svcKey)
+	if svc == nil {
+		return core.Errorf("Service: %s not found", serviceCfg.ServiceName)
+	}
```
abhinandanpb: ah why do we need these checks ? unless we are adding any link sets this may not be needed.

abhinandanpb:
@reshmagiliyal : thanks for taking care of this.

netmaster/objApi/apiController.go
reshmagiliyal: the tenant check and network check is required because it was allowing me to create even an invalid tenant/network. so if teh tenant/network is not present, we need to throw an error.

gaurav-dalvi:
LGTM

abhinandanpb:
LGTM

PR #475, opened 2016-07-23
Contiv Containerization Part 2

rnataraja:
Add Support for Kubernetes.


shaleman:
LGTM
Please rebase

rnataraja:
I have reverted the file that was causing compilation error. There is no need to commit that any more as I see another commit has made the same change. the newer change should not cause any build failure.

PR #484, opened 2016-08-01
Restrict EPG delete with active EPs

DivyaVavili:
PR includes:
- State change in EPG to maintain active EP count and restrict deletion when active EPs are present
- Fix bug in network deletion on active EPs (the operation was partially cleaning up tenant state)
- Fix some minor typo

### DIFF
--- a/netmaster/master/endpointGroup.go
+++ b/netmaster/master/endpointGroup.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"strconv"
 
+	"github.com/contiv/netplugin/core"
 	"github.com/contiv/netplugin/netmaster/docknet"
 	"github.com/contiv/netplugin/netmaster/gstate"
 	"github.com/contiv/netplugin/netmaster/mastercfg"
@@ -149,6 +150,10 @@ func DeleteEndpointGroup(tenantName, groupName string) error {
 		return err
 	}
 
+	if epgCfg.EpCount != 0 {
+		return core.Errorf("Error: EPG %s has active endpoints", groupName)
+	}
+
 	// Delete the endpoint group state
 	gCfg := gstate.Cfg{}
 	gCfg.StateDriver = stateDriver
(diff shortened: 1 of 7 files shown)


gaurav-dalvi:
LGTM

netmaster/master/endpointGroup.go
```diff
@@ -149,6 +150,10 @@ func DeleteEndpointGroup(tenantName, groupName string) error {
 		return err
 	}
 
+	if epgCfg.EpCount != 0 {
+		return core.Errorf("Error: EPG has active endpoints")
```
abhinandanpb: can we mention the epg name here ?  So while debugging we can use epg inspect to list the active endpoints.

netmaster/objApi/apiController.go
```diff
@@ -667,8 +670,12 @@ func (ac *APIController) EndpointGroupDelete(endpointGroup *contivModel.Endpoint
 			endpointGroup.GroupName, endpointGroup.Links.AppProfile.ObjKey)
 	}
 
-	endpointGroupCleanup(endpointGroup)
-	return nil
+	err := endpointGroupCleanup(endpointGroup)
+	if err != nil {
+		log.Errorf("EPG cleanup failed: %+v", err)
```
abhinandanpb: same as above.

systemtests/network_test.go
```diff
@@ -266,6 +266,51 @@ func (s *systemtestSuite) testNetworkAddDeleteNoGateway(c *C, encap string) {
...
+func (s *systemtestSuite) TestNetworkEPGAddDelete(c *C) {
+
+	var err error
+	numContainers := 4
+	networkName := "net1"
+	epgName := "epg1"
+
+	for i := 0; i < s.iterations; i++ {
+		// Create Endpoints
+		network := &client.Network{
```
abhinandanpb: Can we make this multiple networks , multiple epgs ?

abhinandanpb: Also should this test be run in L3 mode as well ? If so we need to add setup bgp step

netmaster/master/endpointGroup.go
DivyaVavili: Adding a EPG name here

netmaster/objApi/apiController.go
DivyaVavili: The previous return will be printed here. So, it should contain the EPG name as well

systemtests/network_test.go
DivyaVavili: I'll add support for multiple EPGs, but not sure if adding the multiple networks will help here and might just end up increasing test execution time. That part is already been tested in "TestNetworkAddDelete*" routines. Does adding/deleting EPG have a different effect in L3 mode? If so, I'll add support for that.

PR #489, opened 2016-08-03
Resolving issue 486

gaurav-dalvi:
This PR is to take care of Issue#486. By default 3 nodes should be running whenever we are running system-tests.
make demo, make start should be working as it is on 2 nodes before.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,8 +76,6 @@ else
 start: update
 	CONTIV_NODE_OS=${CONTIV_NODE_OS} vagrant up
 endif
-system-test-start:
-	CONTIV_NODES=3 vagrant up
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -76,6 +76,8 @@ else
 start: update
 	CONTIV_NODE_OS=${CONTIV_NODE_OS} vagrant up
 endif
+system-test-start:
+	CONTIV_NODES=3 vagrant up
```
gaurav-dalvi: Makes sense. The reason I wanted to do this "3 nodes" only when system-test is running and for all other Makefile targets, this should not affect.

gaurav-dalvi: Updated PR !

PR #493, opened 2016-08-04
Redirect service traffic to contiv bridge

jojimt:
When host bridge is the default gateway, we want to redirect service traffic to contiv bridge.


shaleman:
LGTM

PR #494, opened 2016-08-04
fix host interface name as the first argument to avoid failure.

rchirakk:
This fixes  error messages observed while deleting interface from host bridge.


jojimt:
LGTM

PR #498, opened 2016-08-06
fix nomad demo build.

rchirakk:
fix nomad demo build.
fix permission & godep for go 1.5.x


shaleman:
LGTM
Thanks for taking care of this..

PR #501, opened 2016-08-08
export 9998 to allow running contiv-ui

jainvipin:
allow nating port 80 to port 9998 to expose contiv-ui on the VM


shaleman:
LGTM

Issue #502, opened 2016-08-08
swarm cluster is not up when you do make demo


gaurav-dalvi:
When you set DOCKER_HOST, docker ps is giving error 

```
docker ps
Cannot connect to the Docker daemon. Is the docker daemon running on this host?

```

gaurav-dalvi:
Fixed via this PR : https://github.com/contiv/netplugin/pull/519

Closing !

PR #506, opened 2016-08-10
fixing Issue#502

gaurav-dalvi:
This PR is to fix the issue : https://github.com/contiv/netplugin/issues/502

We need to avoid restart docker service and clearing up all containers, when we bring our nodes in swarm cluster. Swarm is being run as a container now.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -279,7 +279,7 @@ systemctl start etcd || exit 1
 systemctl start consul || exit 1
 
 # start swarm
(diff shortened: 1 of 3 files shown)


jainvipin:
@gaurav-dalvi - do we use python (tnode.py) in `host-restart` scripts?

gaurav-dalvi:
@jainvipin : Yes.

Please take a look at here :
https://github.com/contiv/netplugin/blob/master/Makefile#L180

PR #520, opened 2016-08-12
Adding check.vv flag to system and integration tests

gaurav-dalvi:
Adding check.vv flag to system and integration tests


shaleman:
@gaurav-dalvi Please keep the test start/end markers, so that its easy to see in the logs. LGTM otherwise

PR #521, opened 2016-08-14
minor makefile change to destroy 3 vms by default

abhinandanpb:
observed make build was bringing up 3vms and make stop was destroying only 2. This would cause etcd issues.


gaurav-dalvi:
LGTM. Thanks for taking care of this !

PR #522, opened 2016-08-14
Adding check.vv flag to system and integration tests

gaurav-dalvi:
adding check.vv flag to system and integration tests.


shaleman:
LGTM

PR #523, opened 2016-08-14
add go fmt/lint/vet checks in Makefile

rchirakk:
enable go fmt/lint/vet checks on all netplugin sources
fix minor warnings.


shaleman:
LGTM
Thanks for taking care of this

PR #531, opened 2016-08-18
 Modified runcontainers in TestNetprofileMultipleTenant

reshmagiliyal:


test/systemtests/docker_test.go
```diff
@@ -257,26 +257,26 @@ func (d *docker) startIperfClient(c *container, ip, limit string, isErr bool) er
...
-	success = true
-	break	
-	}else if strings.Contains(bw, "read failed:") { 
-	time.Sleep(2 * time.Second)
-	i++
-	}
-	}      
-
-		if success {
+	for i := 0; i < 10; i++ {
```
gaurav-dalvi: why hard coded number 10 ? 
Cant we make some variable for that ?

shaleman: just for my education, why do we need retries here?

shaleman:
LGTM other than minor comment..

test/systemtests/docker_test.go
abhinandanpb: @shaleman i believe there were cases where iperf was returning error.

PR #542, opened 2016-08-24
include mesos plugin in release tar.bz2

rchirakk:
spawn single vagrant node to create release files.


jainvipin:
@rchirakk - are you calling the binary `contivnet`? too generic?

rchirakk:
Yes, the current cni plug-in name is contivnet

erikh:
I already named the volume driver volcontiv, is it possible to be consistent across the projects?

jainvipin:
@erikh - volcontiv is what we agreed; is it also the name of binary, or the name of the plugin as it shows up as the driver name?
@rchirakk: either we produce on binary, or if they are different we must keep it less generic... what I was alluding to was that when we run the plugin it shows up in management systems as `netcontiv`

rchirakk:
@jainvipin @erikh 
binaries are different for net/vol, we could rename the net plugin to 
> netcontiv

so the config in scheduler looks like 
``
"ipAddress": {
     "networkName": "netcontiv",
     "labels": {
     }
``
& the the plugin binary name also will be 
> netcontiv

shaleman:
LGTM

Issue #540, opened 2016-08-23
network config in etcd not getting cleaned after removing the configuration in APIC and Contiv setup


pradvara:
- Network is created in 103.1.1.0/24 network:
[stack@aci-rack1 ~]$ netctl net list -t india
Tenant  Network  Nw Type  Encap type  Packet tag  Subnet        Gateway      IPv6Subnet  IPv6Gateway
------  -------  -------  ----------  ----------  -------       ------       ----------  -----------
india   net1     data     vlan        0           103.1.1.0/24  103.1.1.254

- Able to view the network through docker
[stack@aci-rack1 ~]$ docker network ls
NETWORK ID          NAME                        DRIVER
e8d943a5c817        aci-rack1/bridge            bridge
6c284567ecac        aci-rack1/docker_gwbridge   bridge
b61a47086153        aci-rack1/host              host
6c258d117c9a        aci-rack1/none              null
a6654c7394b5        aci-rack3/bridge            bridge
7861fb1ed3f8        aci-rack3/host              host
763de7088884        aci-rack3/none              null
ba588fa2740b        aci-rack4/bridge            bridge
a5b6fa72ddc5        aci-rack4/docker_gwbridge   bridge
1793285de524        aci-rack4/host              host
b8b9f3fbde3e        aci-rack4/none              null
270a3baf2adf        aci-rack5/bridge            bridge
966e765f0ed5        aci-rack5/docker_gwbridge   bridge
6d9ccccf3376        aci-rack5/host              host
15509418a12a        aci-rack5/none              null
d899e878e167        db/india                    netplugin

- Docker inspection shows different subnet
[stack@aci-rack1 ~]$ docker network inspect db/india
[
    {
        "Name": "db/india",
        "Id": "d899e878e167c7133c0fc596579526e079789ee1147e2a56140ad401761823ad",
        "Scope": "global",
        "Driver": "netplugin",
        "EnableIPv6": false,
        "IPAM": {
            "Driver": "netplugin",
            "Options": {
                "network": "net1",
                "tenant": "india"
            },
            "Config": [
                {
                    "Subnet": "101.1.1.0/24",
                    "Gateway": "101.1.1.254"
                }
            ]
        },
        "Internal": false,
        "Containers": {
            "36926a569c36a42c47c417075102352331726c5888590d1e66f0f170f9439f3e": {
                "Name": "app1",
[...]

PR #544, opened 2016-08-25
Resolving Issue#540

gaurav-dalvi:
This is regarding two Issues: 

https://github.com/contiv/netplugin/issues/411
https://github.com/contiv/netplugin/issues/540.

Added UT as well.


jojimt:
LGTM

DivyaVavili:
@jojimt: As I was discussing, could you confirm if we attach to skydns container in ACI mode? Looks like with this change, there might be some asymmetric behavior. During network create we are not attaching to DNS container, but during network delete a detach from DNS container might be attempted.

Issue #508, opened 2016-08-10
cannot assign pkt-tag that is greater than 4096 to a vxlan virtual network


yuxiangning:
I met an error when playing with the three nodes cluster.

vagrant@netplugin-node1 ~]$ netctl network create --subnet=100.0.2.0/24 --encap=vxlan --pkt-tag=5000 vtep5000
ERRO[0000] pktTag Value Out of bound

$ netctl global info
Fabric mode: default
Vlan Range: 1-4094
Vxlan range: 1-10000

mhrobinson:
I don't see this issue anymore.  Is it consistently reproducible?

I tried on https://github.com/contiv/netplugin/commit/ff611a2f064496246fa724fe64f46ac9586d4bf4

g1rana:
Looks like issue has been fix'd  from earlier release . Issue doesn't seem to be exist in  current release . Have tried follow version : 

[vagrant@netplugin-node1 ~]$ netctl -v
netctl version
Version: 1.0.0-beta-07-26-2017.21-06-22.UTC
GitCommit: 1e2213f
BuildTime: 07-26-2017.21-06-22.UTC

Here are some logs outputs : 
=====
 netctl global info
Fabric mode: default
Forward mode: bridge
ARP mode: proxy
Vlan Range: 1-4094
Vxlan range: 1-10000
Private subnet: 172.19.0.0/16
=====
netctl network create --subnet=100.0.2.0/24 --encap=vxlan --pkt-tag=5000 vtep5000
Creating network default:vtep5000

[vagrant@netplugin-node1 mhrobinson]$ netctl network ls
Tenant   Network   Nw Type  Encap type  Packet tag  Subnet        Gateway   IPv6Subnet  IPv6Gateway
------   -------   -------  ----------  ----------  -------       ------    ----------  -----------
default  vtep5000  data     vxlan       5000        100.0.2.0/24
=====

dseevr:
Looks like this was fixed 6 days after this issue was reported:  https://github.com/contiv/contivmodel/pull/40

PR #546, opened 2016-08-26
minor svcstats & sanity fix

shaleman:
- service stats fix
- ssh slowness fix for ubuntu

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -67,6 +67,10 @@ sudo systemctl start docker
 rm /etc/docker/key.json
 (service docker restart) || exit 1
 
+# fix SSH slowness
+sed -i -e 's/UsePAM yes/UsePAM no/g' /etc/ssh/sshd_config
+(service sshd restart) || exit 1
+
 usermod -aG docker vagrant
 docker load --input #{gopath_folder}/src/github.com/contiv/netplugin/scripts/dnscontainer.tar || echo "Loading skydns container failed"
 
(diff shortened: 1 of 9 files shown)


netmaster/objApi/objapi_test.go
```diff
@@ -1002,6 +1002,10 @@ func TestNetworkAddDelete(t *testing.T) {
 	verifyNetworkState(t, "default", "contiv", "data", "vxlan", "10.1.1.1", "10.1.1.254", 16, 1, 1, "", "", 0)
 	checkDeleteNetwork(t, false, "default", "contiv")
 
+	// vxlan network with VNI higher than 4K
+	checkCreateNetwork(t, false, "default", "contiv", "", "vxlan", "10.1.1.1/24", "10.1.1.254", 8000, "", "")
+	checkDeleteNetwork(t, false, "default", "contiv")
+
```
gaurav-dalvi: Is this related to Issue : https://github.com/contiv/netplugin/issues/508

Vagrantfile
```diff
@@ -67,6 +67,10 @@ sudo systemctl start docker
 rm /etc/docker/key.json
 (service docker restart) || exit 1
 
+# fix SSH slowness
+sed -i -e 's/UsePAM yes/UsePAM no/g' /etc/ssh/sshd_config
+(service sshd restart) || exit 1
+
```
gaurav-dalvi: are we going to do this for all OSes ? or just for ubuntu ?

gaurav-dalvi:
apart from some minor comments, LGTM

netmaster/objApi/objapi_test.go
shaleman: yes. it fixes #508

Vagrantfile
shaleman: yeah, for all OSes. We don't see any issues with Centos with this change.

gaurav-dalvi:
@jojimt @abhinandanpb : Do we still need this ? If yes then we need to submit new PR for this.

netmaster/objApi/objapi_test.go
mhrobinson: It looks like this commit only tests for https://github.com/contiv/netplugin/issues/508 but doesn't fix it.  I'm not seeing the issue though, so probably some other commit fixed it?

PR #549, opened 2016-08-30
Add a tool to modify cfg objects

shaleman:
This adds a tool to modify cfg state objects behind netplugin's back.
Its meant to be used as debug tool..


jojimt:
Nice! LGTM

PR #553, opened 2016-09-07
Adapting to new gobgp

abhinandanpb:
1) moving to newer version of gobgp
2) l3 test case optimizations
3) uploaded new gobgp binary in box


shaleman:
LGTM

PR #557, opened 2016-09-12
k8s contrib dependancy

abhinandanpb:
1) Pulling contrib if not present in local dir.
2)  updating to latest netplugin version

### DIFF
--- a/vagrant/k8s/setup_cluster.sh
+++ b/vagrant/k8s/setup_cluster.sh
@@ -48,15 +48,28 @@ function GetContiv {
   fi
 }
 
+#GetContrib fetches contrib if not present
+
+function GetContrib {
+	pushd .
+	if [ -f $top_dir/contrib ]; then
+    		echo "netplugin-$contivVer.tar.bz2 found, not fetching."
+  	else
+		git clone https://github.com/jojimt/contrib -b contiv 
+  	fi
+	popd
+}
+
 # kubernetes version to use -- defaults to v1.2.3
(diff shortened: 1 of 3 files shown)


vagrant/k8s/setup_cluster.sh
```diff
@@ -48,15 +48,28 @@ function GetContiv {
...
+  	fi
+	popd
+}
+
 # kubernetes version to use -- defaults to v1.2.3
 : ${k8sVer:=v1.2.3}
 
 # contiv version
-: ${contivVer:=v0.1-05-05-2016.18-38-56.UTC}
+: ${contivVer:=v0.1-09-08-2016.20-56-40.UTC}
```
gaurav-dalvi: Can we make this env variable or something ?
Right now its hard coded.

gaurav-dalvi:
LGTM

vagrant/k8s/setup_cluster.sh
abhinandanpb: no sure what you mean by its hard coded. This is the "env variable". User can export contivVer to any version of his/her choice. If its not exported then it defaults to the above version.

gaurav-dalvi: Got it. Thanks ! Usually after some time we might need to update this default version. We have been facing some issues in net_demo_installer script as it has some old hard coded version but you can set that env variable and change it to new version.

Anyways this looks good then !

PR #559, opened 2016-09-16
Free resources if ep create fails

jojimt:


netmaster/master/netmaster_test.go
```diff
@@ -688,3 +703,94 @@ func TestGetAllocatedIPs(t *testing.T) {
...
+		log.Fatalf("unable to locate network: %s", networkID)
+	}
+
+	allocIPs := ListAllocatedIPs(nwCfg)
+	if allocIPs != expIPs {
+		log.Fatalf("Got addresses '%s' expected '%s'", allocIPs, expIPs)
+	}
+}
+
+func TestEPCreate(t *testing.T) {
```
shaleman: We should deprecate tests based on old style intent.
It might be better to move this test to objApi_test.go which already has logic to setup resource manager etc.

PR #560, opened 2016-09-16
Bgp inspect change and minor fixes

abhinandanpb:
1) latest gobgp v1.11 to get some gobgp fixes
2) fixes bgp inspect
3) moved l3 test cases to use inspect instead of gobgp binary
4) Fixes concurrent map error
5) other minor mods/cleanups/fixes


shaleman:
LGTM

PR #561, opened 2016-09-16
Fix for service ref count issue

abhinandanpb:
Added missing link-set between service and network which was causing ref count issues.

### DIFF
--- a/netmaster/master/servicelb.go
+++ b/netmaster/master/servicelb.go
@@ -16,13 +16,14 @@ limitations under the License.
 package master
 
 import (
+	"reflect"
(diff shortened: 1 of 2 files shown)


gaurav-dalvi:
Do we have test for this already ?

abhinandanpb:
@gaurav-dalvi : for this specific issue ? nope. Decided to take a 2 days break from test cases :) 
It will be part of my next commit have changes for global settings and I am writing cases for them.

gaurav-dalvi:
no problem. as long as we have test case for this, its fine !
LGTM

PR #568, opened 2016-09-28
Global resource resize changes

abhinandanpb:
Changes to : 
1) Allows global settings like vlan and vxlan ranges
2) Unit tests
3) lock for global config modifications
4) Minor fixes

### DIFF
--- a/netmaster/gstate/gstate.go
+++ b/netmaster/gstate/gstate.go
@@ -18,6 +18,9 @@ package gstate
 import (
 	"encoding/json"
 	"errors"
+	"strconv"
+	"strings"
+	"sync"
 
 	"github.com/jainvipin/bitset"
 
@@ -36,6 +39,9 @@ const (
 	operGlobalPath      = operGlobalPrefix + "global"
 	vxlanLocalVlanRange = "1-4094"
 )
+
+//GlobalMutex used to syncronize global configuration changes
+var GlobalMutex sync.Mutex
 
 // AutoParams specifies various parameters for the auto allocation and resource
 // management for networks and endpoints.  This allows for hands-free
@@ -149,7 +155,7 @@ func (g *Oper) Clear() error {
 	return g.StateDriver.ClearState(key)
 }
 
-func (gc *Cfg) initVXLANBitset(vxlans string) (*resources.AutoVXLANCfgResource, uint, error) {
+func (gc *Cfg) initVXLANBitset(vxlans string) (*resources.AutoVXLANCfgResource, error) {
 
 	vxlanRsrcCfg := &resources.AutoVXLANCfgResource{}
 	vxlanRsrcCfg.VXLANs = netutils.CreateBitset(14)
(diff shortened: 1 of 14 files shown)


netmaster/gstate/gstate.go
```diff
@@ -436,3 +488,40 @@ func (gc *Cfg) UnassignNetwork(networkName string) error {
...
+			maxUsed, _ = strconv.Atoi(strings.TrimSpace(tags[len(tags)-1]))
+		}
+	}
+
+	r, err := netutils.ParseTagRanges(ranges, pktTagType)
+	if err != nil {
+		return false
+	}
+
+	if r[0].Min > minUsed || r[0].Max < maxUsed {
```
jojimt: 1) Do we allow disjoint  subranges within ranges? If so, should we be looking at r[len(r)-1].Max?
2) Should minUsed and maxUsed be swapped in the above expression to ensure ranges falls outside the used range?

netmaster/master/netmaster.go
```diff
@@ -159,11 +162,105 @@ func CreateGlobal(stateDriver core.StateDriver, gc *intent.ConfigGlobal) error {
...
+	// check for valid values
+	if gc.NwInfraType != "" {
+		switch gc.NwInfraType {
+		case "default", "aci", "aci-opflex":
+			// These values are acceptable.
+		default:
+			return errors.New("Invalid fabric mode")
+		}
+		masterGc.NwInfraType = gc.NwInfraType
+	}
```
jojimt: Would be good to do this validation before we take the lock.

netmaster/objApi/objapi_test.go
```diff
@@ -1081,6 +1082,124 @@ func TestNetworkAddDelete(t *testing.T) {
...
+	checkCreateNetwork(t, true, "default", "contiv5", "", "vlan", "14.1.1.1/24", "14.1.1.254", 302, "", "")
+
+	//verify creation of network succeeds after vlan change
+	checkCreateNetwork(t, false, "default", "contiv6", "", "vlan", "15.1.1.1/24", "15.1.1.254", 299, "", "")
+	checkInspectNetwork(t, false, "default", "contiv6", "15.1.1.254", 299, 0)
+	//checkInspectGlobal(t, false, "4", "")
+	verifyNetworkState(t, "default", "contiv6", "data", "vlan", "15.1.1.1", "15.1.1.254", 24, 299, 0, "", "", 0)
+
+	//check global state change fails if there are active tags
+	checkGlobalSet(t, true, "default", "50-301", "1-10000", "bridge")
```
jojimt: we should also check the auto vlan allocation case.

netmaster/gstate/gstate.go
abhinandanpb: We dont support disjoint ranges for vxlan at the moment.  Unless there is a need due the amount of vxlan range space available. We can enhance it should need be

netmaster/master/netmaster.go
abhinandanpb: sure we can do that. But we are anyways exiting if invalid? But I can make the change

netmaster/objApi/objapi_test.go
abhinandanpb: Sure I can add them

PR #578, opened 2016-10-13
Fixes #541

### LINKED ISSUE #541: netctl network inspect does not show skydns container name
for e.g 

```
"Oper": {
    "allocatedAddressesCount": 3,
    "allocatedIPAddresses": "10.2.1.1-10.2.1.3",
    "availableIPAddresses": "10.2.1.4-10.2.1.254",
    "dnsServerIP": "10.2.1.1",
    "endpoints": [
      {
        "endpointID": "a84aa9f0556fb16fd4b208f89e4a389a8853d4bd441a6522a25258d1017b831a",
        "homingHost": "netplugin-node1",
        "ipAddress": [
          "10.2.1.1",
          ""
        ],
        "labels": "map[]",
        "macAddress": "02:02:0a:02:01:01",
        "network": "n1.t1"
      },
      {
        "containerID": "d37345ab7e3705692bb1a62346c3752d3733ba91a1ba7501ad4b967e992ba368",
        "containerName": "/prickly_easley",
        "endpointGroupId": 1,
        "endpointGroupKey": "g1:t1",
        "endpointID": "73db8f90fcb1bd70d55581cb78ed07966a75771be25eae6219e3af149fbfed7c",
        "homingHost": "netplugin-node1",
        "ipAddress": [
          "10.2.1.2",
          ""
        ],
        "labels": "map[]",
        "macAddress": "02:02:0a:02:01:02",
        "network": "n1.t1",
        "serviceName": "g1"
      },

```

It should show the name of container having endpoint ID : `a84aa9f0556fb16fd4b208f89e4a389a8853d4bd441a6522a25258d1017b831a`

srampal:


netctl/netctl.go
```diff
@@ -65,7 +65,7 @@ func inspectPolicy(ctx *cli.Context) {
 	tenant := ctx.String("tenant")
 	policy := ctx.Args()[0]
 
-	fmt.Printf("Inspeting policy: %s tenant: %s\n", policy, tenant)
+	fmt.Printf("Inspecting policy: %s tenant: %s\n", policy, tenant)
```
abhinandanpb: I think its time to make this a debug log. .

netplugin/agent/docker_event.go
```diff
@@ -32,7 +32,7 @@ import (
 // container start and die event*/
 func handleDockerEvents(event *dockerclient.Event, ec chan error, args ...interface{}) {
 
-	log.Debugf("Received Docker event: {%#v}\n", *event)
+	log.Infof("Received Docker event: {%#v}\n", *event)
```
abhinandanpb: Please make this debug log.

netplugin/agent/docker_event.go
```diff
@@ -94,7 +95,69 @@ func handleDockerEvents(event *dockerclient.Event, ec chan error, args ...interf
 		if err != nil {
 			log.Errorf("Event:'die' Http error posting endpoint update, Error:%s", err)
 		}
-	}
+
+	case "":
```
abhinandanpb: I am not sure if it will be right to handle every docker event case. We might have to get to root cause. If we have to start container(samalba) with additional parameters

netplugin/agent/docker_event.go
```diff
@@ -94,7 +95,69 @@ func handleDockerEvents(event *dockerclient.Event, ec chan error, args ...interf
...
+		containerID := event.Actor.Attributes["container"]
+		networkID := event.Actor.ID
+
+		if event.Action == "connect" && containerID != "" {
+
+			endpointUpdReq.Event = "connect"
+			defaultHeaders := map[string]string{"User-Agent": "engine-api-cli-1.0"}
+			cli, err := client.NewClient("unix:///var/run/docker.sock", "v1.21", nil, defaultHeaders)
+			if err != nil {
+				panic(err)
```
abhinandanpb: Any Reason to panic if we cant reach the docker client ? I would think logging an error may be the right way to go about it instead of exiting netplugin.

test/systemtests/network_test.go
```diff
@@ -81,6 +83,36 @@ func (s *systemtestSuite) testInfraNetworkAddDelete(c *C, encap string) {
 	}
 }
 
+/* Confirm network inspect includes a dns endpoint (assumes this is the only endpoint on the network currently) */
+func (s *systemtestSuite) testNetworkInspectDNS(tenant, network string) (bool, error) {
```
abhinandanpb: This will need to be move to util_test.go in test/systemtest/

test/systemtests/network_test.go
```diff
@@ -81,6 +83,36 @@ func (s *systemtestSuite) testInfraNetworkAddDelete(c *C, encap string) {
...
+		return false, errors.New("testNetworkInspectDNS has incorrect # of endpoints (should have 1)!")
+	}
+
+	// First endpoint in network inspect should not have a null container ID
+	if netInspect.Oper.Endpoints[0].ContainerID == "" {
+		return false, errors.New("testNetworkInspectDNS endpoint has null containerID!")
+	}
+
+	// First endpoint in network inspect should incude dns in the container name
+	if !strings.Contains(netInspect.Oper.Endpoints[0].ContainerName, "dns") {
```
abhinandanpb: we have a convention to name skydns container. Which is a combination of tenant and skydns. It might be better to check that instead of just "dns"

test/systemtests/network_test.go
```diff
@@ -81,6 +83,36 @@ func (s *systemtestSuite) testInfraNetworkAddDelete(c *C, encap string) {
...
+		return false, errors.New("testNetworkInspectDNS has no endpoint with dns in containerName!")
```
abhinandanpb: might be add info to  error like which network and tenant is this error for. It will be easy to check errors in sanity logs

test/systemtests/network_test.go
```diff
@@ -279,6 +367,10 @@ func (s *systemtestSuite) TestNetworkAddDeleteTenantVLAN(c *C) {
 	s.testNetworkAddDeleteTenant(c, "vlan", s.fwdMode)
 }
 
+func (s *systemtestSuite) TestNetworkAddDeleteTenantWithDns(c *C) {
```
abhinandanpb: same comment as above

netplugin/agent/docker_event.go
DivyaVavili: @srampal: You could change the switch case from event.Status to event.Action, which is more specific. And even for the start case, you could fetch the container info from event.Actor.Attributes["container"] instead of event.ID. Having made these changes, both the connect and start events seem to be doing the same thing. I suggest abstracting it out(preferred) or making a fall-through case statement.

srampal:
We could do that but not sure if someone expects to scrape the CLI output instead of using the API. If yes, removing some lines from CLI output could impact their screen scraping logic … but maybe there aren’t any customers that do that OR we anyway do not support anyone to do CLI screen scraping since they should only be using API for automation.

Also, this is not the only place … fmt.Printf() is used throughout this file in places where a debug log may have been appropriate. Should there be a separate issue opened to track which all Printf()s should be changed to logs ?

netctl/netctl.go
srampal: We could do that but not sure if someone expects to scrape the CLI output instead of using the API. If yes, removing some lines from CLI output could impact their screen scraping logic … but maybe there aren’t any customers that do that OR we anyway do not support anyone to do CLI screen scraping since they should only be using API for automation. Also, this is not the only place … fmt.Printf() is used throughout this file in places where a debug log may have been appropriate. Should there be a separate issue opened to track which all Printf()s should be changed to logs ?

netplugin/agent/docker_event.go
srampal: We aren't handling every docker event case. We are specifically looking for events that have nil status + action of "connect". Given that different users may be using different versions of api to create containers, the docker call backs could be formatted differently. With this check, we are now including the call back which we were ignoring when skydns container was created.

So I think we are good here unless you are seeing something else.

netplugin/agent/docker_event.go
srampal: Ok. I duplicated the logic from the "create" case above this was doing (it did a panic() there as well). But will change both to logging an error and exiting instead of panic().

test/systemtests/network_test.go
```diff
@@ -89,6 +121,10 @@ func (s *systemtestSuite) TestNetworkAddDeleteVLAN(c *C) {
 	s.testNetworkAddDelete(c, "vlan")
 }
 
+func (s *systemtestSuite) TestNetworkAddDeleteWithDns(c *C) {
```
srampal: This particular issue was just to make sure the skydns container shows up in netctl network inspect. So the encap on the network side doesnt really matter. If we want to create a separate dedicated set of tests for DNS operation itself, that should probably a new issue and there we can test both VLAN and VXLAN mode. So maybe testing separately for VLAN and VXLAN can be skipped here since it is independent of this particular issue.

test/systemtests/network_test.go
srampal: See above

netplugin/agent/docker_event.go
srampal: there are some differences but yet a lot can be commonized so will do that.

DivyaVavili: @srampal: Like I mentioned, event.Action should help here

test/systemtests/network_test.go
abhinandanpb: I think i am sure :) The way test suite is designed is 
1) Testcase - basically consits of configs usage and calls util functions to help execute/verify/ the case. Test cases are supposed to be in respective test files.
2) Util functions: like the one you have added need to be in utils. This creates an abstraction on top of the kind of test environment you are running your test case. Your code will break k8s sanity (currently turned off due to a vagrant version issue). 
3) Actual test environment executors: Util functions will call these function depending on the test environment. If its k8s then calls k8s specific executors.If its docker then it calls docker specific executors. 

So you will need to move this to util all the way to executors. This will be a stub for k8s. I hope i have clarified the intention.

test/systemtests/network_test.go
abhinandanpb: The reason i am suggesting the change is - its good to verify that the right skydns container is attached to the network/tenant. If we are validating skydns. We might as well validate the right skydns container is attached. We should use the same mechanism to test other dns containers (if and whenever we decide to do that) so that tenant name is embedded in the dns name.

test/systemtests/util_test.go
```diff
@@ -1003,6 +1003,36 @@ func (s *systemtestSuite) verifyIPs(ipaddrs []string) error {
 	return err
 }
 
+/* Confirm network inspect includes a dns endpoint (assumes this is the only endpoint on the network currently) */
+func (s *systemtestSuite) testNetworkInspectDNS(tenant, network string) (bool, error) {
```
abhinandanpb: very very very minor nit - can we change this is check instead of test.

netctl/netctl.go
abhinandanpb: If possible you can remove the printf's

abhinandanpb:
LGTM other than minor comments. Thanks for taking care of the comments. Please squash all your commits into one.

DivyaVavili:
LGTM. Please squash and resubmit.

test/systemtests/aci_util.go
```diff
@@ -14,7 +14,7 @@ import (
 // EPSpec for aci-gw
 type EPSpec struct {
 	Tenant string `json:"tenant,omitempty"`
-	App    string `json:"app,omitempty"`
```
abhinandanpb: @srampal i hope this change was intentional ?

jojimt:
This should not have been merged. It failed the test. Who merged it and how?

test/systemtests/aci_util.go
gaurav-dalvi: @abhinandanpb : This is not related to the issue we are seeing here. This was found in my aci-gw feature. As it was online change I asked Sanjeev to incorporate in this PR.

srampal:
My bad .. I intended to do squash-commit and apparently ended up doing squash-commit-merge or something like that …

I had already tested my changes several times. The additional 1-line change was from Gaurav.


Rgds,
Sanjeev

abhinandanpb:
@srampal the top of the tree is broken with compile error. Can you please take care of it ?

PR #582, opened 2016-10-25
Solving Issue#573

gaurav-dalvi:
Signed-off-by: Gaurav Dalvi <gaurav1424@gmail.com>

### DIFF
--- a/scripts/python/api/tnode.py
+++ b/scripts/python/api/tnode.py
@@ -105,7 +105,7 @@ class Node:
 
     # Remove all containers on this node
     def cleanupContainers(self):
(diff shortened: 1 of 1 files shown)


scripts/python/api/tnode.py
```diff
@@ -105,7 +105,7 @@ def cleanupDockerNetwork(self):
 
     # Remove all containers on this node
     def cleanupContainers(self):
-        self.runCmd("docker ps -a | grep alpine | awk '{print $1}' | xargs -r docker rm -fv ")
+        self.runCmd("docker ps -a | grep -v swarm | awk '{print $1}' | xargs -r docker rm -fv ")
```
jojimt: Should we explicitly kill the alpine and the skydns? We could have more infra containers other than swarm.

PR #586, opened 2016-10-27
Makefile: update 'clean' target to delete the netplugin pkg directory rather than the root pkg directory

dseevr:
Signed-off-by: Bill Robinson dseevr@users.noreply.github.com

cc: @dcheperd @jojimt

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,7 +87,7 @@ build:
 	make stop
 
 clean: deps
-	rm -rf $(GOPATH)/pkg/*
+	rm -rf $(GOPATH)/pkg/*/github.com/contiv/netplugin/
 	go clean -i -v ./...
 
 update:


Makefile
```diff
@@ -87,7 +87,7 @@ build:
 	make stop
 
 clean: deps
-	rm -rf $(GOPATH)/pkg/*
```
abhinandanpb: This is will not work on mac . You might want to check for the os before using linux_amd64 vs darwin_amd64 ?

dseevr: We could just wildcard it, but I thought netplugin only compiled and ran on Linux because of the `x/sys/unix` library?

abhinandanpb: well netplugin runs only on linux. But for development we would could be on mac, ubuntu,centos etc. clean would be run from ur ws. 
$ make clean
./scripts/deps
rm -rf /Users/aprativa//pkg/*
go clean -i -v ./...

dseevr: Updated the commit to use a wildcard for the OS/arch directory... this comment didn't get collapsed because I didn't edit the specific line you commented on.

abhinandanpb: Thanks adding this change ! LGTM

ppadala: +1 to @abhinandanpb's comment. otherwise LGTM

dseevr:
@abhinandanpb can you merge this, too? 😄

PR #589, opened 2016-10-28
Makefile,Vagrantfile,scripts: fix systemd controls and allow docker version to become variable

erikh:
note that this also gets rid of the bespoke systemd units since they no longer work with 1.12 and up.

The sed should suffice for 1.11 too but I did not test it.

I'm sending the units/sanities to jenkins because it takes forever :)

PTAL and let me know if changes are required.


scripts/docker-tcp.socket
```diff
@@ -1,10 +0,0 @@
-[Unit]
```
gaurav-dalvi: I hope this wont cause trouble in out docker swarm setup. When you do `make demo` it brings up 3 node docker swarm cluster.

erikh:
Sanities are running but not passing, can someone look this over and tell me what's breaking so I can fix it?

erikh:
@abhinandanpb sanities are still breaking here after rebase; I'm going to re-push but the old log is here: http://contiv.ngrok.io/job/Netplugin%20Pull-Request%20Build%20Master/1117/ if you want to review it in case it's a heisenbug.

abhinandanpb:
@erikh some issue with sanity server.

erikh:
@abhinandanpb can we look into this tomorrow? third time I've pushed today and the third failure, I think something else is happening here that neither of us are seeing.

abhinandanpb:
ping @rchirakk . Any idea whats happening here ? Looks like its complaining about image pull issue in mesos test case.

abhinandanpb:
@erikh something to do with docker version ? Looks like mesos test case is not liking the docker version. @rchirakk might be a better person to answer that.

erikh:
@abhinandanpb ok; lmk if I can help debug somehow.

abhinandanpb:
@erikh this might be related to docker client not supporting/updated to support docker 1.12.

erikh:
sure; um, ok.
what do you want to do with this PR then? Are the mesos tests isolated?

abhinandanpb:
@rchirakk will try out a quick fix to use official dockerclient instead of samalba.

erikh:
I could also try upgrading the provisioned mesos to see if that'd help.

rchirakk:
@erikh, another option is disable mesos tests.
I can move it to docker client in another PR.

erikh:
nobody is in a rush. Let's figure out the mesos situation, it's probably a build time thing anyways.

I'll ping you on slack.

erikh:
after discussing with @rchirakk we discovered that the docker client used by our mesos integration is an older client that is incompatible with 1.12. I have advised @rchirakk to utilize the docker/docker client instead, which is the canonical one now.

Until then I don't think this can be merged.

rchirakk:
disabled tests in https://github.com/contiv/netplugin/pull/595, will enable it later with docker/docker client

erikh:
Thanks.

rchirakk:
@erikh please rebase and try

erikh:
There seems to be a problem installing 1.12.3 on ubuntu so I'm investigating that first.

gaurav-dalvi:
@erikh : Could you please remove the change in init_test.go and utils_test.go file ? I see those changes are in other PRs as well . (https://github.com/contiv/netplugin/pull/610)

erikh:
I will do that before this is ready to merge. I will comment here then.

erikh:
@jojimt this is passing and validates docker 1.12.3. Ready to merge, but I cleaned up the commits from #613 so this would be merged independently, so it's running the tests again.

erikh:
Passing and ready.

PR #594, opened 2016-11-01
test/systemtests: remove the trigger_test for Netplugin Disconnects.

erikh:

### DIFF
--- a/test/systemtests/trigger_test.go
+++ b/test/systemtests/trigger_test.go
@@ -86,47 +86,51 @@ func (s *systemtestSuite) TestTriggerNetmasterSwitchover(c *C) {
 }
 
 func (s *systemtestSuite) TestTriggerNetpluginDisconnect(c *C) {
-	network := &client.Network{
-		TenantName:  "default",
-		NetworkName: "private",
-		Subnet:      "10.1.0.0/16",
-		Gateway:     "10.1.1.254",
-		PktTag:      1001,
-		Encap:       "vxlan",
-	}
-	if s.fwdMode != "routing" {
-		network.Ipv6Subnet = "2016:0617::/100"
-		network.Ipv6Gateway = "2016:0617::254"
-	}
-	c.Assert(s.cli.NetworkPost(network), IsNil)
+	c.Skip("Not working")
+	return
+	/*
+		network := &client.Network{
+			TenantName:  "default",
+			NetworkName: "private",
+			Subnet:      "10.1.0.0/16",
+			Gateway:     "10.1.1.254",
+			PktTag:      1001,
+			Encap:       "vxlan",
+		}
+		if s.fwdMode != "routing" {
+			network.Ipv6Subnet = "2016:0617::/100"
+			network.Ipv6Gateway = "2016:0617::254"
+		}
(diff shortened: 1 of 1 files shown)


gaurav-dalvi:
Why remove ? 

we can just skip as of now. We can skip just like this :

https://github.com/contiv/netplugin/blob/2f46153da9f60444ca711c261401072e6da51fc3/test/systemtests/bgp_test.go#L23

erikh:
we can do that. I was under the impression these tests weren't worth fixing?

jojimt:
I also prefer not disabling/removing tests. Can we investigate the failure instead?

erikh:
SGTM, I was just trying to help.

srampal:
Yes, the goal is to fix the tests …

However at least for the last 2-3 weeks (and probably earlier since I wasn’t committing any code earlier), these have been slowing down all sanity builds so apparently these are not easily fixable ?

If so … it may be better to *temporarily* remove/ disable them and concurrently open an issue to fix them properly. Otherwise, they keep slowing down the team’s commit process significantly.


Rgds,
Sanjeev

jojimt:
Disabling a test should be our last option.

srampal:
Agreed that tests should not be removed if they can be fixed *soon*.  But if they cant be fixed soon, maybe they need to be temporarily disabled since they are impacting productivity significantly.

This situation has existed at least for 2-3 weeks since I tried commits and probably existed well before that even. So if this has existed for several weeks, either these cant be fixed easily or we are okay with the impact on productivity. For my prior PR, it took 4 attempts of Jenkins run (4+ hours each) before I got a clean run. For the PR I am trying today, I am already on my 3rd attempt at getting clean Jenkins run ... the first 2 times it failed in one of these flaky test cases that has nothing to do with my change. Hence imo either these test cases need to be fixed asap or need to be temporarily disabled while a proper fix is found. Are others not seeing similar delays ?

srampal:
My latest run failed again (after running for 4 hours) on some unrelated issue (this time it was the test case TestTriggerNetPartition). So its usually one of these TestTrigger test cases that fails very often and is unrelated to the actual code change.

erikh:
I'm going to close this; there's no point in perpetuating it for now. Let's discuss holistically how we want to approach tests with bugs in them separately.

srampal:
If someone commits a change that causes 60-75% of sanity runs to fail incorrectly (and each of these runs takes 4 hours to execute), would we not have had them back out that change and fix it correctly ?


Its the same thing … there are test cases in main which are causing 4 out of 5 sanity runs to fail (at least that’s the failure rate I happen to be seeing). These should be backed out and fixed properly before recommitting. Maybe the real fix is our build server infra combined with flaky test cases … whatever the root cause, we shouldn’t let a situation continue where we have to atempt 4 runs of 4 hours each in order to get one successful run.

erikh:
Fix them?

jojimt:
@srampal If the test code is the cause of the failure, one of us need to fix it. As late as 10 days ago, we had no issues with this test. I don't think there is sufficient urgency to justify taking out critical test cases. We have never done that before. If you're able to repro the failure without your changes and you're unable to track it down, please open an issue and wait for one of us to investigate and fix,

PR #595, opened 2016-11-01
disable mesos tests until docker client is integrated.

rchirakk:
temporarily disable mesos it test to unblock docker 1.12 upgrade (https://github.com/contiv/netplugin/pull/589)

### DIFF
--- a/test/integration/mesos_test.go
+++ b/test/integration/mesos_test.go
@@ -340,9 +340,10 @@ func (t *intgFmt) Format(e *log.Entry) ([]byte, error) {
 	return nt.Format(e)
 }
 
-// TestMesosCniEndpoints : test cni endpoints
+// TestMesosCniEndpoints : test cni endpoints, disabled to move to new docker client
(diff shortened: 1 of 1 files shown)


gaurav-dalvi:
we can just skip as of now. We can skip just like this :

https://github.com/contiv/netplugin/blob/2f46153da9f60444ca711c261401072e6da51fc3/test/systemtests/bgp_test.go#L23

rchirakk:
It is only for 1 test (in loop), not much change required to disable.

gaurav-dalvi:
Please open up another issue to enable these tests until problem is solved. Just for tracking purpose.

gaurav-dalvi:
Please use c.Skip to skip the test. Thats how we have been doing for all other tests to disable them.

Apart from that LGTM.

ppadala:
+1 to Gaurav's comments. Otherwise, this is good to go

rchirakk:
done

PR #596, opened 2016-11-02
Fixing Netplugin disconnect 

abhinandanpb:
NetpluginDisconnect trigger exposed issue with vxlan vrep creation with ofport not found error.
First suspicion was ovsdb operation being async would need more retries. Also cherry picked sukhesh's commit in this area.

### DIFF
--- a/drivers/ovsdbDriver.go
+++ b/drivers/ovsdbDriver.go
@@ -29,7 +29,7 @@ import (
 )
 
 // Max number of retries to get ofp port number
-const maxOfportRetry = 10
+const maxOfportRetry = 20
 
 // OvsdbDriver is responsible for programming OVS using ovsdb protocol. It also
 // implements the libovsdb.Notifier interface to keep cache of ovs table state.


drivers/ovsdbDriver.go
```diff
@@ -617,33 +617,33 @@ func (d *OvsdbDriver) GetOfpPortNo(intfName string) (uint32, error) {
...
-						switch t := value.(type) {
-						case string:
-							return true, t
-						default:
-							// return false, ""
-						}
-					}
-				default:
-					// return false, ""
+	row, err := d.ovs.Transact(ovsDataBase, selectOp)
```
jojimt: Should we keep the cache look up and fallback to Transact only if there is a miss?

abhinandanpb: I believe the issue was with cache not being updated. So it wasnt a miss that was causing the issue. The cache had the vtep but not the ovsdb.

drivers/ovsdbDriver.go
```diff
@@ -617,33 +617,33 @@ func (d *OvsdbDriver) GetOfpPortNo(intfName string) (uint32, error) {
...
+
+	if err == nil && len(row) > 0 && len(row[0].Rows) > 0 {
+		value := row[0].Rows[0]["ofport"]
```
ppadala: what happens if Rows[0] doesn't have ofport key?

jojimt:
@abhinandanpb as discussed, let's change only the retry for now.

abhinandanpb:
ping @jojimt

PR #601, opened 2016-11-03
Fixing issue 597 

abhinandanpb:
Fix to alloc vlan after network creation phase has successfully reached the end.


gaurav-dalvi:
LGTM.

PR #602, opened 2016-11-03
close socket to avoid fd leaks.

rchirakk:
close socket to avoid fd leaks.


jojimt:
LGTM. Thanks for catching this.

unclejack:
LGTM

PR #608, opened 2016-11-04
drivers/ovsdbDriver.go: fix govet UUID

unclejack:


unclejack:
This is green now. 

ping @jojimt @abhinandanpb @erikh @gaurav-dalvi

abhinandanpb:
LGTM. Any reason why this was not hit when the PR that changed this went thru the sanity run ?

unclejack:
@abhinandanpb I don't know right now. I've hit it on my host during the checks.

dseevr:
@unclejack  Weird, I just tested the `[]Foo{{"bar"}}` declaration style with `go vet` on 1.7.3 and 1.5.0 and neither of them complained about it.  I'm curious what the exact error output you got was.

erikh:
Yeah I'm seeing it too, I think it's a golint/go vet version issue

unclejack:
@erikh is probably right. I've seen this happen after switching to Go 1.7.3.

erikh:
Let's merge this if we can today, #610 depends on this which I hope to have ready by EOW.

PR #610, opened 2016-11-08
test/systemtests: test improvements

erikh:
* vastly improving test perf across the board, including BGP modes
* constants for forwarding mode and encapsulation type
* predicates for determining if we're in BGP mode
* BGP mode calls no longer need gating (they do the gating themselves)

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -21,6 +21,9 @@ cluster_ip_nodes = ""
 MEMORY = "4096"
 
 provision_common_once = <<SCRIPT
+sudo sed -i.bak -e 's/^MaxStartups.*$/MaxStartups 1000:100:1000/g' /etc/ssh/sshd_config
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -17,6 +17,80 @@ gopath_folder="/opt/gopath"
 
 cluster_ip_nodes = ""
 
+SUBNET_PREFIX          = "192.168."
```
gaurav-dalvi: I see similar code in this PR : #613
Are those related or just a duplicate code ?

erikh: correct, I just brought in the commit temporarily so we can review that separately, and I can use it here.

I'll amend this when it's merged.

erikh:
the ssh connection timed out; I'll retry.

does this happen a lot?

PR #612, opened 2016-11-08
mgmtfn/k8splugin: add network setup unit test

unclejack:
This PR adds a unit test for some of the network setup functions from driver.go.

This increases the coverage from 25% to 33.4%.

### DIFF
--- a/mgmtfn/k8splugin/driver_test.go
+++ b/mgmtfn/k8splugin/driver_test.go
@@ -1,0 +1,215 @@
+/***
+Copyright 2016 Cisco Systems Inc. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package k8splugin
+
+import (
+	"fmt"
+	"net"
+	"os/exec"
+	"runtime"
+
+	"github.com/vishvananda/netlink"
+	"github.com/vishvananda/netns"
+	. "gopkg.in/check.v1"
+)
+
+type NetSetup struct {
+	newNS    netns.NsHandle
+	globalNS netns.NsHandle
+	pid      int
+	cmd      *exec.Cmd
+	ifName   string
+	link     netlink.Link
+}
+
+var _ = Suite(&NetSetup{})
+
+func (s *NetSetup) SetUpTest(c *C) {
+	la := netlink.NewLinkAttrs()
+	s.ifName = "testlinkfoo"
+	la.Name = s.ifName
+	runtime.LockOSThread()
+	locked := true
+	defer func() {
+		if locked {
+			runtime.UnlockOSThread()
+		}
+	}()
+
+	globalNS, err := netns.Get()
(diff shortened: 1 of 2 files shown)


unclejack:
ping @abhinandanpb @erikh @jojimt @gaurav-dalvi

erikh:
Any chance I could convince you to use gocheck for this? We are using it in a lot of the newer test systems.

mgmtfn/k8splugin/driver_test.go
```diff
@@ -0,0 +1,189 @@
...
+		globalNS.Close()
+	}()
+
+	newNS, err := netns.New()
+	if err != nil {
+		t.Fatal("failed to create new network namespace")
+	}
+	defer newNS.Close()
+
+	cmd := exec.Command("sleep", "9999")
```
erikh: busybox sleep doesn't support it, but both centos and ubuntu support `sleep infinity` which is waht I'm guessing you really want here.

mgmtfn/k8splugin/driver_test.go
```diff
@@ -0,0 +1,189 @@
...
+	// ensure the interface's name has been changed
+	newLink, err := netlink.LinkByName(newName)
+	if err != nil {
+		t.Fatalf("failed to get interface by name: %v", err)
+	}
+	defer netlink.LinkDel(newLink)
+
+	// ensure that the interface's IP address has been set properly
+	addresses, err := netlink.AddrList(newLink, netlink.FAMILY_V4)
+	ifAddr := fmt.Sprintf("%v", addresses[0].IPNet)
```
erikh: couldn't these `%v` prints just be `.String` calls? I presume they implement `fmt.Stringer`

mgmtfn/k8splugin/driver_test.go
```diff
@@ -0,0 +1,189 @@
...
+	// ensure the default gateway route has been set properly
+	routes, err := netlink.RouteList(newLink, netlink.FAMILY_V4)
+	if err != nil {
+		t.Errorf("failed to fetch routes")
+	}
+
+	foundDefaultGW := false
+	foundStaticRoute := false
+	for _, route := range routes {
+		gw := fmt.Sprintf("%v", route.Gw)
```
erikh: same here

erikh:
another advantage of gocheck; if we used it the test fixtures we could do the setns setup/reap so they were re-usable by other tests.

unclejack:
@erikh: I'll make the changes as you've proposed. There's no need to convince me. The reason why I've written it like that was to make the changes as small as possible.

mgmtfn/k8splugin/driver_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+func TestNetSetup(t *testing.T) {
+	la := netlink.NewLinkAttrs()
+	la.Name = "testlinkfoo"
+	newName := "testlinknewname"
+	address := "192.168.68.68/24"
+	defGW := "192.168.68.1"
+	staticRoute := "192.168.32.0/24"
+
+	runtime.LockOSThread()
```
jojimt: @unclejack Just for my understanding, why do we need to lock the runtime here?

unclejack: @jojimt: This is needed to set up the new network namespace and run the process in it (lines 54 - 66).

jojimt: makes sense, thanks!

jojimt:
Will merge once system tests pass. Thanks for taking care of this.

erikh:
needs a kick, this was last committed when the CI was down.

unclejack:
I'll merge this now. Joji has said he'd merge once the system tests pass. I need it for my next PR.

PR #613, opened 2016-11-09
Vagrantfile,.gitignore: Use subnet assignment code from volplugin

erikh:

### DIFF
--- a/test/systemtests/init_test.go
+++ b/test/systemtests/init_test.go
@@ -3,23 +3,25 @@ package systemtests
 import (
 	"flag"
 	"fmt"
+	"os"
+	. "testing"
+
 	"github.com/Sirupsen/logrus"
 	"github.com/contiv/contivmodel/client"
 	"github.com/contiv/remotessh"
 	. "gopkg.in/check.v1"
-	"os"
-	. "testing"
 )
 
 type systemtestSuite struct {
-	vagrant   remotessh.Vagrant
-	baremetal remotessh.Baremetal
-	cli       *client.ContivClient
-	nodes     []*node
-	fwdMode   string
-	basicInfo BasicInfo
-	hostInfo  HostInfo
-	globInfo  GlobInfo
+	vagrant       remotessh.Vagrant
+	baremetal     remotessh.Baremetal
+	cli           *client.ContivClient
+	nodes         []*node
+	fwdMode       string
+	netmasterPort string
+	basicInfo     BasicInfo
+	hostInfo      HostInfo
+	globInfo      GlobInfo
 }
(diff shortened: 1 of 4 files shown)


dseevr:
LGTM lol 👍

erikh:
oh this also gates the port forwarding of 9999 and 80 with an env var called `CONTIV_NOFORWARD`. 

This is all so you can multiple netplugin envs on the same host.

erikh:
I refactored the subnet code to do port assignment too. This way your port will not necessarily be 9999 from the host, but the systemtests will still run appropriately. It also means you will need to ssh into your instances (or provide commandline arguments) to use netctl manually.

erikh:
cc @jojimt on the last requirement, lmk if that's ok or not. This enables isolated, multiple builds per host for developers and CI alike.

abhinandanpb:
@erikh dont we have to create separate vbox bridges for each "vagrant setup"

test/systemtests/init_test.go
```diff
@@ -3,23 +3,25 @@ package systemtests
...
-	fwdMode   string
-	basicInfo BasicInfo
-	hostInfo  HostInfo
-	globInfo  GlobInfo
+	vagrant       remotessh.Vagrant
+	baremetal     remotessh.Baremetal
+	cli           *client.ContivClient
+	nodes         []*node
+	fwdMode       string
+	netmasterPort string
```
gaurav-dalvi: Should we put this in GlobalInfo Struct ?  This structure is for over all systemtest. Same comment for fwd mode field as well.

erikh: the above struct is the receiver for each test function.

erikh:
@abhinandanpb nto sure what you mean, this  works.

abhinandanpb:
@erikh we cant all vagrant setups in the same vbox bridge right. each setups should be connected to different vbox bridges. how are we going to isolate the traffic between setups ?

erikh:
this is creating an isolated network for each network we specify; whether or not a bridge is created I don't know if it matters?

we've used this heavily in volplugin and it's working for me here. tests should pass without issue. If there is a problem, can we explore it then?

abhinandanpb:
@erikh there is a problem. I can explain. Its a one liner change. no big deal.

erikh:
ok lmk what I need to do

erikh:
Does anyone know why this one is still failing? I can't determine from the logs, other than it looks like etcd is down.

erikh:
closing this as it depends on a few things in the #610 branch it seems, so we need to move from that PR. it will contain this code.

PR #614, opened 2016-11-10
test/systemtests: Paralellize network_test.go

erikh:
**this test currently fails**

It is here to record the failure so we can address it later.

### DIFF
--- a/test/systemtests/network_test.go
+++ b/test/systemtests/network_test.go
@@ -11,28 +11,35 @@ import (
 	. "gopkg.in/check.v1"
 )
 
-func (s *systemtestSuite) TestInfraNetworkAddDeleteVXLAN(c *C) {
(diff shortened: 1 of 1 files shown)


test/systemtests/network_test.go
```diff
@@ -11,51 +11,52 @@ import (
...
+	for i := 0; i < 1; i++ {
 		var (
 			netNames = []string{}
 		)
 
 		numInfraNw := 3
 		for networkNum := 0; networkNum < numInfraNw; networkNum++ {
 			network := &client.Network{
-				TenantName:  "default",
+				TenantName:  tenantName,
```
gaurav-dalvi: we need specific tests to be run under default tenant as well. We create that tenant as part of installation and We need to make sure that default tenant is also working. Thats the reason default was hard coded here.

erikh: this test is broken because we *weren't* testing this.

Issue #611, opened 2016-11-08
Continuous control interface flap causing loss of network info


abhinandanpb:
Found in system tests. Network state in docker seems to be temporarily deleted when control interface is flapped continuosly(expected). System tests need deterministic wait to be added instead of random sleep interval.

PR #618, opened 2016-11-12
Fixing TriggerNetpartition test error

abhinandanpb:
RC: Test failed because during etcd flaps we were trying to create containers when etcd was still not healthy and docker was reconnecting to etcd.

### DIFF
--- a/test/systemtests/docker_test.go
+++ b/test/systemtests/docker_test.go
@@ -603,6 +603,38 @@ func (d *docker) checkSchedulerNetworkCreated(nwName string, expectedOp bool) er
 	return err
 }
 
+func (d *docker) checkSchedulerNetworkOnNodeCreated(nwNames []string, n *node) error {
+	ch := make(chan error, 1)
+	for _, nwName := range nwNames {
+		go func(nwName string, n *node, ch chan error) {
+			logrus.Infof("Checking whether docker network %s is created on node %s", nwName, n.Name())
+			cmd := fmt.Sprintf("docker network ls | grep netplugin | grep %s | awk \"{print \\$2}\"", nwName)
+			logrus.Infof("Command to be executed is = %s", cmd)
+			count := 0
+			//check if docker network is created for a minute
+			for count < 60 {
+				op, err := n.runCommand(cmd)
+
+				if err == nil {
+					ret := strings.Contains(op, nwName)
+					if ret == true {
+						ch <- nil
+					}
+					count++
+					time.Sleep(1 * time.Second)
+				}
+			}
+			ch <- fmt.Errorf("Docker Network %s not created on node %s \n", nwName, n.Name())
+		}(nwName, n, ch)
+	}
+	for range nwNames {
+		if err := <-ch; err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 func (d *docker) waitForListeners() error {
 	return d.node.runCommandWithTimeOut("netstat -tlpn | grep 9090 | grep LISTEN", 500*time.Millisecond, 50*time.Second)
 }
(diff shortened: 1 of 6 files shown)


abhinandanpb:
Fixes #611

test/systemtests/docker_test.go
```diff
@@ -603,6 +603,38 @@ func (d *docker) checkSchedulerNetworkCreated(nwName string, expectedOp bool) er
 	return err
 }
 
+func (d *docker) checkSchedulerNetworkOnNodeCreated(nwNames []string, n *node) error {
+	ch := make(chan error, 1)
+	for _, nwName := range nwNames {
+		go func(nwName string, n *node, ch chan error) {
+			logrus.Infof("Checking whether docker network %s is created on node %s", nwName, n.Name())
+			cmd := fmt.Sprintf("docker network ls | grep netplugin | grep %s | awk \"{print \\$2}\"", nwName)
```
jojimt: Should we just do etcdctl ls /contiv.io? That way, it will work for k8s as well...

abhinandanpb: k8s has different interface implementation.

erikh:
general comment: use a `for` with a counter and `len()` instead of iterating over the loop in your waiters impls

abhinandanpb:
sure. any reason for the len to be different ?

erikh:
it's just a nit. feel free to ignore.

test/systemtests/docker_test.go
```diff
@@ -603,6 +603,38 @@ func (d *docker) checkSchedulerNetworkCreated(nwName string, expectedOp bool) er
 	return err
 }
 
+func (d *docker) checkSchedulerNetworkOnNodeCreated(nwNames []string, n *node) error {
+	ch := make(chan error, 1)
```
erikh: Actually that may not happen. I would have to test to be sure. Let's leave it alone instead of sending that through CI again.

test/systemtests/swarm_test.go
```diff
@@ -565,6 +565,38 @@ func (w *swarm) checkSchedulerNetworkCreated(nwName string, expectedOp bool) err
...
+	ch := make(chan error, 1)
+	for _, nwName := range nwNames {
+		go func(nwName string, n *node, ch chan error) {
+			logrus.Infof("Checking whether docker network %s is created on node %s", nwName, n.Name())
+			cmd := fmt.Sprintf("docker network ls | grep netplugin | grep %s | awk \"{print \\$2}\"", nwName)
+			logrus.Infof("Command to be executed is = %s", cmd)
+			count := 0
+			//check if docker network is created for a minute
+			for count < 60 {
+				op, err := n.runCommand(cmd)
```
erikh: what happens when `err != nil` here?

test/systemtests/docker_test.go
```diff
@@ -603,6 +603,38 @@ func (d *docker) checkSchedulerNetworkCreated(nwName string, expectedOp bool) er
...
+	for _, nwName := range nwNames {
+		go func(nwName string, n *node, ch chan error) {
+			logrus.Infof("Checking whether docker network %s is created on node %s", nwName, n.Name())
+			cmd := fmt.Sprintf("docker network ls | grep netplugin | grep %s | awk \"{print \\$2}\"", nwName)
+			logrus.Infof("Command to be executed is = %s", cmd)
+			count := 0
+			//check if docker network is created for a minute
+			for count < 60 {
+				op, err := n.runCommand(cmd)
+
```
erikh: same here

test/systemtests/swarm_test.go
erikh: nvm...

erikh:
well I just made a bunch of noise, sorry.

LGTM

PR #619, opened 2016-11-14
use fmt.Print when there's only one argument

unclejack:


netctl/netctl.go
```diff
@@ -934,11 +934,9 @@ func showVersion(ctx *cli.Context) {
 	if err := getObject(ctx, versionURL(ctx), &ver); err != nil {
 		fmt.Printf("Unable to fetch version information\n")
```
gaurav-dalvi: Why not change here then ?

gaurav-dalvi:
If we are doing this change, we need to make sure we do at all over places in this repo.
Did you look for all fmt.Printf occurrences ?

unclejack:
@gaurav-dalvi It's done now.

gaurav-dalvi:
LGTM !

rchirakk:
whats the significance of this change ?

unclejack:
@rchirakk fmt.Print should be used when there's only one argument. This is related to having cleaner code.

jojimt:
@unclejack Is this based on any go guideline? Normally, the better practice is to use a function that is independent of the number of arguments.

unclejack:
@jojimt `fmt.Print("foo")` is to be used when we don't need to format anything and just print without EOL. `fmt.Println("foo")` would be used when we also want to print EOL. `fmt.printf("foo: %v %v", a, b)` would be the right thing to use when we print with formatting.

If I always use `fmt.Printf`, it's harder for someone to find mistakes when they take a look at the uses of `fmt.Printf` with a single argument.

gaurav-dalvi:
I prefer using fmt.Printf() everywhere to have consistency in the code. (even if with arguments or no arguments).

jojimt:
I agree with @gaurav-dalvi. Printf is a standard function. Let's not make this change. I might change my opinion if you can point me to a go guideline or govet/glint issue.

PR #623, opened 2016-11-15
netplugin/svcplugin: check etcd connection errors.

erikh:
The error is unchecked here which causes the daemon to panic in a quite surprising (and hard to debug) way. I modified error checking up the stack so it actually gets somewhere, but I'm not sure if it's actually going to reach an abort/handling point, can someone verify this?


erikh:
@jojimt ready for departure

PR #625, opened 2016-11-15
Update DevEnv.md

dojacobs:
Added pre-req section for newbies like me.


erikh:
I dunno if this really needs to pass tests to be merged. Just wanted to make a note of that if we're looking to beat down the CI queue.

unclejack:
LGTM

PR #626, opened 2016-11-16
Added arp flood mode to disable proxy-arp

gkvijay:


drivers/ovsdriver.go
```diff
@@ -809,3 +809,29 @@ func (d *OvsDriver) InspectBgp() ([]byte, error) {
...
+
+// GlobalConfigUpdate sets the global level configs like arp-mode
+func (d *OvsDriver) GlobalConfigUpdate(inst core.InstanceInfo) error {
+	// convert the netplugin config to ofnet config
+	// currently, its only ArpMode
+	var cfg ofnet.OfnetGlobalConfig
+	switch inst.ArpMode {
+	case "flood":
+		cfg.ArpMode = ofnet.ArpFlood
+	case "proxy":
```
jojimt: Just to cover upgrade scenario, lets treat "" as proxy. Thanks.

jojimt:
LGTM other than the minor comment

drivers/ovsdriver.go
gkvijay: made the default as proxy.

PR #627, opened 2016-11-16
Added EPG resource

DivyaVavili:
Create EPG resource to allocate EPG IDs. This replaces the global EPG state that was maintained.

### DIFF
--- a/netmaster/gstate/gstate.go
+++ b/netmaster/gstate/gstate.go
@@ -18,6 +18,7 @@ package gstate
 import (
 	"encoding/json"
 	"errors"
+	"math"
 	"strconv"
 	"strings"
 	"sync"
@@ -73,16 +74,15 @@ func (gc *Cfg) Dump() error {
 
 func (gc *Cfg) checkErrors(res string) error {
 	var err error
-	if res == "vlan" {
+	switch res {
+	case "vlan":
 		_, err = netutils.ParseTagRanges(gc.Auto.VLANs, "vlan")
-		if err != nil {
-			return err
-		}
-	} else if res == "vxlan" {
+	case "vxlan":
 		_, err = netutils.ParseTagRanges(gc.Auto.VXLANs, "vxlan")
-		if err != nil {
-			return err
-		}
+	case "epg":
+		return nil
+	default:
+		err = core.Errorf("Undefined resource")
 	}
 	return err
 }
@@ -102,6 +102,11 @@ func Parse(configBytes []byte) (*Cfg, error) {
 	}
 
 	err = gc.checkErrors("vxlan")
+	if err != nil {
+		return nil, err
+	}
+
+	err = gc.checkErrors("epg")
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


netmaster/gstate/gstate.go
```diff
@@ -438,7 +518,18 @@ func (gc *Cfg) UpdateResources(res string) error {
 			log.Errorf("error '%s' updating global oper state %v \n", err, g)
 			return err
 		}
+	} else if res == "epg" {
```
abhinandanpb: This will only be needed if we are updating the resource ranges. Since we will not be doing for epg. should this be ignored ?

netmaster/master/endpointGroup.go
```diff
@@ -73,28 +66,20 @@ func CreateEndpointGroup(tenantName, networkName, groupName string) error {
...
 	}
+	log.Debugf("Allocated EPG ID: %d for EPG: %s", epgID, groupName)
 
 	// Create epGroup state
 	epgCfg := &mastercfg.EndpointGroupState{
 		GroupName:       groupName,
 		TenantName:      tenantName,
 		NetworkName:     networkName,
-		EndpointGroupID: epgID,
+		EndpointGroupID: int(epgID),
```
abhinandanpb: we can make EndpointGroupID as uint itself ?

netmaster/resources/epgresource.go
```diff
@@ -0,0 +1,298 @@
...
+		if err != nil {
+			// continue cleanup
+		}
+	}
+
+	r.Clear()
+}
+
+// Reinit the Resource. Requires a *bitset.BitSet.
+func (r *AutoEPGCfgResource) Reinit(rsrcCfg interface{}) error {
```
abhinandanpb: again implementation may not be needed if we are not letting the range configurable

netmaster/resources/epgresource.go
```diff
@@ -0,0 +1,298 @@
...
+func (r *AutoEPGCfgResource) Allocate(reqVal interface{}) (interface{}, error) {
+	oper := &AutoEPGOperResource{}
+	oper.StateDriver = r.StateDriver
+	err := oper.Read(r.ID)
+	if err != nil {
+		return nil, err
+	}
+
+	var epgID uint
+	if (reqVal != nil) && (reqVal.(uint) != 0) {
```
abhinandanpb: since this resource allocation is only for id. we do not expect to have a specific id requested. so this may not be needed. We could straight away allocated the next free id in the bitset.

netmaster/resources/stateresourcemanager.go
```diff
@@ -76,8 +77,9 @@ func (rm *StateResourceManager) Init() error { return nil }
 // Deinit cleans up the resource manager
 func (rm *StateResourceManager) Deinit() {}
 
+// FindResource returns the requested resource
 // XXX: It might be better to keep cache of resources and avoid frequent etcd reads
-func (rm *StateResourceManager) findResource(id, desc string) (core.Resource, bool, error) {
+func (rm *StateResourceManager) FindResource(id, desc string) (core.Resource, bool, error) {
```
abhinandanpb: any reason to export it ?

abhinandanpb:
LGTM other than minor comments.

DivyaVavili:
Closing this. Working on a separate ID resource to handle this properly. Will raise a separate PR

Issue #603, opened 2016-11-03
use official docker client in integration tests


rchirakk:
integration tests use 'samalba' docker client lib. which doesn't work for docker 1.12

erikh:
\o/

PR #629, opened 2016-11-16
use official docker client in mesos tests.

rchirakk:
fixes https://github.com/contiv/netplugin/issues/603
enable mesos integration tests

### DIFF
--- a/test/integration/mesos_test.go
+++ b/test/integration/mesos_test.go
@@ -25,19 +25,23 @@ import (
 	"os/exec"
 	"strconv"
 	"strings"
-	"time"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/contiv/contivmodel/client"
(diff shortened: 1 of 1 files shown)


gaurav-dalvi:
Do we need to use official docker client at other places in the code as well ?

rchirakk:
only if it fails with the latest version of docker.

gaurav-dalvi:
I think we should replace with official docker client everywhere in our code.

rchirakk:
these tests require the latest client. it may not be necessary in other places.
yea, its good to move to the official client.

gaurav-dalvi:
@rchirakk : Could you please take care of adding official docker client to all other places in the code ?

I am ok with other PR as well. for this PR : LGTM.

jojimt:
LGTM

PR #639, opened 2016-11-28
Fix intermittent test failure

jojimt:
Fix the following failure:
```time="2016-11-27T02:58:24-08:00" level=info msg="cmd \"docker run -itd --name=private-srv0-0-0 --net=private   contiv/alpine sleep 600m\" failed: output below" 
time="2016-11-27T02:58:24-08:00" level=info msg="4cf299b4ec9e7e27a4c76f9c6fec79488a68006f69d8521240021ed93a3b2167\ndocker: Error response from daemon: service endpoint with name private-srv0-0-0 already exists.\n" ```

### DIFF
--- a/test/systemtests/trigger_test.go
+++ b/test/systemtests/trigger_test.go
@@ -276,10 +276,10 @@ func (s *systemtestSuite) TestTriggerNetPartition(c *C) {
 
 			// flap the control interface
 			c.Assert(node.bringDownIf("eth1"), IsNil)
-			time.Sleep(50 * time.Second) // wait till sessions/locks timeout
+			time.Sleep(25 * time.Second) // wait till sessions/locks timeout
 			c.Assert(node.bringUpIf("eth1", nodeIP), IsNil)
 
-			time.Sleep(20 * time.Second)
+			time.Sleep(23 * time.Second)
 			c.Assert(s.verifyVTEPs(), IsNil)
 
 			c.Assert(s.verifyEPs(containers), IsNil)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: test/systemtests/trigger_test.go [hunk 1]
```diff
@@ -276,10 +276,10 @@ func (s *systemtestSuite) TestTriggerNetPartition(c *C) {
 
 			// flap the control interface
 			c.Assert(node.bringDownIf("eth1"), IsNil)
-			time.Sleep(50 * time.Second) // wait till sessions/locks timeout
+			time.Sleep(25 * time.Second) // wait till sessions/locks timeout
 			c.Assert(node.bringUpIf("eth1", nodeIP), IsNil)
 
-			time.Sleep(20 * time.Second)
+			time.Sleep(23 * time.Second)
```
gaurav-dalvi: just curious why previous sleep of 20 seconds was not working and 23 seconds of sleep time is working now ?

jojimt: This is just tuning the etcd cluster bring up time.

erikh: could we poll etcd with a timeout instead? it'll be more reliable than a straight sleep and should exit prematurely if the etcd service comes back sooner.

jojimt: @erikh we do have the polling in place. All I did was adjust the base delay. I added 3 sec here and removed 25 sec above this.

erikh: ah I missed that, thanks.

PR #640, opened 2016-11-29
Port bonding changes

DivyaVavili:
Includes changes to accept multiple interfaces for uplinks and create Portbond with LACP enabled


test/systemtests/bgp_test.go
```diff
@@ -636,7 +636,7 @@ func (s *systemtestSuite) TestBgpMultiTrigger(c *C) {
 				s.CheckBgpConnectionForaNode(c, node.tbnode)
 			}
 		}
-		c.Assert(nodeToStop.startNetplugin("-vlan-if=eth2"), IsNil)
+		c.Assert(nodeToStop.startNetplugin(""), IsNil)
```
jojimt: Why are we changing this?

netplugin/netd.go
```diff
@@ -43,9 +67,9 @@ type cliOpts struct {
 	debug      bool
 	syslog     string
 	jsonLog    bool
-	ctrlIP     string // IP address to be used by control protocols
-	vtepIP     string // IP address to be used by the VTEP
-	vlanIntf   string // Uplink interface for VLAN switching
+	ctrlIP     string      // IP address to be used by control protocols
+	vtepIP     string      // IP address to be used by the VTEP
+	vlanIntf   StringSlice // Uplink interface for VLAN switching
```
jojimt: Can we take the vlan-if argument as a comma separated list?

Vagrantfile
```diff
@@ -276,7 +276,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
             end
 
             node.vm.provision "shell" do |s|
-                s.inline = "echo '#{node_ips[0]} netmaster' >> /etc/hosts; echo '#{node_ips[1]} netmaster' >> /etc/hosts;	echo '#{node_addr} #{node_name}' >> /etc/hosts"
+                s.inline = "echo '#{node_ips[0]} netmaster' >> /etc/hosts; echo '#{node_ips[1]} netmaster' >> /etc/hosts;   echo '#{node_addr} #{node_name}' >> /etc/hosts"
```
jojimt: Looks like a whitespace change? Can we eliminate all such changes?

Vagrantfile
```diff
@@ -156,6 +157,7 @@ end
 
 VAGRANTFILE_API_VERSION = "2"
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
+    config.vm.box_check_update = false
```
DivyaVavili: Had this to speed up testing on my local setup. I could remove it, but I think it's better to have this if we have the constraint of checking the box versions later anyway. Would we miss any box upgrade scenario?

Vagrantfile
```diff
@@ -249,21 +251,19 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
             # create an interface for etcd cluster
             node.vm.network :private_network, ip: node_addr, virtualbox__intnet: "true", auto_config: false
             # create an interface for bridged network
-            if ENV['CONTIV_L3'] then
-              # create an interface for bridged network
-              node.vm.network :private_network, ip: "0.0.0.0", virtualbox__intnet: network_name, auto_config: false
-            else
-              node.vm.network :private_network, ip: "0.0.0.0", virtualbox__intnet: "true", auto_config: false
-            end
```
DivyaVavili: This setting is required to set each each interface in a separate network. Had a discussion with @shaleman and @abhinandanpb regarding this as well; this could be the default and not be based on CONTIV_L3 check.

netplugin/netd.go
DivyaVavili: Yes, I've made changes to StringSlice Set function to handle comma separated list as well

test/systemtests/bgp_test.go
DivyaVavili: The test suite already brings up the node with the necessary data interface configured in the test config file. There is no need to pass an vlan-if argument here. Verified that this is redundant. We shouldn't be doing the hard-coding here and the test config file handles this scenario.

Vagrantfile
jojimt: @erikh pls comment.

drivers/ovsSwitch.go
```diff
@@ -562,24 +570,34 @@ func (sw *OvsSwitch) DeleteVtep(vtepIP string) error {
...
+			}
+			uplinkType = ofnet.BondType
+		} else {
+			// Ask OVSDB driver to add the port as a trunk port
+			err = sw.ovsdbDriver.CreatePort(intfList[0], "", uplinkName, 0, 0, 0)
+			if err != nil {
+				log.Errorf("Error adding uplink %s to OVS. Err: %v", intfList[0], err)
+				return err
+			}
+			uplinkType = ofnet.PortType
```
jojimt: OK. Pls open an issue to track.

Vagrantfile
erikh: Hmm. I don't think this is an issue, but can this be done in its own commit so we can revert it if it becomes a problem?

jojimt: That sounds good. @DivyaVavili let's do that instead of combining with this commit.

DivyaVavili: Sure. Will do.

PR #642, opened 2016-11-30
Makefile: don't call start three times in all-CI

unclejack:


gaurav-dalvi:
Whats the purpose of this change ? 
I did not understand. It just changes name of some targets.

unclejack:
@gaurav-dalvi `all-CI` would end up calling start three times. The reason was that ssh-build and system-test were calling it as well. This change adds targets for ssh-build and system-test without start. This change helps speed up the provisioning for the tests.

erikh:
I think the right approach is to not call *stop* at hte right times instead of not calling start. vagrant will happily move forward and exit 0 if the vms are already started, but do something destructive if they are stopped.

unclejack:
@erikh: Some of the provisioning steps are run anyway.

gaurav-dalvi:
LGTM.

erikh:
can we stop approving changes that don't even pass tests yet?

unclejack:
The CI build is green now. PTAL

gaurav-dalvi:
Sure @erikh . Will take of that from next time.

@unclejack : Can you please take a look at here https://github.com/contiv/netplugin/blob/master/docs/DevEnv.md

with this change do we need to change this documentation ?

unclejack:
@gaurav-dalvi The only update to make to that document is a change to step 5. We can optimize it so that we use `system-test-nostart` there.

I'd rather make that change after we merge this PR so the CI benefits from these changes.

gaurav-dalvi:
Thanks @unclejack . Please make sure when this PR is merged you change that file as well.

erikh:
did you consider just making the build task not stop the vms?

That would literally be the only change you needed to make, I think, and reduces the complexity we're introducing here.

unclejack:
I'll close this now in favor of a rewrite of the Makefile.

PR #415, opened 2016-06-17
resolve http/https proxy conflict between host and vm

gwanii:
I'd started VM demo cluster for testing with proxy settings. 

To get free from the GFW(mainly for download vagrant box from atlas.hashicorp.com),  set http proxy on my host: `export http_proxy=http://127.0.0.1:8787 && export https_proxy=$http_proxy`.

Unfortunately, my vm cant visit golang.org, then also set http proxy in Makefile(line78): `http_proxy=http://10.0.2.2:8787 https_proxy=http://10.0.2.2:8787 vagrant up`. VMs use NAT network, so http proxy to virtual router(with IP 10.0.2.2). Then, run `make demo` and host proxy confict with vm proxy settings.

PR #647, opened 2016-12-05
Vagrantfile: make {http,https}_proxy all caps variables

unclejack:
This PR is a replacement for #415. An environment where the host and the VMs require different proxies needs this. #415 would have required a rebase.

This also changes the double quotes to single quotes for these variables.

Replaces #415 
Closes #415


erikh:
These are both valid sets of values and the CI uses the one that already exists. We need to support both if we're going to change this.

gaurav-dalvi:
LGTM

PR #648, opened 2016-12-06
scripts/unittests: add per test timeouts

unclejack:
This PR adds timeouts to the unit tests.

### DIFF
--- a/scripts/unittests
+++ b/scripts/unittests
@@ -13,6 +13,7 @@ cp -rf $GOSRC/github.com/contiv/netplugin/mgmtfn/k8splugin/certs /tmp
 run_in_vagrant=false
 coverage_basic=false
 coverage_detail=false
+timeout=${TIMEOUT:=10m}
 test_packages="github.com/contiv/netplugin/drivers \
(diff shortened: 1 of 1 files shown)


dseevr:
Thanks for doing this 👍

gaurav-dalvi:
LGTM.

erikh:
10m is the default go test timeout. This doesn't really solve much unless I'm missing something?

unclejack:
@erikh This is what I've found in the source code:
```
timeout          = flag.Duration("test.timeout", 0, "if positive, sets an aggregate time limit for all tests")
```

erikh:
```
        -timeout t
            If a test runs longer than t, panic.
            The default is 10 minutes (10m).
```

erikh:
I strong suggest making it variable so we can drive it through the environment (and adjust it at the CI level).

unclejack:
@erikh I've updated it to make it specify 10m by default and override it when the env var is set. A default of ten minutes makes sense now. We don't need special cases to not specify the timeout and it just falls back to the default. PTAL

PR #649, opened 2016-12-06
return directly instead of using if where possible

unclejack:


dseevr:
Nice finds.  These could be simplified to just `return`ing the function call directly, no?

unclejack:
You're right, but hindsight is always 20/20, heh. PTAL

PR #650, opened 2016-12-06
Copy netplugin, netmaster and ovs logs after systemtest is done.

gkvijay:
Increased the retry count for checking the port state in OVS.

### DIFF
--- a/drivers/ovsdbDriver.go
+++ b/drivers/ovsdbDriver.go
@@ -29,7 +29,7 @@ import (
 )
 
 // Max number of retries to get ofp port number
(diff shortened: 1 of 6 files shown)


gaurav-dalvi:
How will this work in case of netplugin and netmaster running as a container ?

erikh:
Please make this a shell script. I reviewed the original changes and they're quite a bit of effort for 1-2 scp calls in the makefile. this does not belong as a part of our suite.

gkvijay:
Closing this as we need a log collection mechanism outside of systemtests

PR #653, opened 2016-12-09
set vagrant box update to false

DivyaVavili:
Adding vagrant box_check_update to false. Since we are checking for box version constraint, we can leverage this check to fasten box bringup.


DivyaVavili:
@erikh @jojimt FYI

unclejack:
LGTM

PR #655, opened 2016-12-09
Vagrantfile: enforcement of minimum Vagrant version (currently 1.7.3)

### LINKED ISSUE #654: restrict vagrant up to only if vagrant version is 1.8.6+
for all vagant files here : https://github.com/contiv/netplugin/tree/master/vagrant

dseevr:

Fixes #654

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,6 +1,8 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
+require File.join(File.dirname(__FILE__), "vagrant", "vagrant_version_check")
+
 require 'fileutils'
 
 BEGIN {
--- a/vagrant/k8s/Vagrantfile
+++ b/vagrant/k8s/Vagrantfile
@@ -1,5 +1,8 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
+
+require File.join(File.dirname(__FILE__), "..", "vagrant_version_check")
+
 require "rubygems"
 require "json"
 require 'fileutils'
--- a/vagrant/mesos-cni/Vagrantfile
+++ b/vagrant/mesos-cni/Vagrantfile
@@ -1,6 +1,8 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
+require File.join(File.dirname(__FILE__), "..", "vagrant_version_check")
+
 require 'fileutils'
 
 # netplugin_synced_gopath="/opt/golang"
--- a/vagrant/mesos-docker/Vagrantfile
+++ b/vagrant/mesos-docker/Vagrantfile
@@ -1,6 +1,8 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
+require File.join(File.dirname(__FILE__), "..", "vagrant_version_check")
+
 require 'fileutils'
 
 # netplugin_synced_gopath="/opt/golang"
--- a/vagrant/nomad-docker/Vagrantfile
+++ b/vagrant/nomad-docker/Vagrantfile
@@ -1,6 +1,8 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
+require File.join(File.dirname(__FILE__), "..", "vagrant_version_check")
+
 require 'fileutils'
 
 # netplugin_synced_gopath="/opt/golang"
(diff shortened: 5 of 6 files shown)


gaurav-dalvi:
Thanks for taking care of this lightening fast. I hope you have tested this.
Also could you please make sure this is applicable for all the vagrant files under ? 
https://github.com/contiv/netplugin/tree/master/vagrant

dseevr:
You can easily test it by changing the `MINIMUM_VAGRANT_VERSION` to `1.8.8` or something higher than what you have installed.

Do we want to add this to all of those files as well?

gaurav-dalvi:
Yes let us add for all vagrantfiles in our repo.

dseevr:
@gaurav-dalvi  As mentioned on Slack, it looks like CI itself verified that this patch works.  😄 

I put the version check into a single file called `vagrant/vagrant_version_check.rb` and it's included at the top of every other `Vagrantfile`.

You can manually test it by checking out this branch, editing the `MINIMUM_VAGRANT_VERSION` in `vagrant_version_check.rb` to something higher and then trying a `vagrant up` in each dir.

dseevr:
@gaurav-dalvi if you would prefer getting this merged in sooner rather than later, I can change the minimum version to `1.8.5`.  This will let it work with the current CI setup and we already have code in place to disable the functionality which triggered the showstopper bug in `1.8.5` (the `insert_key = false` stuff).  `1.8.5` was the first version version to support VirtualBox 5.1, too.

erikh:
https://github.com/kubernetes/kubernetes/commit/da3a2d78e75eabfedd95560b99843c934089fdf9

dseevr:
@erikh yep, I remember that 😄 we already have it disabled regardless of version: https://github.com/contiv/netplugin/blob/847832e7a498ded9e5b8b2d50aff1d1bcf88a50a/Vagrantfile#L183

dseevr:
Per offline discussion, the minimum version target has been changed to 1.8.5 as its bugs don't affect us and we won't have to make any changes to CI.

gaurav-dalvi:
LGTM

unclejack:
LGTM

jojimt:
@dseevr:  What is the reason for this change? I have been using vagrant 1.7.3 with netplugin all along without any problem.

dseevr:
@jojimt 1.7.3 doesn't support vbox 5.1

I believe @gaurav-dalvi requested this to reduce the number of people hitting problems because of running outdated/incompatible versions of Vagrant

jojimt:
Most of our dev machines have vagrant 1.7.3 and vbox 5.0. I would like to understand exactly what problem we are solving here.

dseevr:
@jojimt I have no objection to lowering it to 1.7.3

cc @gaurav-dalvi

dseevr:
@unclejack the problem is not 1.8.5 (we disable the insertion of the "insecure" key on all versions so that bug doesn't affect us at all), it's that we have people coming into the Slack running ancient versions of Vagrant and complaining that netplugin doesn't work.

This is a simple check that will improve the end user experience by giving them an immediate step they can take to resolve their problem rather than them having to seek out someone on the project team (usually Gaurav lol) only to find out the ancient software they're running is the problem

gaurav-dalvi:
yes I requested that change. I have seen people struggling to use our software. Enforcing this change would give them notice much before if they are using wrong versions of vagrant.

jojimt:
I am ok with adding a check, as long as we are not forcing people to upgrade for no reason.

dseevr:
@gaurav-dalvi Joji doesn't want to force people to upgrade to 1.8.5.  If they try to run vbox 5.1 with an older version of Vagrant that doesn't support it, they'll have to resolve that themselves.

@jojimt I just updated the PR to enforce 1.7.3 as a minimum, PTAL (CI is failing for a completely unrelated reason)

jojimt:
ok, pls retrigger sanity and i'll merge once it goes through.

dseevr:
@jojimt CI finally passed here

xiaodan1:
I have a based problem, how to deploy on two host?

unclejack:
@xiaodan1: Please open a new issue using https://github.com/contiv/netplugin/issues/new.

xiaodan1:
Thanks, I will to look at.

PR #657, opened 2016-12-12
embedded  nameserver for tenant based lookups

rchirakk:
embedded DNS for netplugin


netplugin/nameserver/nameserver.go
```diff
@@ -0,0 +1,673 @@
...
+	}
+}
+
+func (ens *NetpluginNameServer) processStateEvent() {
+	for {
+		select {
+		case <-ens.svcErrChan:
+			dnsLog.Warnf("nameserver restarted service watcher")
+			ens.incVrfErrStats("", "svcWatchRestart")
+			ens.startSvcWatch()
```
jojimt: Let's inert a delay before we restart the watch.

netplugin/nameserver/nameserver.go
```diff
@@ -0,0 +1,673 @@
...
+				ens.delService(svcState.Prev)
+			} else {
+				// add again
+				ens.addService(svcState.Curr)
+			}
+
+		case <-ens.epErrChan:
+			dnsLog.Warnf("nameserver restarted endpoint watcher")
+			ens.incVrfErrStats("", "epWatchRestart")
+			ens.startEndpointWatch()
```
jojimt: Here as well.

netplugin/nameserver/nameserver.go
```diff
@@ -0,0 +1,673 @@
...
+		sync.RWMutex
+		tenantStats map[string]map[string]uint64
+	}
+}
+
+type nameRecord struct {
+	v4Record net.IP
+	v6Record net.IP
+}
+
```
rchirakk: done

rchirakk:
@jojimt PTAL

netplugin/nameserver/nameserver.go
```diff
@@ -0,0 +1,705 @@
...
+	return rr, len(rr)
+}
+
+func (dt *dnsTables) lookUpEndPointV4Record(nameList map[string]bool, name string) ([]dns.RR, int) {
+	rr := []dns.RR{}
+	for i := range nameList {
+		if record, ok := dt.endpointTbl[i]; ok {
+			if record.v4Record.IsUnspecified() != true {
+				r := func() dns.RR {
```
DivyaVavili: Why do we require a function here? Shouldn't simple initialization of type dns.A suffice here?

netplugin/nameserver/nameserver.go
```diff
@@ -0,0 +1,705 @@
...
+	return rr, len(rr)
+}
+
+func (dt *dnsTables) lookUpEndPointV6Record(nameList map[string]bool, name string) ([]dns.RR, int) {
+	rr := []dns.RR{}
+	for i := range nameList {
+		if record, ok := dt.endpointTbl[i]; ok {
+			if record.v6Record.IsUnspecified() != true {
+				r := func() dns.RR {
+					r := new(dns.AAAA)
```
DivyaVavili: Same as above. Simple initialization should suffice.

netplugin/nameserver/nameserver.go
```diff
@@ -0,0 +1,705 @@
...
+
+	dnsLog.Infof("[tenant: %s]add service name:%s, ipadddr: %s",
+		svc.Tenant, svc.ServiceName, svc.IPAddress)
+	tenant := svc.Tenant
+	tenMap := ens.getBucket(tenant)
+	tenMap.RWMutex.Lock()
+	defer tenMap.RWMutex.Unlock()
+	tenantTables, ok := tenMap.tenantTables[tenant]
+	if !ok {
+		tenantTables = new(dnsTables)
```
DivyaVavili: Can we a function to initialize all tenant tables? The initialization seems to be done at multiple places.

netplugin/nameserver/nameserver.go
```diff
@@ -0,0 +1,705 @@
...
+
+func lookUpServiceV4Record(record nameRecord, name string) ([]dns.RR, int) {
+	rr := []dns.RR{}
+
+	if record.v4Record.IsUnspecified() != true {
+		r := func() dns.RR {
+			r := new(dns.A)
+			r.A = record.v4Record
+			r.Hdr = dns.RR_Header{Name: name + ".", Rrtype: dns.TypeA,
+				Class: dns.ClassINET, Ttl: 64}
```
DivyaVavili: The original recommendation is to keep the TTL to 1 day. But considering that containers are going to be short lived, we could probably choose a small meaningful value like 5 min/300 sec. @jojimt @gkvijay Please comment as well.

netplugin/nameserver/nameserver.go
```diff
@@ -0,0 +1,705 @@
...
+
+type tenantBucket struct {
+	sync.RWMutex
+	tenantTables map[string]*dnsTables
+}
+
+const commonK8sTenant = "common"
+
+// NetpluginNameServer config
+type NetpluginNameServer struct {
```
rchirakk: not really, we can hook any nameserver in netplugin, and this file implements netplugin nameserver,
hence the name

netplugin/nameserver/nameserver.go
rchirakk: No, then a type has to be changed, which happens here in the closure

netplugin/nameserver/nameserver.go
```diff
@@ -0,0 +1,705 @@
...
+			if record.v4Record.IsUnspecified() != true {
+				r := func() dns.RR {
+					r := new(dns.A)
+					r.A = record.v4Record
+					r.Hdr = dns.RR_Header{Name: name + ".", Rrtype: dns.TypeA,
+						Class: dns.ClassINET, Ttl: 64}
+					return r
+				}()
+				rr = append(rr, r)
+				if len(rr) >= 10 {
```
rchirakk: selected based on  the number of response that can fit in response.

netplugin/nameserver/nameserver.go
jojimt: 1 day is definitely not appropriate. Anything of the order of minutes should be fine. Especially considering this is inline, the 64 sec may be ok.

netplugin/nameserver/nameserver.go
dseevr: @DivyaVavili not just for your understanding 😄 @rchirakk could you leave a comment (in the code) as to how this value was selected?  bare numeric constants are kind spooky if it's not ultra obvious why they were used

netplugin/nameserver/nameserver.go
dseevr: @DivyaVavili  there's no difference between `new(dns.AAAA)` and `&dns.AAAA{}` in this case... the & syntax is just a lot more common in my experience

(`new()` can get pointers to types which can't be initialized with &, and that's the only time you *have* to use it)

netplugin/nameserver/nameserver.go
rchirakk: it happens only in add functions, it's too small for a function

rchirakk:
@DivyaVavili PTAL

netplugin/nameserver/nameserver.go
DivyaVavili: dns.RR is an interface. As long as type dns.A implements the dns.RR interface(which it does), you can directly assign it or use it in place of RR. There is no need of type conversion here.

netplugin/nameserver/nameserver.go
DivyaVavili: @dseevr: I meant the whole initialization for dns.RR. See my previous comment.

dseevr: @DivyaVavili ahh, i was confused because the comment was on this line 😄

netplugin/nameserver/nameserver.go
DivyaVavili: Yes, a couple of minutes is fine. 64 is just a bit network geeky to me :) I'm OK with 2 min.

netplugin/nameserver/nameserver.go
DivyaVavili: @dseevr: Guessed so 😄

netplugin/nameserver/nameserver.go
rchirakk: yes, it is added to a slice, that is why the type conversion

DivyaVavili: You can append to the interface slice without the type conversion

netplugin/nameserver/nameserver.go
```diff
@@ -0,0 +1,710 @@
...
+			}
+		}
+		if tenantTables.endpointTbl != nil {
+			delete(tenantTables.endpointTbl, eps.EndpointID)
+		}
+	}
+}
+
+func (ens *NetpluginNameServer) incTenantStats(tenant string, name string) {
+	ens.stats.RWMutex.Lock()
```
DivyaVavili: RWMutex doesn't need to be called explicitly here. You can call Lock/Unlock directly on stats as you have done in the next step. Please do the same for tenant map

netplugin/nameserver/nameserver.go
rchirakk: yes, updated

netplugin/nameserver/nameserver.go
rchirakk: yes, we can use embedding, updated

DivyaVavili:
LGTM. We can merge once the sanities pass.

PR #661, opened 2016-12-15
remove skydns, use inline dns service instead.

rchirakk:
remove skydns, use inline dns instead.


unclejack:
@rchirakk Can you update this, please? I've opened #687 to clean up the Vagrantfiles and remove the dnscontainer.tar Docker image.

jojimt:
@DivyaVavili PTAL

PR #669, opened 2016-12-21
LACP monitoring

DivyaVavili:
This PR includes changes to monitor lacp status in interface table and communicate the changes to ofnet.

Test cases executed:
1. Remove member link and check that it is not used by control plane to send any packets
2. Add member link and check that it is included in the list of active links and the packets are load balanced
3. Check multiple link updates at the same time


jojimt:
LGTM

PR #671, opened 2016-12-22
Godeps,vendor: bump remotessh to include pty fixes

unclejack:
This PR updates remotessh to the latest current version. This new version contains the fix which makes sudo work.


gaurav-dalvi:
LGTM

PR #680, opened 2017-01-05
upgrade uplink changes

DivyaVavili:
PR handles upgrade scenario for uplinks. 
**Cases:** 
- If the previous uplink was an individual interface, it is silently bonded to form a port bond
- If the uplink interfaces are already part of the OVS bridge, the user is warned that the interfaces are being made of the port bond and proceeds

### DIFF
--- a/drivers/ovsSwitch.go
+++ b/drivers/ovsSwitch.go
@@ -19,6 +19,8 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"reflect"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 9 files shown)


DivyaVavili:
Fixes contiv/ccn#79

drivers/ovsSwitch.go
```diff
@@ -572,6 +574,57 @@ func (sw *OvsSwitch) DeleteVtep(vtepIP string) error {
...
+
+	/* Cleanup any other individual ports that may exist */
+	for _, intf := range intfList {
+		if sw.ovsdbDriver.IsIntfNamePresent(intf) {
+			log.Infof("Deleting old uplink port: %+v", intf)
+			err = sw.ovsdbDriver.DeletePort(intf)
+			if err != nil {
+				break
+			}
+			if !portCreateReq {
```
jojimt: This check can be eliminated.

jojimt:
@DivyaVavili pls retrigger sanity

PR #682, opened 2017-01-06
Initial creation of cleanup script for Jenkins job.

dojacobs:
New cleanup script for Jenkins CI jobs, such as Netplugin_CI.

Attempts to forcibly remove every VirtualBox VM that is on the Jenkins Node.

This will be run as a post-build script, and will hopefully solve a number of issues CI has been having when it runs on a "dirty" VirtualBox environment. 

I tested this script on my test Jenkins server with a copy of Netplugin_CI.

### DIFF
--- a/scripts/jenkins_cleanup.sh
+++ b/scripts/jenkins_cleanup.sh
@@ -16,7 +16,7 @@ rm -rf /home/admin/VirtualBox\ VMs/* || true
 rm -rf .vagrant/* || true
 rm -f *.vdi || true
 
-for f in $(vboxmanage list vms | grep inaccessible | awk {'print $2'} | cut -d'{' -f2 | cut -d'}' -f1); do
+for f in $(vboxmanage list vms | awk {'print $2'} | cut -d'{' -f2 | cut -d'}' -f1); do
     echo $f
     vboxmanage unregistervm --delete $f || true
 done
--- a/scripts/jenkins_cleanup.sh
+++ b/scripts/jenkins_cleanup.sh
@@ -1,0 +1,22 @@
+#/!/bin/bash
+
+# Script run on the Jenkins Node after the build(tests) are completed.
+# Forcibly cleans up _ALL_ VirtualBox VMs, not just those created during
+# the Jenkins run.
+
+# Doing each command with "|| true" so that even if a command fails, it won't
+# cause Jenkins to mark the build as failed.
+
+set -e
+
+cd $WORKSPACE/src/github.com/contiv/netplugin
+vagrant destroy -f || true
+
+rm -rf /home/admin/VirtualBox\ VMs/* || true
+rm -rf .vagrant/* || true
+rm -f *.vdi || true
+
+for f in $(vboxmanage list vms | grep inaccessible | awk {'print $2'} | cut -d'{' -f2 | cut -d'}' -f1); do
+    echo $f
+    vboxmanage unregistervm $f || true
+done
--- a/scripts/jenkins_cleanup.sh
+++ b/scripts/jenkins_cleanup.sh
@@ -18,5 +18,5 @@ rm -f *.vdi || true
 
 for f in $(vboxmanage list vms | grep inaccessible | awk {'print $2'} | cut -d'{' -f2 | cut -d'}' -f1); do
     echo $f
-    vboxmanage unregistervm $f || true
+    vboxmanage unregistervm --delete $f || true
 done


gaurav-dalvi:
Thanks @dojacobs for taking care of this. LGTM.
Let us wait for @jojimt to merge this

unclejack:
@dojacobs Have you tried to run `make start` after running this script with Vagrant 1.8?

dojacobs:
Yes, my jenkins node is running Vagrant 1.8.7.

Again this script is intended only for use by Jenkins to clean up VirtualBox after each run of the Netplugin_CI job.  It is a modified form of Bill's "scorch earth" alias.

DivyaVavili:
@dojacobs: Just for my understanding, why is this change required now? I believe the stale-vm issue is something that we started seeing just recently. Has anything changed recently to trigger this? I'm not sure if we need to integrate jenkins specific scripts into netplugin. Could it be made part of the CI process itself?

scripts/jenkins_cleanup.sh
```diff
@@ -0,0 +1,22 @@
...
+cd $WORKSPACE/src/github.com/contiv/netplugin
+vagrant destroy -f || true
+
+rm -rf /home/admin/VirtualBox\ VMs/* || true
+rm -rf .vagrant/* || true
+rm -f *.vdi || true
+
+for f in $(vboxmanage list vms | grep inaccessible | awk {'print $2'} | cut -d'{' -f2 | cut -d'}' -f1); do
+    echo $f
+    vboxmanage unregistervm $f || true
```
unclejack: This should be `vboxmanage unregistervm --delete` if we'd like to remove those VMs.

unclejack:
LGTM

dojacobs:
@DivyaVavili :  We've been seeing problems with Jenkins' nodes for quite some time, and with differing symptoms that all eventually pointed to a dirty VirtualBox environment.  The old solution was to simply offline the node until someone (Erik?) could manually clean it, or reinstall VirtualBox.

While Netplugin_CI does call "make stop" as a cleanup step this is apparently inadequate for the purposes of CI.  Even after a few successful runs of Netplugin_CI on my dev node, enough cruft and orphaned VMs built up to recreate the same issues you were seeing with the other Contiv nodes.  CI is not very useful if it still requires manual intervention all the time.

Since this script is only called from Netplugin_CI, I put it in the netplugin/scripts directory.  It is only intended to be used by Jenkins - not developers or customers.  Maybe at some point we can consider creating a separate CI repository and move this script over there.

Finally, while it is possible to enter raw shell commands into Jenkins as a step, it is better to make Jenkins call scripts, which can be managed by GIT as any other file.

dojacobs:
@jojimt  - Please merge so I can update Jenkins' Netplugin_CI to start using this.

scripts/jenkins_cleanup.sh
```diff
@@ -0,0 +1,22 @@
...
+
+cd $WORKSPACE/src/github.com/contiv/netplugin
+vagrant destroy -f || true
+
+rm -rf /home/admin/VirtualBox\ VMs/* || true
+rm -rf .vagrant/* || true
+rm -f *.vdi || true
+
+for f in $(vboxmanage list vms | grep inaccessible | awk {'print $2'} | cut -d'{' -f2 | cut -d'}' -f1); do
+    echo $f
```
jojimt: You might want to delete runningvms as well.

dseevr: `vagrant destroy -f` at the top will handle the running ones.  This is just to handle any which may have gotten into a weird state rather than shutting down cleanly.

We can always amend this script if we find more edge cases

dseevr:
@jojimt this is ready for squash and merge

jojimt:
Yes, once the build finishes

dojacobs:
So, 2 of the Netplugin_CI configs failed.
1 failed due to a repo not responding.
1 failed due to a error with K8s.

And the changes suggested to delete all VMs did not work.  I'll have to figure out why and re(re-re-re)submit this PR again.  Hold one.

dojacobs:
Script seems to be OK despite some errors from VBoxManage about being unable to delete VMs.  This might be due to a race condition between vagrant destroy -f and the other commands in the script.  The script is written in such a way that it always "Succeeds" as far as Jenkins is concerned.

Go ahead and do the Merge.

jojimt:
The test needs to pass before I can merge. Please retrigger and get it to pass.

dojacobs:
Tests passed.

dojacobs:
@joji - is there anything else needed here?  otherwise, merge when you get a chance and I'll update Jenkins to use this new script.  Thanks

PR #687, opened 2017-01-09
finish the skydns cleanup & clean up the Vagrantfile a bit

unclejack:
This PR removes the leftover scripts/dnscontainer.tar skydns image and makes a few small changes to clean up the Vagrantfile. We should use the Vagrant hostname setup.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -25,26 +25,24 @@ cluster_ip_nodes = ""
 provision_common_once = <<SCRIPT
 ## setup the environment file. Export the env-vars passed as args to 'vagrant up'
 echo Args passed: [[ $@ ]]
-echo -n "$1" > /etc/hostname
-hostname -F /etc/hostname
 
 echo 'export GOPATH=#{gopath_folder}' > /etc/profile.d/envvar.sh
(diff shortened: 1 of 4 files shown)


Vagrantfile
```diff
@@ -25,8 +25,6 @@ cluster_ip_nodes = ""
 provision_common_once = <<SCRIPT
 ## setup the environment file. Export the env-vars passed as args to 'vagrant up'
 echo Args passed: [[ $@ ]]
-echo -n "$1" > /etc/hostname
```
neelimamukiri: Can we remove the node_name parameter to provision_common_once at the caller site and I suppose shift the rest of the arguments here?

unclejack: @neelimamukiri: Thank you for the suggestion. I'll take care of it.

unclejack: @neelimamukiri: The changes required to address this comment are ready. PTAL

unclejack:
@neelimamukiri @jojimt This PR is green now. PTAL when you have the time.

gaurav-dalvi:
Do we have dependency on this : https://github.com/contiv/netplugin/pull/661

unclejack:
@gaurav-dalvi: No, these two can be worked on independently from each other.

PR #697, opened 2017-01-14
Fix instances of VLAN_IF, NETMASTER_IF references in the documentation.

neelimamukiri:
They should be __VLAN_IF__ and __NETMASTER_IF__.
Also adding a troubleshooting section to debug common issues.

### DIFF
--- a/install/k8s/README.md
+++ b/install/k8s/README.md
@@ -6,8 +6,9 @@ Install kubernetes 1.4 or higher using http://kubernetes.io/docs/getting-started
 
 ### Install Contiv
 
-* Replace all instances of NETMASTER_IP in contiv/contiv.yaml with the master IP
-* Replace VLAN_IF with the data plane interface
+* Replace all instances of `__NETMASTER_IP__` in contiv/contiv.yaml with the master IP.
+* Replace `__VLAN_IF__` with the data plane interface.
+  If there is no requirement to create vlan based networks there is no need for a seperate data interface and `__VLAN_IF__` can be set to "". If vlan based networks are to be created then a separate data interface is mandatory which can be set appropriately.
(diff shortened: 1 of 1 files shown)


gaurav-dalvi:
LGTM.

install/k8s/README.md
```diff
@@ -6,8 +6,8 @@ Install kubernetes 1.4 or higher using http://kubernetes.io/docs/getting-started
 
 ### Install Contiv
 
-* Replace all instances of NETMASTER_IP in contiv/contiv.yaml with the master IP
-* Replace VLAN_IF with the data plane interface
+* Replace all instances of `__NETMASTER_IP__` in contiv/contiv.yaml with the master IP.
+* Replace `__VLAN_IF__` with the data plane interface. If there is only one interface set `__VLAN_IF__` to `""` or delete the VLAN_IF section.
 * Optional: Replace the contiv version(v0.1-11-30-2016.20-08-20.UTC) with the desired release/test version.
```
abhinandanpb: minor nit: I guess we can say if there is no requirement to create vlan based networks there is not need for a seperate data interface and VLAN_IF can be set to "". If vlan based networks are to be created then a separate data interface is mandatory which can be set appropriately. 

Also does the __VLAN_IF__ support multiple comma seperated interfaces ? Since contiv can support it in bridge mode. Probably worth a mention.

install/k8s/README.md
```diff
@@ -6,8 +6,8 @@ Install kubernetes 1.4 or higher using http://kubernetes.io/docs/getting-started
...
 * On the management node, run
```
rhim: This needs to sound less prescriptive :-). Something like, on the management node, create a default network and default epg. Example follows:
...

unclejack:
The CI failure is irrelevant.

LGTM

abhinandanpb:
LGTM

PR #700, opened 2017-01-18
Add k8s yaml file to support ACI

neelimamukiri:


install/k8s/README.md
```diff
@@ -6,14 +6,31 @@ Install kubernetes 1.4 or higher using http://kubernetes.io/docs/getting-started
...
+    kubectl create secret generic aci.key --from-file=<path name of aci.key file> -n kube-system
+    ```
 * On the management node, run
 ```sh
 kubectl apply -f contiv.yaml
 ```
 * Get netctl from a Contiv release or local build. Contiv releases are available from https://github.com/contiv/netplugin/releases
+* Optional ACI only step:
+```sh
+netctl set global --fwd-mode routing
```
abhinandanpb: This is not a valid config in aci mode.

gaurav-dalvi: `netctl global set --fabric-mode aci --vlan-range 1150-1170`

vlan range must be configured in APIC before hand !

neelimamukiri: Thanks for checking. I've updated with the correct instruction for ACI/routing.

abhinandanpb:
LGTM

PR #711, opened 2017-01-23
fix CI related failures

### LINKED ISSUE #703: Make netplugin_ci more robust to repo failures
netplugin_ci relies on some external repos for the build, namely epel and ceph.  Unfortunately, there are a lot of transitory network issues that cause one or the other to become unavailable, and then that particular netplugin_ci job fails, but the entire PR job fails as well.

Should make the build script or Jenkins detect these failures, sleep for a bit, and then start over from the beginning.  (for a maximum of number of retries.)

This should hopefully reduce the number of failed PR jobs due to stupid network/repository issues.

unclejack:
This PR makes a change to fix the ceph and other repository related errors.

Fixes #703

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -52,6 +52,11 @@ fi
 # Change ownership for gopath folder
 chown -R vagrant #{gopath_folder}
 
+# Remove the unneeded ceph repository if it exists
+echo "Remove the unneeded ceph repository if it exists"
+rm -f /etc/yum.repos.d/ceph.repo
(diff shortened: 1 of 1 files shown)


unclejack:
This is probably quicker to check on a local machine. Everything should be fine if the provisioning succeeds.

unclejack:
@dojacobs @jojimt @rhim The status of this PR is green. Please take a look when you have the time.

PR #717, opened 2017-01-25
Expose clustermode via global oper

jojimt:


jojimt:
@abhinandanpb would you take a look?

abhinandanpb:
LGTM

PR #721, opened 2017-01-26
Vagrantfile: bump Go to 1.7.5

unclejack:
This bumps Go to the just released Go 1.7.5.

The changelog can be seen here: https://github.com/golang/go/issues?q=milestone%3AGo1.7.5


dseevr:
This is very good because it includes the memory corruption bug fix which was backported from the forthcoming 1.8 release:  https://github.com/golang/go/issues/18410

PR #722, opened 2017-01-27
update the default data interface as per Vagrantfile

rchirakk:


unclejack:
This change has no effect on the actual Vagrant environment used for development and test. This file is generated from the Makefile when `make system-test` is executed.

dseevr:
@unclejack does this file even need to be checked in in that case?

unclejack:
@dseevr We don't need to have it in here. However, we should replace it with documentation on how to generate such a file by hand (or how to call the tool which does it).

rchirakk:
@unclejack @dseevr 
It is the default configuration file and the network  interface config depends on the Vagrantfile.
In case you want to run a part of system test 
"go test -v -timeout 480m ./test/systemtests -check.v -check.f "TestBasicXXX"
then you need this config file or you need to generate one.

PR #728, opened 2017-02-03
node_test.go: avoid scp fingerprint prompt

unclejack:
This PR fixes an issue related to copying files via scp. scp will try to prompt the user to say if they trust the fingerprint of the host. This happens when the ssh and scp commands have never been used to connect to that host from the host which initiates the connection.


gaurav-dalvi:
LGTM

neelimamukiri:
Not sure if its relevant here, but I am also using  UserKnownHostsFile=/dev/null for the installer tests to avoid adding the host/key to the known_hosts.

jojimt:
LGTM

PR #739, opened 2017-02-10
Fix for k8 and k8 sanity 

abhinandanpb:
The change contains the following : 
1) New k8 base boxes for both demo and sanity (contains built in guestadditions and basic tools)
2) Bug fixes for etcd flaps ,netplugin/netmaster bring up scenarios
3) fix to syncronize netplugin operations
4) timing issues in system test

### DIFF
--- a/netplugin/agent/state_event.go
+++ b/netplugin/agent/state_event.go
@@ -62,8 +62,6 @@ func processInfraNwCreate(netPlugin *plugin.NetPlugin, nwCfg *mastercfg.CfgNetwo
 	log.Infof("Got endpoint create resp from master: %+v", mresp)
 
 	// Take lock to ensure netPlugin processes only one cmd at a time
-	netPlugin.Lock()
-	defer func() { netPlugin.Unlock() }()
 
 	// Ask netplugin to create the endpoint
 	netID := nwCfg.NetworkName + "." + nwCfg.Tenant
@@ -116,8 +114,6 @@ func processNetEvent(netPlugin *plugin.NetPlugin, nwCfg *mastercfg.CfgNetworkSta
 	// Also network create events need to be processed before endpoint creates
 	// and reverse shall happen for deletes. That order is ensured by netmaster,
 	// so we don't need to worry about that here
-	netPlugin.Lock()
-	defer func() { netPlugin.Unlock() }()
 
 	operStr := ""
 	if isDelete {
@@ -143,8 +139,6 @@ func processEpState(netPlugin *plugin.NetPlugin, opts core.InstanceInfo, epID st
 	// Also network create events need to be processed before endpoint creates
 	// and reverse shall happen for deletes. That order is ensured by netmaster,
 	// so we don't need to worry about that here
-	netPlugin.Lock()
-	defer func() { netPlugin.Unlock() }()
 
 	// read endpoint config
 	epCfg := &mastercfg.CfgEndpointState{}
@@ -182,8 +176,6 @@ func processBgpEvent(netPlugin *plugin.NetPlugin, opts core.InstanceInfo, hostID
 		log.Debugf("Ignoring Bgp Event on this host")
 		return err
 	}
-	netPlugin.Lock()
-	defer func() { netPlugin.Unlock() }()
 
 	operStr := ""
 	if isDelete {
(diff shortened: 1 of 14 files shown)


unclejack:
@abhinandanpb: Please split this up into logical chunks. We should have one commit for the Kubernetes Vagrantfile and one for each set of changes. There are too many changes to put in a single commit.

Do we still need multiple boxes to run the k8s tests?

abhinandanpb:
I would have split if my commits for each issue was huge . commits are inter dependent.
We need k8 boxes until Divya's kubeadm changes are going to be merged which will happen pretty soon.

unclejack:
I understand, but it's more difficult to understand which issue each change addresses. That's the reason why I've asked that question.

rhim:
Looking at the code, there are three logical parts I think:
1. netplugin synchronization issues (majority of this PR)
2. etcd flap (small)
3. Pre-built vagrant boxes for k8s-centos to expedite k8s test setup. (small)

Maybe keeping 1 in this PR and moving 2 and 3 into another PR will clean up the confusion. Just a suggestion.

unclejack:
@abhinandanpb This needs to be rebased on master.

netplugin/agent/state_event.go
```diff
@@ -323,8 +307,6 @@ func processServiceLBEvent(netPlugin *plugin.NetPlugin, svcLBCfg *mastercfg.CfgS
 	portSpecList := []core.PortSpec{}
 	portSpec := core.PortSpec{}
 
-	netPlugin.Lock()
-	defer func() { netPlugin.Unlock() }()
```
unclejack: Did you run into an issue with `defer netplugin.Unlock()`?

abhinandanpb: not really. I just used this convention.

netplugin/plugin/netplugin.go
```diff
@@ -241,6 +288,7 @@ func (p *NetPlugin) Reinit(cfg Config) {
 	}
 
 	defer func() {
```
jojimt: If the function returns before hitting this line, would we end up still holding the lock?

test/systemtests/trigger_test.go
```diff
@@ -169,7 +169,7 @@ func (s *systemtestSuite) TestTriggerNetpluginDisconnect(c *C) {
 			c.Assert(node.startNetplugin(""), IsNil)
 
 			c.Assert(node.exec.runCommandUntilNoNetpluginError(), IsNil)
-			time.Sleep(20 * time.Second)
+			time.Sleep(30 * time.Second)
```
jojimt: Instead of increasing this, can we increase the number of retries?

test/systemtests/trigger_test.go
```diff
@@ -460,7 +462,7 @@ func (s *systemtestSuite) TestTriggers(c *C) {
 				c.Assert(node.rotateLog("netplugin"), IsNil)
 				c.Assert(node.startNetplugin(""), IsNil)
 				c.Assert(node.exec.runCommandUntilNoNetpluginError(), IsNil)
-				time.Sleep(20 * time.Second)
+				time.Sleep(30 * time.Second)
```
jojimt: Same here.

netplugin/agent/state_event.go
unclejack: Can you change all of these to `defer netplugin.Unlock()`, please? The `defer func() {}` should be used only when there's more than one line of code to run in defer.

rhim:
Fix the specific lines for this PR if any, and file an issue for cleanup for later please.

Thanks,
- Himanshu
     .:|:.:|:.

netplugin/agent/state_event.go
abhinandanpb: This code has been removed. not added.

netplugin/plugin/netplugin.go
abhinandanpb: agreed. It doesnt stack up the defers. Looks like we might have bug also in the previous code. will fix it.

test/systemtests/trigger_test.go
abhinandanpb: i guess this needs a revisit as well. I open an issue to track this.

netplugin/agent/state_event.go
unclejack: @abhinandanpb: I'm sorry. I've missed that.

PR #745, opened 2017-02-16
Fixing ovs version in container

neelimamukiri:
We have only tested with ovs 2.3 and 2.5, so hardcoding the later
version here. We need to update this as our test matrix changes.
Also adding -d flag to allow debug messages for netplugin and netmaster.
Ubuntu 16 upgrade is required to get the newer ovs version.


abhinandanpb:
LGTM. Thanks for taking care of this.

neelimamukiri:
@jojimt, PTAL when you get a chance. @abhinandanpb was debugging a BGP setup where he found that OVS version being used it 2.4 and as we've tested with 2.3 and 2.5 only, I've made this change to bump the version. He also needed the debug flag for verbose logs.

PR #750, opened 2017-02-18
Docker v2plugin support

gkvijay:
+ Makefile targets (demo-v2plugin) for the complete workflow
  and for subtasks like creating/restarting plugin
+ Stubs for new libnetwork apis
+ Changes to use the name of v2plugin in docknet apis
+ Scripts to create the v2plugin
+ Python supporting scripts for v2plugin start/stop
+ Updated docker and libnetwork libraries


Vagrantfile
```diff
@@ -67,7 +68,18 @@ else
...
+    if [[ #{docker_version} == *"rc"* ]]; then
+        echo "Getting pre-release docker version #{docker_version} "
+        curl -fsSL https://test.docker.com/ | sh
+    elif [[ #{docker_version} == *"cs"* ]]; then
+            echo "Getting cs docker version #{docker_version} "
+            curl -fsSL https://packages.docker.com/1.13/install.sh | sh
+    else
+        echo "Getting released docker version #{docker_version} "
+        curl https://get.docker.com | sed s/docker-engine/docker-engine-#{docker_version}/ | bash
+            echo "Getting released docker version #{docker_version} "
```
jojimt: These lines repeated by mistake?

Vagrantfile
```diff
@@ -298,6 +310,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
                   ENV["CONTIV_CLUSTER_STORE"] || "etcd://localhost:2379",
                   ENV["CONTIV_DOCKER_VERSION"] || docker_version,
                   ENV['CONTIV_NODE_OS'] || "",
+                  ENV['CONTIV_V2PLUGIN_NAME'] || "contiv/v2netplugin:0.1",
```
jojimt: I think we should probably use the netplugin version string as tag, instead of 0.1. But, this is ok for now.

install/v2plugin/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+FROM ubuntu:16.04
+MAINTAINER Cisco Contiv (http://contiv.github.io/)
+
+RUN mkdir -p /run/docker/plugins
+RUN mkdir -p /etc/openvswitch
+RUN mkdir -p /var/run/contiv/log
+
+RUN apt-get update
+RUN apt-get --assume-yes  install openvswitch-switch
```
jojimt: We need to specify the ovs version here. Please check the latest install scripts. @neelimamukiri had added this.

mgmtfn/dockplugin/netDriver.go
```diff
@@ -371,6 +371,58 @@ func leave(w http.ResponseWriter, r *http.Request) {
...
+func networkAllocate(w http.ResponseWriter, r *http.Request) {
+	resp := api.AllocateNetworkResponse{}
+
+	content, err := json.Marshal(resp)
+	if err != nil {
+		httpError(w, "Could not generate a response", err)
+		return
+	}
+
+	w.Write(content)
```
jojimt: Is this just a stub function? Can we eliminate the redundant return statement?

mgmtfn/dockplugin/netDriver.go
```diff
@@ -371,6 +371,58 @@ func leave(w http.ResponseWriter, r *http.Request) {
...
+	return
+}
+
```
jojimt: Same comment for all the above routines.

netmaster/main.go
```diff
@@ -23,12 +23,14 @@ import (
...
 	log "github.com/Sirupsen/logrus"
 	"github.com/contiv/netplugin/netmaster/daemon"
+	"github.com/contiv/netplugin/netmaster/docknet"
 	"github.com/contiv/netplugin/version"
 )
 
 type cliOpts struct {
 	help         bool
 	debug        bool
+	pluginName   string
```
jojimt: Is there a need to pass the plugin name?

Godeps/Godeps.json
```diff
@@ -1,7 +1,7 @@
 {
 	"ImportPath": "github.com/contiv/netplugin",
 	"GoVersion": "go1.7",
-	"GodepVersion": "v78",
+	"GodepVersion": "v75",
```
jojimt: You should probably update your godep version.

Vagrantfile
gkvijay: location to fetch the rc (release candidate) and cs (commercial support) versions of docker are different from the released versions

Vagrantfile
gkvijay: Just put the default name as contiv/netplugin:0.1, but we can override this with the env variable. Will update the release scripts to create/release plugin with release version tag as a new PR

mgmtfn/dockplugin/netDriver.go
gkvijay: Eventhough we dont do any processing of these APIs, we need to return a valid response to avoid failure in dockerd. Will remove the return

mgmtfn/dockplugin/netDriver.go
gkvijay: will remove the redundant return

netmaster/main.go
gkvijay: netmaster needs to know the plugin name so that it can use it in the docknet apis to create the docker network using the proper plugin name

Godeps/Godeps.json
gkvijay: will update godep version and push

install/v2plugin/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+RUN apt-get update && \
+    apt-get --assume-yes install openvswitch-switch=2.5.0-0ubuntu1 && \
+    apt-get --assume-yes install net-tools iptables && \
+    apt-get --assume-yes install iputils-ping netcat
+
+COPY netplugin netmaster netctl startcontiv.sh /
+
+ENTRYPOINT ["/startcontiv.sh"]
+#CMD ["/startcontiv.sh"]
```
rhim: Remove.

Vagrantfile
rhim: Why use a separate versioning scheme for v2netplugin container image? Shouldn't we use the same version as what is in netplugin/version/CURRENT_VERSION?

install/v2plugin/Dockerfile
gkvijay: will do

Vagrantfile
gkvijay: this is a default version for dev and the plugin is not pushed to hub. In the release script we should use the correct version and push it to the hub. Will add that in a following PR. Need to discuss that with you, joji and neelima

gkvijay: Need to decide the official name of released plugin as well

Vagrantfile
jojimt: True, but you have the echo and curl repeated twice in this else block...

gkvijay: Thanks. I missed that. Removed the repeated lines

install/v2plugin/Dockerfile
```diff
@@ -0,0 +1,15 @@
+# Docker v2plugin container with OVS / netplugin / netmaster 
+
+FROM ubuntu:16.04
+MAINTAINER Cisco Contiv (http://contiv.github.io/)
+
+RUN mkdir -p /run/docker/plugins /etc/openvswitch /var/run/contiv/log
```
unclejack: Please change this and the lines below to the following:
```
RUN mkdir -p /run/docker/plugins /etc/openvswitch /var/run/contiv/log \
 && apt-get update \
 && apt-get install -y openvswitch-switch=2.5.0-0ubuntu1 \
        net-tools iptables iputils-ping netcat \
 && rm -f /var/lib/apt/lists/*
```

gkvijay: Updated with the change

Vagrantfile
```diff
@@ -67,7 +68,16 @@ else
     # cleanup openstack-kilo repo if required
     yum remove docker-engine -y || :
     yum-config-manager --disable openstack-kilo
-    curl https://get.docker.com | sed s/docker-engine/docker-engine-#{docker_version}/ | bash
+    if [[ #{docker_version} == *"rc"* ]]; then
```
unclejack: These changes will become redundant soon. The Docker engine versioning has changed. The `-cs` edition won't exist any more. The regular Docker engine version will become `-ce` (community edition).

Can we skip these changes for now, please?

Godeps/Godeps.json
```diff
@@ -128,13 +128,18 @@
 		},
 		{
 			"ImportPath": "github.com/docker/docker/pkg/plugins",
-			"Comment": "v1.4.1-13296-ge904f58",
-			"Rev": "e904f58e199836fa43ee1c6736e689ef58e4fe10"
+			"Comment": "docs-v1.12.0-rc4-2016-07-15-3221-gccba791",
```
unclejack: Why is the rc4 version of Go used here? We should use stable versions.

Godeps/Godeps.json
```diff
@@ -128,13 +128,18 @@
 		},
```
unclejack: All the Godeps changes should be in their own commit. This will help us track down potential issues caused by this change.

The commit which updates these libraries should be the first one. This PR is quite big and we should split it into multiple commits.

Vagrantfile
gkvijay: Since we are working with pre-released docker versions as well, we need the rc version. cs is the customer support version which was suggested by docker folks.

Godeps/Godeps.json
jojimt: @unclejack I don't think what you're asking is always possible. Many times godep update will require corresponding changes to just make things compile.

dseevr: @jojimt I think he means:

```sh
make godep-save
git add Godeps
git add vendor
git commit -m "updated Godeps"
```

Just so that it generates a separate commit, not a separate PR.  As an example, I just did this here:  https://github.com/contiv/objdb/pull/20

gkvijay: Split it into two commits as suggested by @unclejack

install/v2plugin/Dockerfile
gkvijay: updated the ovs version in Dockerfile

Vagrantfile
unclejack: The `-rc` thing should be OK. The `-cs` part will go away soon. There's no point in adding something for the `-cs` versions of Docker at this point. https://github.com/docker/docker/pull/31075

Godeps/Godeps.json
```diff
@@ -624,6 +629,111 @@
 			"ImportPath": "github.com/coreos/go-semver/semver",
 			"Comment": "v0.2.0-5-g5e3acbb",
 			"Rev": "5e3acbb5668c4c3deb4842615c4098eb61fb6b1e"
+		},
+		{
+			"ImportPath": "github.com/docker/docker/vendor/github.com/Sirupsen/logrus",
```
rchirakk: Is the nested vendoring going to create multiple versions of logrus/other packages in netplugin?

netmaster/main.go
```diff
@@ -90,6 +96,8 @@ func execOpts(opts *cliOpts) {
 	if opts.debug {
 		log.SetLevel(log.DebugLevel)
 	}
+
+	docknet.UpdatePluginName(opts.pluginName)
```
rchirakk: do you need to check if netmaster is started in docker mode here ?

gkvijay: Here the plugin name is being updated by netmaster when it starts

Godeps/Godeps.json
gkvijay: updated to avoid nested vendoring by picking specific packages

unclejack:
The tests pass for the regular netplugin setup without the v2 plugin.

I'll try to run the systemtests with the v2 plugin.

unclejack:
@gkvijay: This doesn't hook into test/systemtests. This is required to run the tests with this plugin.

gkvijay:
@unclejack, v2plugin is not enabled for systemtest. That will be tested/fixed as a separate PR. We need an alpha quality plugin build ready for plugin certification with docker

PR #751, opened 2017-02-21
Makefile: make a few small fixes & refactor

unclejack:
This PR makes the following changes to the Makefile:
- removes trailing whitespace & unneeded tabs
- use the PKG_DIRS variable to pass arguments to the targets invoked by checks; fixes weird behavior
- moves the shell completion installation to its own target & uses it where it makes sense (in the container based builds and in the vagrant VMs) to not require sudo when running `make run-build`
- ~makes the all-CI target run the checks and attempt to compile before spinning up VMs~ removed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 
 DEFAULT_DOCKER_VERSION := 1.12.6
 SHELL := /bin/bash
-EXCLUDE_DIRS := bin docs Godeps scripts test vagrant vendor install 
+EXCLUDE_DIRS := bin docs Godeps scripts test vagrant vendor install
 PKG_DIRS := $(filter-out $(EXCLUDE_DIRS),$(subst /,,$(sort $(dir $(wildcard */)))))
 TO_BUILD := ./netplugin/ ./netmaster/ ./netctl/netctl/ ./mgmtfn/k8splugin/contivk8s/ ./mgmtfn/mesosplugin/netcontiv/
 HOST_GOBIN := `if [ -n "$$(go env GOBIN)" ]; then go env GOBIN; else dirname $$(which go); fi`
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -30,8 +30,8 @@ all: build unit-test system-test ubuntu-tests
 # 'all-CI' target is used by the scripts/CI.sh that passes appropriate set of
 # ENV variables (from the jenkins job) to run OS (centos, ubuntu etc) and
 # sandbox specific(vagrant, docker-in-docker)
-all-CI: stop clean start
-	make ssh-build
+all-CI: stop clean
+	make checks run-build start ssh-build
```
jojimt: This means that the host needs to have build tools installed AND that every time we upgrade the build tools (e.g. go ver) we will need to update every single host.

unclejack: This is already necessary now. The system tests are compiled on the host anyway.

The code is compiled twice now. The first build is done on the host to ensure that the code can be compiled. The second build is done to ensure that we run binaries netplugin, netmaster and netctl binaries built with the Go version specified in our Makefile.

unclejack: This only applies to the CI itself. It doesn't affect other environments.

jojimt: @unclejack compiling the system tests and compiling netplugin are different. I am confused as to what value this change brings. We are not really expecting anyone to push a change that does not even compile. Why add an additional step?

unclejack: This is how things were before this PR:
1. stop VMs if they were running (not the case, no VMs were on, safe to do this nonetheless)
2. clean up
3. start the VMs & provision them - takes ~10 minutes
4. run make checks as part of ssh-build - this connects to the first node to run `make checks` and compile everything
5. the tests are run

This is how things would be with this PR:
1. stop the VMs if they were running (same as before)
2. clean up
3. run `make checks` on the host to ensure the code passes the checks (no formatting issues and so on)
4. run `make run-build` to ensure the code compiles
5. start the VMs - takes ~10 minutes
6. run `make ssh-build`, just like step 4 before
7. the tests are run as before

This is an optimization to avoid provisioning the VMs. It takes about 10 minutes to boot and provision the three VMs on the CI.

The advantage of this PR is that we give the developer feedback sooner if there's something wrong with the code. We also avoid wasting time and keeping three CI nodes busy for 10 minutes or more.

unclejack:
@jojimt: Could you let me know what changes you'd like me to make to this PR, please?

unclejack:
I've reverted the changes made to the CI target. The other changes are needed for other improvements and other work.

@jojimt: PTAL

PR #756, opened 2017-02-22
move key-store watch to go routine

rchirakk:
1. move key-store watch to go routine to unblock
2. add delay between restarts.


jojimt:
LGTM. Good catch.

PR #758, opened 2017-02-23
systemtests: make iperf bandwidth checks parallel

unclejack:


abhinandanpb:
@unclejack we cant parallelize iperf for vagrant vms. The bandwidth usage reduces drastically due to parallelization. So the tests will always pass because the bandwidth will be less than the expected policy.

unclejack:
@abhinandanpb: This runs at most three containers in parallel when we use the automatically generated cfg.json for the system tests.

We run tests with low bandwidth limits between 6 Mbps and 19Mbps. I've done some tests with VirtualBox. It can handle 500 Mbps between VMs with netplugin. I've managed to get 25 Gbps on the same host with 16 TCP concurrent sessions. Based on the limits we set in the tests, we only go up to between 18 Mbps and 57 Mbps when we run the tests.

unclejack:
This will require a different approach. I'll close this now.

unclejack:
This requires some other changes as well. I'll close this for now in the hope it's going to be something we can add later.

PR #763, opened 2017-02-28
Docker 1.13 swarm-mode support for contiv

gkvijay:
+ Moved to docker client library in docknet (from samalba)
+ Create networks in docker with 'Attachable' option 
+ Use swarm mode as an env option and create swarm-mode cluster
+ demo-v2plugin target to create v2plugin and enable swarm-mode
+ Updated plugin to use alpine base image

### DIFF
--- a/install/v2plugin/startcontiv.sh
+++ b/install/v2plugin/startcontiv.sh
@@ -1,48 +1,70 @@
 #!/bin/bash
-#Initialize complete contiv container. Start OVS and Net Plugin
 
 set -e 
 
 ### Pre-requisite on the host
-# etcd 
-# sudo modprobe openvswitch
-# sudo mkdir -p /var/contiv
+# run a cluster store like etcd or consul
 
 mkdir -p /var/run/contiv/log
 mkdir -p /var/run/openvswitch
 mkdir -p /etc/openvswitch
 
-echo "V2 Plugin logs" > /var/run/contiv/log/plugin_bootup.log
+BOOTUP_LOGFILE="/var/run/contiv/log/plugin_bootup.log"
 
-echo "Loading OVS" >> /var/run/contiv/log/plugin_bootup.log
-(modprobe openvswitch) || (echo "Load ovs FAILED!!! run 'sudo modprobe openvswitch' from the host before re-enabling the plugin" >> /var/run/contiv/log/plugin_bootup.log && while true; do sleep 1; done)
+echo "V2 Plugin logs" > $BOOTUP_LOGFILE
 
-echo "Starting OVS" >> /var/run/contiv/log/plugin_bootup.log
+if [ $iflist == "" ]; then
+    echo "iflist is empty. Host interface(s) should be specified to use vlan mode" >> $BOOTUP_LOGFILE
+fi
+if [ $ctrl_ip != "none" ]; then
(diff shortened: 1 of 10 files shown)


install/v2plugin/startcontiv.sh
```diff
@@ -1,48 +1,71 @@
...
-echo "V2 Plugin logs" > /var/run/contiv/log/plugin_bootup.log
+BOOTUP_LOGFILE="/var/run/contiv/log/plugin_bootup.log"
 
-echo "Loading OVS" >> /var/run/contiv/log/plugin_bootup.log
-(modprobe openvswitch) || (echo "Load ovs FAILED!!! run 'sudo modprobe openvswitch' from the host before re-enabling the plugin" >> /var/run/contiv/log/plugin_bootup.log && while true; do sleep 1; done)
+echo "V2 Plugin logs" > $BOOTUP_LOGFILE
 
-echo "Starting OVS" >> /var/run/contiv/log/plugin_bootup.log
+if [ $iflist == "" ]; then
+    echo "iflist is empty. Please specify the host interface(s) to use" >> $BOOTUP_LOGFILE
```
jojimt: This should not be mandatory. If only vxlan mode is used, we don't need to specify vlan-if.

netmaster/docknet/docknet.go
```diff
@@ -160,23 +164,24 @@ func CreateDockNet(tenantName, networkName, serviceName string, nwCfg *mastercfg
...
+			IPAM:           &ipamCfg,
+			Options:        netPluginOptions,
+			Attachable:     true,
 		}
 
 		log.Infof("Creating docker network: %+v", nwCreate)
 
 		// Create network
-		resp, err := docker.CreateNetwork(&nwCreate)
+		resp, err := docker.NetworkCreate(context.Background(), docknetName, nwCreate)
```
jojimt: If this is created in background, would we get error notifications? Or is it guaranteed to eventually succeed?

netmaster/docknet/docknet.go
```diff
@@ -212,14 +217,15 @@ func DeleteDockNet(tenantName, networkName, serviceName string) error {
...
+	defaultHeaders := map[string]string{"User-Agent": "engine-api-cli-1.0"}
+	docker, err := client.NewClient("unix:///var/run/docker.sock", "v1.23", nil, defaultHeaders)
 	if err != nil {
 		log.Errorf("Unable to connect to docker. Error %v", err)
 		return errors.New("Unable to connect to docker")
 	}
 
 	// check whether the network is present in docker
-	_, err = docker.InspectNetwork(docknetName)
+	_, err = docker.NetworkInspect(context.Background(), docknetName)
```
jojimt: Same question here.

install/v2plugin/startcontiv.sh
gkvijay: Will make it a warning and continue.

netmaster/docknet/docknet.go
gkvijay: docker client library uses cancellable threads for the REST access and hence require a context to be passed. Here we are using the Background context which doesn't have a timeout or deadline. It waits until it gets a response or http error and the NetworkCreate api returns the response/error.

jojimt:
LGTM

Issue #753, opened 2017-02-21
error with netctl aci-gw set command


gaurav-dalvi:
[root@pod12-srv1 yum.repos.d]# netctl aci-gw set -n "topology/pod-1/node-201,topology/pod-1/node-202" -d "Contiv-PD" -e yes -i no
ERRO[0000] aciGw not found
see this issue with

contiv/aci-gw:02-02-2017.2.1_1h
contiv/aci-gw:12-01-2016.2.1_1h

jojimt:
Looks like a bug in netctl. Works with the GUI.

PR #765, opened 2017-02-28
Fix https://github.com/contiv/netplugin/issues/753

jojimt:
Need to allow aciGW to be created even when it is not already present because we don't create one by default.


jojimt:
@abhinandanpb PTAL

abhinandanpb:
the error check was only to avoid netctl crashes when netmaster is down. I guess in this case its going to be fine. LGTM

PR #769, opened 2017-03-02
k8s/Vagrantfile: use linked clones & tweak VMs

unclejack:
This PR makes a few changes to vagrant/k8s/Vagrantfile:
- use linked clones for the VMs
This helps speed up the creation of the VMs. This also avoids writing 16+ GB for every `make k8s-test` when creating the VMs via `vagrant up`.
- set the amount of RAM to 2 GB for all kubernetes node VMs
- ~set the number of CPUs to 2 for all Kubernetes node VMs~
- make the quagga VMs use linked clones, KVM paravirt provider and virtio interfaces
The quagga VMs are left as they were with their old CPU and RAM configuration.

`make k8s-test` was writing ~400 GB of data for a run of the tests which took around 6-7 hours. The kubernetes node VMs were swapping to the disk due to the lack of RAM memory.

### DIFF
--- a/vagrant/k8s/Vagrantfile
+++ b/vagrant/k8s/Vagrantfile
@@ -98,6 +98,18 @@ SCRIPT
 cluster = read_cluster_config
 create_etc_hosts(cluster)
 
+def customize(v)
+  # make all nics 'virtio' to take benefit of builtin vlan tag
+  # support, which otherwise needs to be enabled in Intel drivers,
+  # which are used by default by virtualbox
+  # changes the settings for the first 5 NICs, regardless of presence
+  1.upto(5) do |n|
+    v.customize ['modifyvm', :id, "--nictype#{n}", 'virtio']
+  end
+  v.customize ['modifyvm', :id, '--paravirtprovider', "kvm"]
+  v.linked_clone = true if Vagrant::VERSION >= "1.8"
+end
+
 Vagrant.configure(2) do |config|
   if ENV['CONTIV_NODE_OS'] && ENV['CONTIV_NODE_OS'] == "rhel7" then
     config.registration.manager = 'subscription_manager'
@@ -117,6 +129,11 @@ Vagrant.configure(2) do |config|
       quagga1.vm.network "private_network",
                        ip: "70.1.1.2",
                        virtualbox__intnet: "contiv_blue"
+
+      config.vm.provider 'virtualbox' do |v|
+        customize(v)
+      end
+
       quagga1.vm.provision "shell" do |s|
         s.inline = provision_quagga
         s.args = [http_proxy, https_proxy]
@@ -137,6 +154,10 @@ Vagrant.configure(2) do |config|
       quagga2.vm.network "private_network",
                        ip: "50.1.1.200",
                        virtualbox__intnet: "contiv_yellow"
+
+      config.vm.provider 'virtualbox' do |v|
+        customize(v)
+      end
 
       quagga2.vm.provision "shell" do |s|
         s.inline = provision_quagga
@@ -192,22 +213,11 @@ Vagrant.configure(2) do |config|
         c.vm.hostname = vm_name
         c.vm.network :private_network, ip: member_info["contiv_control_ip"], virtualbox__intnet: "true"
         c.vm.network :private_network, ip: member_info["contiv_network_ip"], virtualbox__intnet: network_name, auto_config: false
-
         c.vm.provider "virtualbox" do |v|
-          if vm_name == "k8master" then
-            v.memory = 2048
-          else
-            v.memory = 1024
-          end
-          # make all nics 'virtio' to take benefit of builtin vlan tag
-          # support, which otherwise needs to be enabled in Intel drivers,
-          # which are used by default by virtualbox
-          v.customize ['modifyvm', :id, '--nictype1', 'virtio']
-          v.customize ['modifyvm', :id, '--nictype2', 'virtio']
-          v.customize ['modifyvm', :id, '--nictype3', 'virtio']
+          customize(v)
+          v.memory = 2048
           v.customize ['modifyvm', :id, '--nicpromisc2', 'allow-all']
           v.customize ['modifyvm', :id, '--nicpromisc3', 'allow-all']
-          v.customize ['modifyvm', :id, '--paravirtprovider', "kvm"]
         end # v
 
         # RHEL box that is available with the CDK comes with an older version


unclejack:
I've stopped the tests on the CI. There's no point in wasting time.

vagrant/k8s/Vagrantfile
```diff
@@ -107,6 +107,18 @@ SCRIPT
 cluster = read_cluster_config
 create_etc_hosts(cluster)
 
+def customize(v, id)
```
dseevr: Why is `id` passed in here?  It's not referenced in the body of this method anywhere

dseevr:
Let's get this merged ASAP

gkvijay:
@unclejack, Hope you did L3 systemtests for this change

vagrant/k8s/Vagrantfile
unclejack: I've fixed this issue.

unclejack:
@gkvijay: I've tried to run all the Makefile targets which are related to Kubernetes. The VMs still boot properly. The tests seem to be running just as they were running before.

I've removed the change which bumped the number of virtual CPUs to two virtual to avoid overloading the host machine. Some of the VMs can use a lot of CPU.

gkvijay:
@unclejack Since you have changes for quagga vms as well, its good to verify l3 tests. Please try docker environment. Thats what L3_CI does (http://contiv-ci.ngrok.io/job/Netplugin_OnDemand_L3_CI/171/)

unclejack:
@gkvijay That's a different Makefile target. It uses the main Vagrantfile from the top level of this repository. The Vagrantfile modified by this PR would only affect make `make k8s-l3-cluster`, `make k8s-cluster` and `make k8s-test`. The VMs for Kubernetes are still starting and they're running.

Please let me know if there's anything else you'd like to discuss.

unclejack:
This has passed the CI. PTAL

dseevr:
@gkvijay PTAL when you have a minute, this will help us out with improving the k8s systemtests

unclejack:
Please take a look at this. It's still green and it's still relevant. It'd be useful when it comes to debugging the issues we run into with the Kubernetes system tests.

dseevr:
@gkvijay bump

gkvijay:
@DivyaVavili is looking into k8s systemtests. @DivyaVavili please review.

dseevr:
FWIW we should probably be using linked clones/kvm/virtio in _every_ Vagrantfile for performance reasons... I'll create an issue to track it

DivyaVavili:
I see that though linked clones have a short bring up time and save disk space, it can have performance hit. I understand that this must be over time, but I'm not sure if it would have any impact during our system-test timeframe. We should run the tests and evaluate the performance during system-tests.

@unclejack: Have you seen a significant saving in time with this change?  I see that you've mentioned that the make targets bring up the VMs and run the tests; but could you give us an estimate of the savings in time for the entire run-time?

dseevr:
@DivyaVavili where do you see linked clones having performance penalties?  I see one commenter on Quora claiming that, but it's unsubstantiated and makes no sense...

DivyaVavili:
@dseevr: The comment on Quora is new to me. The one I checked is from VMware: https://www.vmware.com/support/ws5/doc/ws_clone_typeofclone.html which is substantiated I guess. It might/might not be a problem a problem for us, but lets evaluate it/discuss it first before dismissing the possibility.

unclejack:
@DivyaVavili VMware related performance issues associated with snapshots don't matter for our use case. VMware snapshot performance is usually discussed for scenarios where VMs are persistent and they have life cycles of at least a few months. Snapshots can cause performance issues when a lot of data is written to the cloned disk. This isn't an issue for us because we write very small amounts of data during the system tests and we delete the VMs in a few hours from their creation. We don't allow the VMs to write so much to allow performance to degrade.

Since our system tests don't have to write hundreds of megabytes or tens of gigabytes (to install software, to download some huge files and so on), the snapshots are beneficial. We don't have to create and delete the parent VM every single time.

Linked clones help with the system tests. We only pay the cost of extracting each VM once. Given the size of the current Vagrant boxes we're using, this saves a lot of time.

dseevr:
FWIW I asked in #virtualbox last night if there were any performance differences with full or linked clones and the answer was "no"

If we don't enable this, we will write petabytes of data in the next year which will unnecessarily wear our drives, and we will miss out on the big time savings on CI that every other Vagrantfile we have is taking advantage of

DivyaVavili:
I've been checking up on this as well and from what I understand, it's an issue only if we have parent block modified during writing. And in our scenarios my understanding is that we would be writing any logs on the delta block. What happens to the KV store data - is that also on the delta block?

Also, do we already have this tested with other Vagrantfile(s) and in CI? If so, we can have this change; else I would rather wait until the k8s system-tests are in a proper shape. We don't want to be debugging timing issues if there is a performance penalty coz of this change.

DivyaVavili:
Or are we seeing the timing issues because of the performance on full VMs?

dseevr:
We already use this pretty much everywhere else lol

```
./contiv/netplugin/vagrant/nomad-docker/Vagrantfile:        v.linked_clone = true if Vagrant::VERSION =~ /^1.8/
./contiv/netplugin/vagrant/mesos-cni/Vagrantfile:        v.linked_clone = true if Vagrant::VERSION =~ /^1.8/
./contiv/netplugin/Vagrantfile:        v.linked_clone = true if Vagrant::VERSION >= "1.8"
./contiv/ofnet/Vagrantfile:        v.linked_clone = true if Vagrant::VERSION >= "1.8.0"
./contiv/build/vendor/ansible/Vagrantfile:                vb.linked_clone = true if Vagrant::VERSION =~ /^1.8/
./contiv/remotessh/Vagrantfile:    v.linked_clone = true if Vagrant::VERSION =~ /^1.8/
./contiv/demo/mesos/Vagrantfile:     vb.linked_clone = true
./contiv/install/cluster/Vagrantfile:      v.linked_clone = true if Vagrant::VERSION >= '1.8'
./contiv/volplugin/ansible/Vagrantfile:                vb.linked_clone = true if Vagrant::VERSION =~ /^1.8/
./contiv/volplugin/Vagrantfile:        vb.linked_clone = true if Vagrant::VERSION =~ /^1.8/
./contiv/ansible/Vagrantfile:                vb.linked_clone = true if Vagrant::VERSION =~ /^1.8/
```

(Interesting to see all the different ways people write Ruby)

DivyaVavili:
OK. For my understanding, was there any issue with vagrant 1.8 version and linked clones? What's the need for the explicit check?

dseevr:
@DivyaVavili linked_clone was added in 1.8, so if you try to set the `linked_clone` var in an earlier version it will likely raise a `NoMethodError`.  https://www.hashicorp.com/blog/vagrant-1-8/

unclejack:
This PR has been rebased and everything is working properly. These changes will save time and will enable us to run this on the CI.

vagrant/k8s/Vagrantfile
```diff
@@ -98,6 +98,18 @@ SCRIPT
 cluster = read_cluster_config
 create_etc_hosts(cluster)
 
+def customize(v)
+  # make all nics 'virtio' to take benefit of builtin vlan tag
+  # support, which otherwise needs to be enabled in Intel drivers,
+  # which are used by default by virtualbox
+  # changes the settings for the first 5 NICs, regardless of presence
+  1.upto(5) do |n|
```
DivyaVavili: There are only 3 NICs in the original Vagrantfile with nictype virtio. I don't fully understand how these translate to inside the VM - Does it create 5 NICs with these properties or does it mention that apply this property to the first 5 NICs in the VM?

Also, as I understand this, the order of the NICs matter and it's hardly coupled with the quagga VMs. Did you test it with L3? Lets not change the ordering of the NICs.

unclejack: @DivyaVavili: This doesn't change the order. That's not possible as far as I know. They're indexed and all the changes you run affects a particular one (e.g.: changing the setting for NIC with index 1 won't make it NIC with index 2; it won't affect the listing in the OS either).

I've made those changes to make the code future proof. Should we need more NICs, we don't need to make any changes for the next NICs we need.

k8s-l3-cluster is most likely broken if it hasn't been tested with the previous set of changes.

unclejack: https://github.com/contiv/netplugin/blob/master/Vagrantfile#L202 is exactly the same. We've been using this for a while in the main Vagrantfile. That Vagrantfile has been used hundreds of times since that PR was merged.

unclejack:
I'll merge this now.

gkvijay:
@rhim,@unclejack: this PR was merged even though there was a change request from a core networking team member. This is not a good practice to follow in netplugin repo.

unclejack:
@gkvijay: Please take a look at the replies to that comment. The main Vagrantfile has been modified in the exact same way a while ago.

The relevant changes are:
https://github.com/contiv/netplugin/blob/master/Vagrantfile#L202-L211
https://github.com/contiv/netplugin/blob/master/Vagrantfile#L288-L290
https://github.com/contiv/netplugin/blob/master/Vagrantfile#L313-L315
https://github.com/contiv/netplugin/blob/master/Vagrantfile#L354-L359

This PR has been open for almost 4 months. I've asked for feedback after leaving that comment and after providing additional explanation.

gkvijay:
@unclejack: Just because a PR did not get approved for a long duration you should not ignore the questions raised by a core contributor in netplugin

unclejack:
@gkvijay: I didn't ignore them. I've answered them.

DivyaVavili:
@unclejack: I'm not happy with the way this PR was handled. My reservations were not fully answered. Anyway, this is not the right forum to discuss this. Let's take it offline.

PR #770, opened 2017-03-02
Fixed a bug where values read from .vagrant-state were being set as envvars with trailing newlines attached

dseevr:
This PR fixes a bug I discovered where the inline shell provisioner in Vagrantfile can end up generating a bash script which has a newline in the middle of an `if` statement (where the Docker version is checked) if a Vagrant state file is present.  This causes the provisioning process to fail completely.


jojimt:
LGTM

PR #773, opened 2017-03-03
Install netctl with contiv installation and delete kube-dns

neelimamukiri:
setup_cluster.py had issues with comparison as the variables were not quotes, so adding the quotes as well.


jojimt:
LGTM

PR #778, opened 2017-03-10
Netplugin Container: Cleanup script and add remove option

neelimamukiri:
Fixes https://github.com/contiv/install/issues/61. This change is required along with https://github.com/contiv/install/pull/62 to cleanup OVS bridges & ports on uninstall and restart of the netplugin pods.
The netmaster container was also starting netplugin and was conflicting with the netplugin container. Modified the script to run either as netmaster or netplugin and to exit when either of those crash. If netplugin/netmaster process stops, we will continue to restart it.

### DIFF
--- a/scripts/netContain/contivNet.sh
+++ b/scripts/netContain/contivNet.sh
@@ -1,30 +1,36 @@
 #!/bin/bash
 #Initialize complete contiv container. Start OVS and Net Plugin
 
+cstore="$CONTIV_ETCD"
+vtep_ip="$VTEP_IP"
+vlan_if="$VLAN_IF"
+
+set -euo pipefail
+
 reinit=false
 plugin="docker"
-vtep_ip="$VTEP_IP"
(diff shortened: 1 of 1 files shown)


scripts/netContain/contivNet.sh
```diff
@@ -1,30 +1,36 @@
...
+set -euo pipefail
+
 reinit=false
 plugin="docker"
-vtep_ip="$VTEP_IP"
-fwd_mode="bridge"
-cstore="$CONTIV_ETCD"
-cmode="bridge"
 netmaster=false
 netplugin=true
```
jojimt: Should this be initialized to false?

neelimamukiri: Default is that it is run as netplugin. Only if we choose to run as netmaster, with the -m parameter, I am setting it to false in which case we will only run as netmaster.

PR #781, opened 2017-03-13
Fix/Optimize state store APIs

yuva29:
1. Write/Read/ReadAll should handle `Key not found` explicitly

2. Optimze the code to iterate only when connection is refused/etcd cluster is unavailable

Signed-off-by: Yuva Shankar <yuva29@users.noreply.github.com>

### DIFF
--- a/state/consulstatedriver.go
+++ b/state/consulstatedriver.go
@@ -70,71 +70,88 @@ func processKey(inKey string) string {
 // Write state to key with value.
 func (d *ConsulStateDriver) Write(key string, value []byte) error {
 	key = processKey(key)
-	_, err := d.Client.KV().Put(&api.KVPair{Key: key, Value: value}, nil)
-	if err != nil && (api.IsServerError(err) || strings.Contains(err.Error(), "EOF") ||
-		strings.Contains(err.Error(), "connection refused")) {
-		for i := 0; i < maxConsulRetries; i++ {
-			_, err = d.Client.KV().Put(&api.KVPair{Key: key, Value: value}, nil)
-			if err == nil {
-				break
-			}
 
+	var err error
+
+	for i := 0; i < maxConsulRetries; i++ {
+		_, err = d.Client.KV().Put(&api.KVPair{Key: key, Value: value}, nil)
+		if err != nil && (api.IsServerError(err) || strings.Contains(err.Error(), "EOF") || strings.Contains(err.Error(), "connection refused")) {
 			// Retry after a delay
 			time.Sleep(time.Second)
+			continue
 		}
+
+		// when err == nil or anything other than connection refused
+		return err
 	}
 
+	// err could be anything other than connection refused
 	return err
 }
 
 // Read state from key.
 func (d *ConsulStateDriver) Read(key string) ([]byte, error) {
 	key = processKey(key)
-	kv, _, err := d.Client.KV().Get(key, nil)
-	if err != nil {
-		if api.IsServerError(err) || strings.Contains(err.Error(), "EOF") ||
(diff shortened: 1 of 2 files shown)


state/consulstatedriver.go
```diff
@@ -70,71 +70,80 @@ func processKey(inKey string) string {
...
+			if api.IsServerError(err) || strings.Contains(err.Error(), "EOF") || strings.Contains(err.Error(), "connection refused") {
 				time.Sleep(time.Second)
+			} else {
+				return []byte{}, err
 			}
-		} else {
-			return []byte{}, err
+		} else if kv == nil {
+			return []byte{}, core.Errorf("Key not found")
+		} else if err == nil {
```
jojimt: Check for err==nil here is redundant.

state/consulstatedriver.go
```diff
@@ -70,71 +70,80 @@ func processKey(inKey string) string {
...
+				time.Sleep(time.Second)
+				continue
+			} else {
+				return [][]byte{}, err
+			}
+		} else if kvs == nil {
+			// Consul returns success and a nil kv when a key is not found,
+			// translate it to 'Key not found' error
+			return nil, core.Errorf("Key not found")
+		} else if err == nil {
```
jojimt: Here as well.

state/etcdstatedriver.go
```diff
@@ -105,46 +103,49 @@ func (d *EtcdStateDriver) Read(key string) ([]byte, error) {
...
+			for _, node := range resp.Node.Nodes {
+				values = append(values, []byte(node.Value))
+			}
+			return values, nil
+		} else if client.IsKeyNotFound(err) {
+			return [][]byte{}, core.Errorf("Key not found")
+		} else if err.Error() == client.ErrClusterUnavailable.Error() {
+			// Retry after a delay
+			time.Sleep(time.Second)
+		}
```
jojimt: What happens if it's some other error?

state/etcdstatedriver.go
```diff
@@ -105,46 +103,49 @@ func (d *EtcdStateDriver) Read(key string) ([]byte, error) {
...
+
+	for i := 0; i < maxEtcdRetries; i++ {
+		resp, err = d.KeysAPI.Get(ctx, key, &client.GetOptions{Quorum: true})
+		if err == nil {
+			return []byte(resp.Node.Value), nil
+		} else if client.IsKeyNotFound(err) {
+			return []byte{}, core.Errorf("Key not found")
+		} else if err.Error() == client.ErrClusterUnavailable.Error() {
+			// Retry after a delay
+			time.Sleep(time.Second)
```
jojimt: What if it's some other error?

yuva29:
@jojimt I changed the logic to handle the conditions you pointed. PTAL

state/etcdstatedriver.go
jojimt: @yuva29 Any reason why you did not retain the original logic? Are you trying to anything beyond just placing the whole logic inside a single for loop?

yuva29: `Key not found` case wasn't handled earlier in `etcd` driver. Also, I have removed unnecessary `else` statements to align with go standards. Logic is pretty much the same except the way its coded.

jojimt: @yuva29 Please retain the check for resp.Node != nil as discussed. Also, please eliminate unnecessary else statements in the above if chain (if there is a return, there is no need for an else. Not sure why go vet did not catch it...).

yuva29:
@jojimt PTAL

PR #782, opened 2017-03-13
Fix ineffectual assignments

yuva29:
Fix errors reported by ineffassign 

Signed-off-by: Yuva Shankar <yuva29@users.noreply.github.com>

### DIFF
--- a/drivers/ovsdbDriver.go
+++ b/drivers/ovsdbDriver.go
@@ -217,7 +217,7 @@ func (d *OvsdbDriver) createDeleteBridge(bridgeName, failMode string, op oper) e
 	}
 
 	// simple insert/delete operation
-	brOp := libovsdb.Operation{}
+	var brOp libovsdb.Operation
 	if op == operCreateBridge {
 		bridge := make(map[string]interface{})
 		bridge["name"] = bridgeName
@@ -307,7 +307,7 @@ func (d *OvsdbDriver) CreatePort(intfName, intfType, id string, tag, burst int, 
 
 	// insert/delete a row in Interface table
 	idMap := make(map[string]string)
-	intfOp := libovsdb.Operation{}
+	var intfOp libovsdb.Operation
 	intf := make(map[string]interface{})
 	intf["name"] = intfName
 	intf["type"] = intfType
@@ -332,7 +332,7 @@ func (d *OvsdbDriver) CreatePort(intfName, intfType, id string, tag, burst int, 
 	}
 
 	// insert/delete a row in Port table
-	portOp := libovsdb.Operation{}
+	var portOp libovsdb.Operation
 	port := make(map[string]interface{})
 	port["name"] = intfName
 	if tag != 0 {
@@ -437,7 +437,7 @@ func (d *OvsdbDriver) CreatePortBond(intfList []string, bondName string) error {
 		intfUUIDList = append(intfUUIDList, intfUUID...)
 
 		// insert/delete a row in Interface table
-		intfOp := libovsdb.Operation{}
+		var intfOp libovsdb.Operation
 		iface := make(map[string]interface{})
 		iface["name"] = intf
 
@@ -452,7 +452,7 @@ func (d *OvsdbDriver) CreatePortBond(intfList []string, bondName string) error {
 	}
 
 	// Insert bond information in Port table
-	portOp := libovsdb.Operation{}
+	var portOp libovsdb.Operation
 	port := make(map[string]interface{})
 	port["name"] = bondName
 	port["vlan_mode"] = "trunk"
@@ -471,6 +471,9 @@ func (d *OvsdbDriver) CreatePortBond(intfList []string, bondName string) error {
 	lacpMap := make(map[string]string)
 	lacpMap["lacp-fallback-ab"] = "true"
 	port["other_config"], err = libovsdb.NewOvsMap(lacpMap)
+	if err != nil {
+		return err
+	}
 
 	portUUIDStr := bondName
 	portUUID := []libovsdb.UUID{{GoUuid: portUUIDStr}}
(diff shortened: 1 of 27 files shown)


test/systemtests/k8setup_test.go
```diff
@@ -364,7 +364,7 @@ func (k *kubernetes) tcFilterShow(bw string) error {
 	rate := strings.Split(output[1], "burst")
 	regex := regexp.MustCompile("[0-9]+")
 	outputStr := regex.FindAllString(rate[0], -1)
-	outputInt, err := strconv.ParseInt(outputStr[0], 10, 64)
+	outputInt, _ := strconv.ParseInt(outputStr[0], 10, 64)
```
dseevr: Same here

test/systemtests/swarm_test.go
```diff
@@ -336,7 +337,7 @@ func (w *swarm) tcFilterShow(bw string) error {
 	rate := strings.Split(output[1], "burst")
 	regex := regexp.MustCompile("[0-9]+")
 	outputStr := regex.FindAllString(rate[0], -1)
-	outputInt, err := strconv.ParseInt(outputStr[0], 10, 64)
+	outputInt, _ := strconv.ParseInt(outputStr[0], 10, 64)
```
dseevr: Same here

test/integration/utils_test.go
```diff
@@ -347,7 +347,7 @@ func tcFilterCheckBw(expBw, expBurst int64) error {
 	rate := strings.Split(output[1], "burst")
 	regex := regexp.MustCompile("[0-9]+")
 	outputStr := regex.FindAllString(rate[0], -1)
-	outputInt, err := strconv.ParseInt(outputStr[0], 10, 64)
+	outputInt, _ := strconv.ParseInt(outputStr[0], 10, 64)
```
dseevr: I would just omit the second assignment here like you did in docker_test.go

netctl/netctl.go
```diff
@@ -576,6 +576,10 @@ func inspectTenant(ctx *cli.Context) {
 	errCheck(ctx, err)
 
 	content, err := json.MarshalIndent(ten, "", "  ")
+	if err != nil {
+		log.Debugf("Failed to marshal indent %v", ten)
```
dseevr: You need to import logrus at the top here

unclejack:
@yuva29: The CI has failed. It seems to be related to the code changes made by the PR.

yuva29:
@unclejack Yeah. I'm debugging it. I'll push it again once i'm done.

netplugin/plugin/netplugin_test.go
```diff
@@ -195,28 +198,3 @@ func TestNetPluginInitInvalidConfigMissingNetworkDriverName(t *testing.T) {
 		t.Fatalf("plugin init succeeded, should have failed!")
 	}
 }
-
-func TestNetPluginInitInvalidConfigMissingFwdMode(t *testing.T) {
```
yuva29: Removing this test as its no longer valid. https://github.com/contiv/netplugin/blob/0f843fa92f923f6f37378c3956ae02ab3855350e/netplugin/plugin/netplugin.go#L299 there's always a fwd mode.

drivers/ovsdriver.go
```diff
@@ -843,6 +843,10 @@ func (d *OvsDriver) InspectNameserver() ([]byte, error) {
 	}
 
 	ns, err := d.nameServer.InspectState()
+	if err != nil {
+		log.Errorf("Error inspecting nameserver state: %v", err)
```
rchirakk: 'return' in case of error

drivers/ovsdbDriver.go
```diff
@@ -217,7 +217,7 @@ func (d *OvsdbDriver) createDeleteBridge(bridgeName, failMode string, op oper) e
 	}
 
 	// simple insert/delete operation
-	brOp := libovsdb.Operation{}
+	var brOp libovsdb.Operation
```
jojimt: If I understood the golang right, the above two statements have the same effect. Was this flagged by the tool? Just curious.

jojimt:
"Fix errors reported by ineffassign" @yuva29 is there a reference doc for this tool somewhere?

unclejack:
@jojimt: I don't know if there's a reference. However, you can install the tool using `go get github.com/gordonklaus/ineffassign` (I've mentioned this because some tools have a sub-package called cmd where the actual source for the installable binary lives) and take a look at the repository here: https://github.com/gordonklaus/ineffassign.

jojimt:
The reason I ask is some of the changes look questionable to me. It is easy for personal taste to creep in disguised as a tool and I would like to avoid that. E.g. many of the patterns addressed here have been seen extensively in other open source repos. I am not saying no here. But I would like to understand a bit better as to what we are doing here.

yuva29:
@jojimt yeah right. These are errors flagged by https://github.com/gordonklaus/ineffassign. Its the same ineffassign used in go report card here https://goreportcard.com/report/github.com/contiv/netplugin#ineffassign

dseevr:
FWIW, the actual Go source tends to prefer the `var`-style declaration when a variable should have its default value and when a struct is being initialized with no custom values

rchirakk:
I think the tools is NOT checking  'var' vs short declaration.
It checks if the initial value is being used or not.
for example
```for i:=0; i<10; i++ {}```
is not an flagged  as initial value i=0 is used inside the for loop
same for 
```
j:=0
for ;; j<10 {j++}
```

dseevr:
@yuva29 could you amend this PR so that `ineffassign` runs automatically like the other tools?  I think the `misspell` PR I added was the most recent one which you could use as a reference (use blame to find the commit if you need it)

Issue #779, opened 2017-03-12
Building contiv network container fails


gerritbinnenmars:
After git clone contiv/netplugin the following command fails:
[root@localhost netplugin]# scripts/netContain/BuildContainer.sh 
+ ARG1=none
+ '[' none == reinit ']'
+ sudo modprobe openvswitch
+ imgName=contiv-network-plugin
+ dockerBuildIt . contiv-network-plugin
++ docker build .
++ grep 'Successfully built'
++ cut -d ' ' -f 3
The command '/bin/sh -c make build' returned a non-zero code: 2
+ imgId=
+ [[ '' =~ [0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f] ]]
+ echo 'contiv-network-plugin Image was not built properly'
contiv-network-plugin Image was not built properly
+ return 255
+ '[' 255 '!=' 0 ']'
+ echo 'Failed building Contiv Image Bailing out Err 0'
Failed building Contiv Image Bailing out Err 0
+ exit

gerritbinnenmars:
Please, upgrade the Dockerfile to: FROM golang:1.7 to solve this issue

dseevr:
The step the build process is failing on doesn't have anything to do with Go:

```
Step 8/14 : COPY ./netplugin /contiv/bin/
lstat netplugin: no such file or directory
```

@unclejack do you know what's going on here?

unclejack:
@dseevr: Have you run `make run-build` first?

dseevr:
@unclejack I ran `./scripts/netContain/BuildContainer.sh` as documented [here](https://github.com/contiv/netplugin/tree/master/scripts/netContain#introduction) and this is the step the `docker build` command failed on

gerritbinnenmars:
Correct, the original error occurs in make run-build in the go install step in the subdirectory netplugin.
Below it the output when this step is done manually. See also: https://github.com/docker/engine-api/issues/385

root@5ddc7f2af0b2:/go/src/github.com/contiv/netplugin# GOGC=1500 go install -v ./netplugin
github.com/contiv/netplugin/vendor/github.com/docker/engine-api/client/transport
github.com/contiv/netplugin/version
# github.com/contiv/netplugin/vendor/github.com/docker/engine-api/client/transport
[...]

PR #785, opened 2017-03-15
refactor netContain

unclejack:
- improves the main Dockerfile
- refactors scripts/netContain
- fixes the formatting of the `contivd.sh` and `contivInit.sh` scripts

The commit which refactors scripts/netContain makes the following changes:
- adds the scripts/netContain/bin directory to .gitignore
- renames `BuildContainer.sh` to `build_image.sh` and `ReleaseContainer.sh` to `release_image.sh`
- rewrites the netContain Dockerfile to make the built image smaller and to reduce the number of layers
- cleans up `release_image.sh`
- moves `contivNet.sh` and `ovsInit.sh` to `scripts/netContain/scripts`
- makes some small changes to `contivNet.sh` and `ovsInit.sh`
- cleans up README.md


scripts/netContain/Dockerfile
```diff
@@ -17,30 +17,18 @@
 
 FROM ubuntu:16.04
```
gaurav-dalvi: Though this is not part of your change , Any specific reason on why are we using ubuntu image here ? It actually fattens up our final container image.

unclejack: This is something we can discuss later. I don't plan to convert this Dockerfile to something else in this PR.

neelimamukiri: @unclejack, not related to this change, but when you are looking to change the Dockerfile, can you please take a look at https://github.com/contiv/netplugin/blob/master/install/v2plugin/Dockerfile and keep things similar if possible.

scripts/netContain/README.md
```diff
@@ -1,42 +1,62 @@
...
 ## Executing Containerized Contiv
 
-   Running as Net Master
+   Running netmaster
 ```
-	docker run -itd --pid=host --net=host --name=contivNet  --privileged   -v /etc/openvswitch:/etc/openvswitch -v /etc/kubernetes/ssl:/tmp/kubernets/ssl -v /var/run/:/var/run/  -v /var/contiv:/var/contiv rajenata/contiv:0.1  -m  -p kubernetes -c etcd://127.0.0.1:2379
+  docker run -itd --pid=host --net=host --name=contivNet \
+--privileged -v /etc/openvswitch:/etc/openvswitch \
+-v /etc/kubernetes/ssl:/tmp/kubernets/ssl -v /var/run/:/var/run/ \
+-v /var/contiv:/var/contiv rajenata/contiv:0.1  -m  -p kubernetes \
```
dseevr: `rajenata/contiv:0.1`?  Since we're removing this person as MAINTAINER, we should probably use a different image name here and the other spots in this file

scripts/netContain/README.md
```diff
@@ -1,42 +1,62 @@
...
+  docker run -itd --pid=host --net=host --name=contivNet --privileged \
+-v /etc/openvswitch:/etc/openvswitch -v /var/run:/var/run \
+-v /etc/kubernetes/ssl:/tmp/kubernetes/ssl -v /var/contiv:/var/contiv \
+rajenata/contiv:0.1 -v ${WORKER_IP} -i  ${VLAN_IF} -f routing \
+-p kubernetes -c etcd://127.0.0.1:2379
 ```
-  The --pid=host option is used in the case of Kubernetes environment so that the netplugin can exec nsenter and modify the container networking.
- 
-  It is mandatory to provide either vtep-ip or vlan interface when using as netplugin.
+  The --pid=host option is used for Kuerbenetes to allow netplugin to
```
dseevr: Typo here

scripts/netContain/build_image.sh
```diff
@@ -0,0 +1,66 @@
...
+fi
+
+docker run --name=$TEMPORARY_CONTAINER_NAME $BUILD_ENV_IMAGE_NAME || true
+
+echo "Copying the Contiv binaries..."
+for f in $NETPLUGIN_BINARIES; do
+	docker cp $TEMPORARY_CONTAINER_NAME:/go/bin/$f scripts/netContain/bin/
+done
+
+docker rm -f $TEMPORARY_CONTAINER_NAME
```
dseevr: This can use `-fv` to also delete the volume

unclejack:
@dseevr: Thanks. I've made the changes. PTAL

jojimt:
@unclejack It is not clear how this addresses https://github.com/contiv/netplugin/issues/779. Could you please elaborate?

Dockerfile
```diff
@@ -26,15 +26,16 @@ FROM golang:1.7.5
 # a proxy.
 #ENV http_proxy ""
 #ENV https_proxy ""
-ENV GOPATH /go/
+ARG http_proxy
```
unclejack: This makes it possible to use an HTTP and HTTPS proxy.

Dockerfile
```diff
@@ -26,15 +26,16 @@ FROM golang:1.7.5
...
+ARG https_proxy
+ENV GOPATH=/go/ NET_CONTAINER_BUILD=1
```
unclejack: We can use a single layer for two variables.

scripts/netContain/Dockerfile
```diff
@@ -17,30 +17,18 @@
...
-RUN mkdir -p /contiv/bin
-RUN mkdir -p /contiv/scripts
-
-COPY ./netplugin /contiv/bin/
-COPY ./netmaster /contiv/bin/
-COPY ./contivk8s /contiv/bin/
-COPY ./netctl /contiv/bin/
-COPY ./ovsInit.sh /contiv/scripts/
-COPY ./contivNet.sh /contiv/scripts/
+COPY ./bin /contiv/bin/
```
unclejack: We don't need to copy every single file which is stored in the main netContain. We can just copy this directory instead. We're also able to remove that directory to clean up without removing every file by hand.

scripts/netContain/README.md
```diff
@@ -1,42 +1,62 @@
-## Introduction
```
unclejack: Removes redundant whitespace. This pollutes diffs when added/removed. It's better to keep an eye on it and remove it.

scripts/netContain/README.md
```diff
@@ -1,42 +1,62 @@
-## Introduction 
+## Introduction
 
-This Directory contains scripts required to containerize Contiv and deploy it in either docker or kubernetes modes.
-The script BuildContainer.sh should be executed from the top level directory as follows 
+This directory contains scripts required to containerize Contiv and
```
unclejack: Rewrites the text to avoid some longer lines and to make some corrections.

scripts/netContain/ReleaseContainer.sh
```diff
@@ -1,83 +0,0 @@
-contiv_version=""
```
unclejack: This script is just moved over to `release_image.sh` and formatted. I know marketing material refers to Docker images as containers, but a container is instantiated from an image, similar to how an object is an instance of a class in some programming languages.

scripts/netContain/build_image.sh
```diff
@@ -0,0 +1,66 @@
...
+FINAL_IMAGE_NAME="contivbase"
+
+NETPLUGIN_BINARIES="netplugin netmaster netctl contivk8s"
+
+function get_image_id() {
+	docker inspect --format '{{.ID}}' $1
+	return $?
+}
+
+function build_image_from_path() {
```
unclejack: This function has been rewritten. It was ignoring the best practices of using `docker build`. `docker build` accepts a `-t` parameter to tag an image. There's no need to have complex code to figure out if the build process has an image ID in the output or if it's some kind of error.

The exit codes of the commands was ignored before. This wasn't particularly ideal.

scripts/netContain/build_image.sh
```diff
@@ -0,0 +1,66 @@
...
+	fi
+
+	echo "Image $2 has been built"
+	return 0
+}
+
+set -x
+set -e
+
+rm -rf scripts/netContain/bin
```
unclejack: Clean up the temporary bin directory. This directory is used to make it easier to remove the binaries and to ensure that we don't leave binaries around for no good reason.

scripts/netContain/build_image.sh
```diff
@@ -0,0 +1,66 @@
...
+mkdir scripts/netContain/bin
+
+build_image_from_path . $BUILD_ENV_IMAGE_NAME
```
unclejack: This builds the Docker image which compiles the netplugin, netctl, netmaster and contivk8s binaries as part of the build process.

scripts/netContain/build_image.sh
```diff
@@ -0,0 +1,66 @@
...
+if [ "$?" -ne 0 ]; then
+	echo "Failed to build the "
+	exit 1
+fi
+
+docker run --name=$TEMPORARY_CONTAINER_NAME $BUILD_ENV_IMAGE_NAME || true
```
unclejack: This creates a dummy container to be able to copy the built binaries from it.

scripts/netContain/build_image.sh
```diff
@@ -0,0 +1,66 @@
...
+
+echo "Copying the Contiv binaries..."
+for f in $NETPLUGIN_BINARIES; do
```
unclejack: Makes it easier to maintain this script. There's no need to duplicate this to copy each file individually.

scripts/netContain/build_image.sh
```diff
@@ -0,0 +1,66 @@
...
+	docker cp $TEMPORARY_CONTAINER_NAME:/go/bin/$f scripts/netContain/bin/
+done
+
+docker rm -fv $TEMPORARY_CONTAINER_NAME
```
unclejack: The dummy container is no longer useful. We've copied the files. We delete the dummy container.

scripts/netContain/build_image.sh
```diff
@@ -0,0 +1,66 @@
...
+
+echo "Building the final Docker image..."
+build_image_from_path scripts/netContain contivbase
```
unclejack: This builds the final netContain image with the binaries we've just built.

scripts/netContain/build_image.sh
```diff
@@ -0,0 +1,66 @@
...
+if [ "$?" -ne 0 ]; then
+	echo "Failed to build the final Docker image"
+	exit 1
+fi
+
+rm -rf scripts/netContain/bin
```
unclejack: Clean up after ourselves. There's no point in leaving this around.

scripts/netContain/contivInit.sh
```diff
@@ -6,19 +6,16 @@ set -x
```
unclejack: This script is changed to format it properly. No other changes are made.

scripts/netContain/contivd.sh
```diff
@@ -1,79 +1,76 @@
 #!/usr/bin/bash
```
unclejack: This script is only modified to format the script properly, to remove trailing whitespace and follow best practices.

jojimt:
These annotations are definitely good :) But, can you help by answering my question which was: *It is not clear how this addresses #779. Could you please elaborate?* Possible answers could be A) No it does not address that issue B) The issue was caused by xyz and has been addressed by changes in lines xxx. I was hoping to see B.

scripts/netContain/scripts/contivNet.sh
```diff
@@ -0,0 +1,137 @@
+#!/bin/bash
```
unclejack: This script has some minor changes made to it from what I recall. Git refused to provide a diff for it, no matter how much effort I've put into making it do so (e.g. rename + changes).

scripts/netContain/scripts/ovsInit.sh
```diff
@@ -1,25 +1,26 @@
 #!/bin/bash
-#Start OVS in the Contiv Container
```
unclejack: This script is formatted. I've also made some minor adjustments.

.gitignore
```diff
@@ -46,6 +46,9 @@ cscope*
 install/v2plugin/rootfs
 install/v2plugin/config.json
 
+# netcontain temporary build files
+scripts/netContain/bin
```
unclejack: Adds the bin directory to .gitignore. We really don't need to commit netplugin binaries to the repository. This was an issue before.

scripts/netContain/Dockerfile
```diff
@@ -17,30 +17,18 @@
 
 FROM ubuntu:16.04
 
-MAINTAINER Rajesh Nataraja <rajenata@cisco.com>
```
unclejack: This guy is no longer maintaining this Dockerfile.

scripts/netContain/Dockerfile
```diff
@@ -17,30 +17,18 @@
...
-
 # Make sure to Modify the Proxy Server values if required 
 # ENV export http_proxy=http://proxy.localhost.com:8080
 # ENV export https_proxy=https://proxy.localhost.com:8080
 
+RUN apt-get update \
```
unclejack: This produces a single layer smaller layer and the Dockerfile is cleaner. The apt lists clutter up the resulting image for no good reason without any benefit.

The official Docker Dockerfile from the docker/docker takes a similar approach to package installation and command formatting. The Docker docs for Dockerfile best practices recommend the same thing.

scripts/netContain/build_image.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
```
unclejack: This entire script is now much safer to run. It doesn't clutter the Docker daemon with images and it also doesn't rebuild the main netplugin image from scratch every single time. The advantage is that this wastes less time, less bandwidth and it's much safer overall.

The risk of running into build errors is also greatly diminished. I've tested this build script thoroughly.

unclejack:
@jojimt: The script responsible for building those images wasn't safe. It wasn't cleaning up after itself. I was running into failures as well. It had some side effects (leaving binaries around, not taking advantage of the `docker build` cache properly, not having temporary files in .gitignore, mixing up runtime steps such as `modprobe` with the build process, it had comments left around, it wasn't using `docker build -t some_image_name`, it was using some hack to parse the Docker image ID and many more issues).

I was also running into some errors for repeated builds. I had to make changes as I was going through the script to address those issues. As part of the cleanup, I've made some minor adjustments in other places as well. All of this is meant to leave the scripts and the code in a better state than before.

I've continued to make adjustments until I could build these images repeatedly without errors. All of this is meant to make it less likely to break in the future. If it breaks, it'll be easier to understand where to start to fix the issues.

It might be difficult to understand why I've made the changes to the scripts (not the cosmetic formatting changes), even though those changes and what they do is well understood. The process of testing and making changes until everything works without side effects is what has lead to those changes.

This is why this PR fixes #779: it fixes the initial and repeated builds done by executing `./scripts/netContain/build_image.sh`.

Please let me know if you wish any other changes to be made or to discuss anything else related to these changes.

jojimt:
I was hoping to see a description of the root cause and how the changes address that root cause, something beyond "I have built it multiple times". As these changes themselves don't appear to introduce any breakage, I will merge this. But, whether it addresses https://github.com/contiv/netplugin/issues/779 will now need to be answered by the submitter of the issue.

unclejack:
@jojimt: I've worked for about 4 hours to make these changes and to test them properrly. I didn't keep a log of all the encountered errors.

PR #788, opened 2017-03-16
Update dockerclient to use the official docker client

DivyaVavili:
- Use docker/docker/client
- Remove references to samalba/dockerclient
- Update references of docker/engine-api/... to docker/docker/...


.gitignore
```diff
@@ -52,4 +52,7 @@ scripts/netContain/bin
 # version
 version/version_gen.go
 
+# test files
+test/systemtests/cfg.json
```
gkvijay: is this file being removed from git as well

DivyaVavili: Not removed from git. I think it's better to have one sample cfg.json for reference

DivyaVavili:
@jojimt: I've changed the way the docker events channel is closed. PTAL.

PR #793, opened 2017-03-18
include 'latest' tag in the image

rchirakk:


unclejack:
I've tested these changes.

dseevr:
Is this just to aid in local development/testing?

If at some point in the future we have to go back to a past release and do a point release to fix a bug, this could mean that your local ":latest" image is not even close to the latest code that has been officially released.

PR #795, opened 2017-03-20
Makefile: clean up & use targets where possible

unclejack:
This PR cleans up the Makefile a bit. It makes the following changes:
- it uses the existing targets where possible
`make foo && make bar` is the same as `make foo bar`
- `export VARNAME=VARVALUE \ &&` is used instead instead of `VARNAME=VARVALUE command1, VARNAME=VARVALUE command2` to avoid needless repetition 
- the same change is made for targets where we need to change the directory for multiple commands

The changes are kept to a minimum for this PR to make it easier to review and to keep track of the changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,10 +96,7 @@ install-shell-completion:
 	sudo cp scripts/contrib/completion/bash/netctl /etc/bash_completion.d/netctl
 endif
 
-build:
-	make start
-	make ssh-build
-	make stop
+build: start ssh-build stop
 
 clean: deps
 	rm -rf $(GOPATH)/pkg/*/github.com/contiv/netplugin/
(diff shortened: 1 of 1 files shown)


unclejack:
The CI status is green. PTAL

jojimt:
@abhinandanpb @rchirakk Please review.

Makefile
```diff
@@ -90,10 +90,7 @@ install-shell-completion:
 	sudo cp scripts/contrib/completion/bash/netctl /etc/bash_completion.d/netctl
 endif
 
-build:
-	make start
-	make ssh-build
-	make stop
+build: start ssh-build stop
```
rchirakk: is n't this going to break parallel builds ? -jX

unclejack: @rchirakk -jX shouldn't be used at all in the context of this repository. It doesn't make sense.

unclejack: We're already using something similar in the following places:
https://github.com/contiv/netplugin/blob/master/Makefile#L33
https://github.com/contiv/netplugin/blob/master/Makefile#L81
https://github.com/contiv/netplugin/blob/master/Makefile#L188
https://github.com/contiv/netplugin/blob/master/Makefile#L227

We would've run into this already. Please let me know if you run into any issues, with or without this PR.

dseevr:
@unclejack this needs a rebase

unclejack:
@dseevr Thanks! I've rebased the PR.

unclejack:
This PR is green after the rebase. PTAL

dseevr:
@abhinandanpb @rchirakk ping

rhim:
@unclejack fix conflicts, then merge

unclejack:
I've fixed the conflicts and the tests pass. I'll merge this now.

PR #798, opened 2017-03-20
VPP integration

brecode:
This patch provides integration for VPP as the data plane:
Currently supporting single host integration with:

1. Create/Delete Network
2. Create/Delete Endpoints
3. Add/Delete ACL rules


dseevr:
Hi @brecode,

CI is failing because of missing dependencies.  It looks like you need to run:

```
make godep-save
git add Godeps
git add vendor
git commit -m "update godeps"
```

and push the result to trigger a new build.  Thanks!

brecode:
Hi @dseevr, I also think that there has to be a separate branch for VPP as I am changing some default values (OVS-VPP constants) and the build process would require installation of VPP. I'll add a README file as well but let me know if we can have a separate branch for that. 

Thanks

jojimt:
@brecode We generally don't like to create separate branches. Until a feature is ready, we usually leave it in the private fork and raise PR only when a functional block is ready for at least alpha level use.

unclejack:
`vendor/github.com/fdio-stack/go-vpp/.DS_Store ` shouldn't be here. 
https://github.com/contiv/netplugin/pull/798/commits/fd379599dc4f1a9535b9c7340950345a497107e6#diff-2e056dab6b6b256952ffad82bc38c9c5

rhim:
@brecode Adding to what @jojimt said earlier - this is a feature for which I would like to see a complete "switch" - one that can completely swap in/out VPP for OVS.

dseevr:
@brecode It looks like you're running an old version of godep... your last commit actually deleted _all_ the dependencies Godeps.json

Are you running an older version of Go?  We're mostly running 1.7.5 here.  I would upgrade to 1.7.5 and run the commands again

unclejack:
There are many binaries committed on this branch. All the binaries must go away. The `.DS_Store` files need to be removed as well.

The build process which installs the dependencies (compiling required libraries) needs to be in a script put in `scripts/`.

Build tags need to be added to specify whether VPP support is built or not. Since this would be an experimental feature for the time being, this is highly desirable. This must not impact the OVS based functionality.

Please comment if you'd like us to help.

PR #799, opened 2017-03-20
.gitignore: ignore OS generated files

unclejack:
This PR adds some OS specific files to .gitignore. This will make it easier for other contributors and for everyone to avoid adding these files by accident.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,15 @@ install/v2plugin/config.json
 # netcontain temporary build files
 scripts/netContain/bin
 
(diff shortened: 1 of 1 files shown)


dseevr:
Looks good.  Could you also do a `find . -name "..." | xargs git rm -f` to find and remove any that are currently checked in?

(I've had to do this more times than I'd care to admit 🙄 )

unclejack:
@jojimt: Can you take a look at this, please? It's green.

PR #807, opened 2017-03-28
Support host access for vxlan routing mode

jojimt:
Signed-off-by: Joji Mekkatt <joji@cisco.com>

This PR adds support for host access for vxlan networks when fwd-mode is configured as routing. Host access relies on an "infra" network being created with name *contivh1* and a default gateway specified.
E.g. *netctl net create -n infra -s 132.1.1.0/24 -g 132.1.1.1 contivh1* 
This will make all vxlan networks accessible locally from the host.

In addition a host NAT interface is implicitly created which will allow NAT access to outside networks masqueraded behind the host interface. This also enables service endpoints external to the cluster to be access via the service proxy.

This PR eliminates the need for the hostBridge.


jojimt:
@DivyaVavili @rchirakk PTAL

netplugin/agent/state_event.go
```diff
@@ -79,6 +138,11 @@ func processInfraNwCreate(netPlugin *plugin.NetPlugin, nwCfg *mastercfg.CfgNetwo
 		return err
 	}
 
+	// add host access routes for vxlan networks
+	if nwCfg.NetworkName == contivVxGWName {
```
rchirakk: is there a dependency on tenant here ?

drivers/ovsSwitch.go
```diff
@@ -53,7 +54,31 @@ type OvsSwitch struct {
...
+// getPvtIP returns a private IP for the port
+func (sw *OvsSwitch) getPvtIP(portName string) net.IP {
+	ipStr := ""
+	if strings.Contains(portName, "vport") {
+		port := strings.Replace(portName, "vvport", "", 1)
+		portNum, err := strconv.Atoi(port)
+		if err == nil {
+			log.Infof("portname: %s portnum: %d", portName, portNum)
+			ipStr, _ = netutils.PortToHostIPMAC(portNum, sw.hostPvtNW)
+			log.Infof("PvtIP: %s", ipStr)
```
jojimt: Good point. Will change it.

drivers/ovsdriver.go
```diff
@@ -196,19 +196,12 @@ func (d *OvsDriver) Init(info *core.InstanceInfo) error {
...
-		info.FwdMode, nil)
-	if err != nil {
-		log.Fatalf("Error creating host switch. Err: %v", err)
-	}
-
 	if maxPortNum > 0xfffe {
 		log.Fatalf("Host bridge logic assumes maxPortNum <= 0xfffe")
 	}
 
 	// Add host port.
```
jojimt: We don't need to as the datapath handles it.

netplugin/agent/state_event.go
jojimt: Basically, this can be used only with a single tenant for now. Host access will be available to the tenant where this interface belongs.

drivers/ovsSwitch.go
DivyaVavili: Same here. Please remove if not required.

drivers/ovsSwitch.go
```diff
@@ -53,7 +54,31 @@ type OvsSwitch struct {
...
+}
+
+// getPvtIP returns a private IP for the port
+func (sw *OvsSwitch) getPvtIP(portName string) net.IP {
+	ipStr := ""
+	if strings.Contains(portName, "vport") {
+		port := strings.Replace(portName, "vvport", "", 1)
+		portNum, err := strconv.Atoi(port)
+		if err == nil {
+			log.Infof("portname: %s portnum: %d", portName, portNum)
```
jojimt: Will change to debugf.

drivers/ovsSwitch.go
```diff
@@ -801,28 +818,17 @@ func (sw *OvsSwitch) AddHostPort(intfName string, intfNum, network int, isHostNS
...
-
-		// Create a Veth pair
-		err := createVethPair(intfName, ovsPortName)
-		// Set the OVS side of the port as up
-		err = setLinkUp(ovsPortName)
-		if err != nil {
-			log.Errorf("Error setting link %s up. Err: %v", ovsPortName, err)
-			return "", err
-		}
+		log.Infof("Host port in container name space -- ignore")
```
jojimt: Fair point. But, we are not ready to completely get rid of it as yet.

rhim:
@jojimt could you open an issue for installer work with specifics?

jojimt:
Updated https://github.com/contiv/install/issues/82

rhim:
@jojimt where does the IP address range for the infra network come from?

jojimt:
It is user-configurable. For installer, we can pick something like 172.18.0.0/16 perhaps as default. But, there should be a way for the user to over-ride.

PR #808, opened 2017-03-28
test/systemtests: rename aci_util.go to add _test

unclejack:
This PR renames `aci_util.go` to `aci_util_test.go`. This change is needed because this code is tied to code from other files with `_test` in their names. This means this code can only compile when running the tests and it would throw an error in some cases. In order to avoid any potential issues, it's renamed to have the _test suffix as the others.


unclejack:
This PR's status is green now. PTAL

PR #811, opened 2017-03-30
version/CURRENT_VERSION: bump to 1.0.0-beta.5

unclejack:


jojimt:
@unclejack Can we avoid having to make a commit each time we need to make a release?

PR #814, opened 2017-04-04
Updating Godeps to experimental L3 changes and test fixes

abhinandanpb:
1) Container godep update to latest ofnet that contains changes to support L3 without bgp for cloud use case and bgp cleanups
2) L3 test fix for intermittent failures (Need to be revisited after observation).


jojimt:
LGTM

PR #815, opened 2017-04-05
Issue 813: Add support for default-group in ACI mode

srampal:
@joji PTAL

### DIFF
--- a/netmaster/master/endpoint.go
+++ b/netmaster/master/endpoint.go
@@ -118,6 +118,14 @@ func CreateEndpoint(stateDriver core.StateDriver, nwCfg *mastercfg.CfgNetworkSta
 	epCfg.ServiceName = ep.ServiceName
 	epCfg.EPCommonName = epReq.EPCommonName
 
+	// In ACI mode, if a pod does not have a group label, we will assume "default-group"
+	isAci, _ := IsAciConfigured()
+
+	if isAci && (len(epCfg.ServiceName) == 0) {
+		epCfg.ServiceName = "default-group"
+		log.Infof("Over-riding null group with default-group for ep %s nw %s", epCfg.EndpointID, epCfg.NetID)
+	}
+
 	if len(epCfg.ServiceName) > 0 {
 		epgCfg = &mastercfg.EndpointGroupState{}
 		epgCfg.StateDriver = stateDriver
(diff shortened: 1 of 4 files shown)


jojimt:
As discussed, please add a check for default-net and add the new test to Makefile.

srampal:
@Joji  Added the test to Makefile … however had to retain the logic of not depending on default-net so that the system tests (runContainer function in docker_test.go and so on) would not need to be refactored to become ACI aware. We can discuss more details offline if needed.

Re-submited PR with latest changes PTAL

test/systemtests/docker_test.go
```diff
@@ -67,6 +67,12 @@ func (d *docker) runContainer(spec containerSpec) (*container, error) {
 		netstr = "--net=" + netstr
 	}
 
+        // Handle the case of "default-group" not specified explicitly 
+        // but only for the aci case (detected here via special aci network name)
+        if (spec.serviceName == "") && (spec.networkName == "default-net-aci"){
+                netstr = "--net=default-group"
```
jojimt: Why do we have this? This defeats the purpose of the test. The whole point is that in ACI mode, the code you added takes care of this automatically.

test/systemtests/aci_test.go
```diff
@@ -87,6 +87,94 @@ func (s *systemtestSuite) TestACIMode(c *C) {
...
+        containersDefault := append( containersA, containersC...)
+
+	// Verify containers within the combined default-group can ping each other 
+	c.Assert(s.pingTest(containersDefault), IsNil)
+	// Verify containers within epgb can ping each other
+	c.Assert(s.pingTest(containersB), IsNil)
+	// Verify containers within the combined default-group can't ping epgb containers
+	c.Assert(s.pingFailureTest(containersDefault, containersB), IsNil)
+
+	log.Infof("Triggering netplugin restart")
```
jojimt: There is no need to do netplugin restart test specifically for this. That's already covered separately. Please remove it from here.

srampal:
@joji  made the changes you suggested  PTAL
Removed the netplugin restart case
Moved the system test case for this to k8s (make k8s-test)

jojimt:
@srampal Please make sure tests pass.

srampal:
Done @joji  PTAL

netmaster/master/endpoint.go
```diff
@@ -118,6 +118,13 @@ func CreateEndpoint(stateDriver core.StateDriver, nwCfg *mastercfg.CfgNetworkSta
 	epCfg.ServiceName = ep.ServiceName
 	epCfg.EPCommonName = epReq.EPCommonName
 
+	// In ACI mode, if a pod does not have a group label, we will assume "default-group"
+	isAci, _ := IsAciConfigured()
+
+	if isAci && (len(epCfg.ServiceName) == 0) {
+		epCfg.ServiceName = "default-group"
```
jojimt: Please add a log.Infof here with a message we are overriding withe default-group.

PR #819, opened 2017-04-07
modify version to not rely on code gen

### LINKED ISSUE #702: Commit/version/build time vars should be set by the linker rather than relying on codegen
Right now, the `version` package can't be compiled without running a script which outputs another .go file into its directory.  This setup prevents this package from being re-used by other projects (I hit this recently).

Rather than relying on a script to generate a .go file at build time (see [this commit](https://github.com/contiv/netplugin/blob/c891b87f51f75c6c3cad0bde6bb92b0dab38b2fb/version/generate_version#L31-L44)), we can have all the variables in our code and simply set them using the `-X` linker flag (e.g., [here](https://github.com/contiv/volplugin/blob/0453e29b4d9916ae35ae492b9f3f302ff72bcc58/Makefile#L123) and [here](https://github.com/contiv/volplugin/blob/0453e29b4d9916ae35ae492b9f3f302ff72bcc58/volsupervisor/volsupervisor/main.go#L13-L14)).  This will let other projects use the struct in here rather than rolling their own.

unclejack:
This PR makes the required changes to remove the use of code generation from the version package.

The goal of this PR is to make this code usable via `go get`. Other Go based projects have a similar approach for their builds.

Fixes #702


unclejack:
This PR is important. It's part of the changes which will help us remove the need to bump the build in `version/CURRENT_VERSION`.

jojimt:
How about an issue to track what we are addressing here, as well as some unit test code?

unclejack:
The issue which is addressed through this PR is #702.

unclejack:
This code can be tested quite easily by running `make tar` and `make run-build`. This code has also been executed on the CI when the binaries were built to run the tests.

Please let me know if there's anything else you'd like to see tested.

PR #820, opened 2017-04-10
add BUILD_VERSION to override the version

unclejack:
This PR makes the necessary changes to be able to avoid bumping the version for every single alpha/beta/rc release. The BUILD_VERSION environment variable is added to maintain consistency with the other components of Contiv.

version/CURRENT_VERSION is still used. However, it will be relevant for actual releases (1.0.0, 1.0.1, 1.1.0, 1.1.1). The version defined in the file will also be used for the automated releases as it has been used before.

This change is relevant mostly to alpha/beta/rc releases which require a build number without a build version bump commit.

These changes help with the request made in this comment: https://github.com/contiv/netplugin/pull/811#issuecomment-290426185

This PR is so small because it builds on top of #819.


Makefile
```diff
@@ -80,7 +80,7 @@ checks: go-version gofmt-src golint-src govet-src misspell-src
 
 run-build: deps checks clean
 	cd $(GOPATH)/src/github.com/contiv/netplugin && \
-	USE_RELEASE=${USE_RELEASE} \
+	USE_RELEASE=${USE_RELEASE} BUILD_VERSION=${BUILD_VERSION} \
 	TO_BUILD="${TO_BUILD}" VERSION_FILE=${VERSION_FILE} \
```
dseevr: Do we still need `VERSION_FILE` here and the write to `$VERSION_FILE` in build.sh?

dseevr:
Related to the previous commit:

```
TO_BUILD="./netplugin/ ./netmaster/ ./netctl/netctl/ ./mgmtfn/k8splugin/contivk8s/ ./mgmtfn/mesosplugin/netcontiv/" VERSION_FILE=/tmp/netplugin-version \
scripts/build.sh
tr: warning: an unescaped backslash at end of string is not portable
```

I think it's complaining about line 4 of build.sh

dseevr:
Could you also remove `version/version_gen.go` from `.gitignore`?

If people have previously run a `make demo` or a build in their code dir and run it again after the previous PR was merged, they will get a `gofmt` error because one of the old codegen files is still hanging around.  With it being removed from `.gitignore`, their `git status` will show that the file is untracked and it will be obvious what happened.

Makefile
unclejack: @dseevr Yes, do you happen to have a suggestion on how to change that to not need it any more?

.gitignore
```diff
@@ -58,9 +58,6 @@ scripts/netContain/bin
 ehthumbs.db
 Thumbs.db
 
-# version
-version/version_gen.go
```
unclejack: @dseevr: I've removed this from the .gitignore.

Makefile
dseevr: Oh, I see, it's not directly used but it's used by the `release` target later

unclejack:
This PR is green. PTAL

unclejack:
@dseevr: I've addressed all the issues.

This PR is green and ready for review. PTAL

dseevr:
LGTM

scripts/build.sh
```diff
@@ -1,9 +1,13 @@
 #!/bin/bash
 
 BUILD_TIME=$(date -u +%m-%d-%Y.%H-%M-%S.UTC)
-VERSION=$(cat version/CURRENT_VERSION | tr -d \\)
+VERSION=$(cat version/CURRENT_VERSION | tr -d '\n')
 PKG_NAME=github.com/contiv/netplugin/version
 
+if [ -n "$BUILD_VERSION" ]; then
```
rhim: I find this a bit confusing to follow this flow. VERSION is defined first, then it is replaced by BUILD_VERSION, then BUILD_VERSION is defined again based on VERSION. If nothing else, it requires a bit of commenting to clarify what's defined and used.

version/CURRENT_VERSION
```diff
@@ -1 +1 @@
-1.0.0-beta.6
+1.0.0-beta
```
rhim: also nice to define how this version will change

unclejack: This change will affect the versioning as follows:
- alpha/beta/etc releases made automatically use whatever is defined here together with a timestamp suffix
- releases made by hand won't require bumping the build for beta/rc/etc releases made by hand
- regular releases will have the full version in this file (1.0.0, 1.1.0, 1.0.2)

This makes it a lot easier to release beta builds.

unclejack:
@rhim: I've added some documentation related to BUILD_VERSION. PTAL

unclejack:
This is green again after #826 got merged. I'll go ahead and merge now.

PR #826, opened 2017-04-12
scripts/netContain/Dockerfile: install ovs 2.5.x

unclejack:
This PR makes the netplugin Docker image use any OVS 2.5.x version found in the Ubuntu repositories. Previously released netplugin images aren't affected by this change.

It makes sense to install any 2.5.x version available. No new features are introduced as part of these maintenance releases. The bug fixes will be good to have.

This fixes the build of this image and it addresses the issue we have with the broken build on the CI.

### DIFF
--- a/scripts/netContain/Dockerfile
+++ b/scripts/netContain/Dockerfile
@@ -22,7 +22,7 @@ FROM ubuntu:16.04
 # ENV export https_proxy=https://proxy.localhost.com:8080
 
 RUN apt-get update \
- && apt-get install -y openvswitch-switch=2.5.0-0ubuntu1 \
+ && apt-get install -y openvswitch-switch=2.5.2-0ubuntu0.16.04.1 \
         net-tools \
         iptables \
  && rm -rf /var/lib/apt/lists/*


scripts/netContain/Dockerfile
```diff
@@ -22,7 +22,7 @@ FROM ubuntu:16.04
 # ENV export https_proxy=https://proxy.localhost.com:8080
 
 RUN apt-get update \
- && apt-get install -y openvswitch-switch=2.5.0-0ubuntu1 \
+ && apt-get install -y openvswitch-switch=2.5* \
```
gaurav-dalvi: 1: Do we need to change version in our ansible repo as well ? 
2: As we have not started supporting ubuntu officially, do we need to add this specific to ubuntu ? 
3: Some time back, we had issues with ovs repo, so we changed to openstack kilo repo to install ovs.
https://github.com/contiv/ansible/blob/master/roles/contiv_network/tasks/ovs.yml#L3
Do we need to follow same for ubuntu as well ?

dseevr: Should this be `2.5.*`?  I don't think a `2.55.x` release will come out any time soon, but still :joy:

unclejack: 1. Yes, we need to change this in ansible as well.
2. This is a Dockerfile for a Docker image. We use this for our netplugin Docker images.
3. I don't think that's the case for now. This OVS version seems fine for now. We can revisit this at any point in the future.

unclejack: @dseevr: I've made the change.

dseevr:
For posterity, this is the CI error we saw:

```
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 openvswitch-switch : Depends: openvswitch-common (= 2.5.0-0ubuntu1) but 2.5.2-0ubuntu0.16.04.1 is to be installed
[91mE: Unable to correct problems, you have held broken packages.
[0mRemoving intermediate container fac27c759bd8
The command '/bin/sh -c apt-get update  && apt-get install -y openvswitch-switch=2.5.0-0ubuntu1         net-tools         iptables  && rm -rf /var/lib/apt/lists/*' returned a non-zero code: 100
make: *** [host-build-docker-image] Error 100
make: *** [all-CI] Error 2
```

unclejack:
The current CI run is here: https://contiv-ci.ngrok.io/job/Netplugin%20Pull-Request%20Build%20Master/1851/

gkvijay:
can you fix the release number (say 2.5.2-0ubuntu0.16.04.1) until we add support for system-tests

unclejack:
I've made the change as desired.

PR #830, opened 2017-04-13
fix release

unclejack:
This PR makes a few changes to fix the release. The BUILD_VERSION which overrides the version wasn't being passed through to the VM. The version file wasn't being picked up properly because it was created on the VM.

These changes address the issue. Please see the comments in the code changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ NAME := netplugin
 # We are using date based versioning, so for consistent version during a build
 # we evaluate and set the value of version once in a file and use it in 'tar'
 # and 'release' targets.
-VERSION_FILE := /tmp/$(NAME)-version
+VERSION_FILE := $(NAME)-version
 VERSION := `cat $(VERSION_FILE)`
 TAR_EXT := tar.bz2
 TAR_FILENAME := $(NAME)-$(VERSION).$(TAR_EXT)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -12,7 +12,7 @@ NAME := netplugin
 # We are using date based versioning, so for consistent version during a build
 # we evaluate and set the value of version once in a file and use it in 'tar'
 # and 'release' targets.
-VERSION_FILE := /tmp/$(NAME)-version
+VERSION_FILE := $(NAME)-version
```
unclejack: This was needed because this file was only on the VM, not on the host's file system.

Makefile
```diff
@@ -280,6 +280,7 @@ tar: clean-tar
 
 clean-tar:
 	@rm -f $(TAR_LOC)/*.$(TAR_EXT)
+	@rm -f ${VERSION_FILE}
```
unclejack: This cleans up the version file.

Vagrantfile
```diff
@@ -54,11 +55,6 @@ else
     curl -sSL https://storage.googleapis.com/golang/go#{go_version}.linux-amd64.tar.gz  | sudo tar -xz -C /usr/local
 fi
 
-if [[ $# -gt 9 ]] && [[ $9 != "" ]]; then
```
unclejack: This was probably not having the intended outcome.

This was putting `export classic_mode` before this change in the `/etc/profile.d/envvar.sh` file. This isn't helpful at all.

I've removed all this code. This doesn't affect the resulting `envvar.sh`.

PR #832, opened 2017-04-13
Makefile: run tar and clean-tar targets on the CI

unclejack:
This PR adds the `tar` and `clean-tar` targets to the CI. This ensures this test runs on the CI and would help catch potential issues with the tarballs which need to be released.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,7 @@ all-CI: stop clean start
 		&& cd /opt/gopath/src/github.com/contiv/netplugin \
 		&& make host-unit-test host-integ-test host-build-docker-image"'
 	make system-test
+	make tar clean-tar
 
 test: build unit-test system-test ubuntu-tests
 


jojimt:
From what I can tell, all we are doing here is a fresh build & tar. Instead of running this as part of CI which already does "build", can we make this test part of the release script? That way, we will avoid adding more time to an already time-consuming CI.

unclejack:
@jojimt: The release target depends on the `tar` target and it calls `clean-tar` when it finishes running. This makes the release target a superset of this test.

You're right that it does add more to the execution time of the tests. Would you prefer that we make this change later on when we address the long execution of the tests? I don't have a strong opinion either way. I just wish to ensure everything gets tested as much as possible.

jojimt:
Perhaps you could add a verification to the release target. That should cover what you're trying to cover here.

unclejack:
This will require a different approach. I'll close this now.

PR #835, opened 2017-04-14
Adding listen & control url parameters

neelimamukiri:


DivyaVavili:
Are we using contivInit.sh? If so, we have to expose it in that script as well. If not, we have to cleanup the file.

DivyaVavili:
Changes in this file LGTM

neelimamukiri:
We are not using contivInit.sh for the k8s installer. I am not sure if it is being used else where. We can make the changes to that and test it in a different PR - if it is being used else where.

neelimamukiri:
@DivyaVavili, @rhim, we need this PR for the corresponding installer change https://github.com/contiv/install/pull/108

scripts/netContain/scripts/contivNet.sh
```diff
@@ -16,6 +16,8 @@ cleanup=false
 cstore_param=""
 vtep_ip_param=""
 vlan_if_param=""
+control_url=":9999"
```
rchirakk: do you need to pass default control_url/listen_url ?
Isn't that taken care by netmaster if not specified ?

neelimamukiri: You don't need to pass the arguments to netmaster. These are the defaults.

PR #836, opened 2017-04-14
verify that the tenant name exists before listing config

rchirakk:
fiixes https://github.com/contiv/ccn/issues/292


jojimt:
LGTM

PR #837, opened 2017-04-14
refactor Vagrantfile slightly & make it possible to install Docker CE 17.03+

unclejack:
This PR makes a few small changes. The first commit makes the following changes:
- removes argument passing to the provisioning script by using interpolation and templates
- removes some unneeded `if var != empty && var == value` by moving these to the top list of variables
- converts some strings to use single quotes for quoting
- moves one script outside of the Vagrant provisioning block
- removes numbered variables from the provisioning scripts with templated and interpolated strings

The reason why I've made these changes was to make it easier to follow what variables are used, how and where. This makes it possible to get rid of most hacks and it's much easier to go over the Vagrantfile.

The second commit makes it possible to install and use Docker 17.03 CE and newer. This can be tested using `CONTIV_DOCKER_VERSION=17.03.1.ce make stop start`.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -15,15 +15,20 @@ BEGIN {
 }
 
 # netplugin_synced_gopath="/opt/golang"
-go_version = ENV["GO_VERSION"] || "1.7.5"
-docker_version = ENV["CONTIV_DOCKER_VERSION"] || "1.12.6"
-docker_swarm = ENV["CONTIV_DOCKER_SWARM"] || "classic_mode"
-gopath_folder="/opt/gopath"
+go_version = ENV['GO_VERSION'] || '1.7.5'
+docker_version = ENV['CONTIV_DOCKER_VERSION'] || '1.12.6'
+docker_swarm = ENV['CONTIV_DOCKER_SWARM'] || 'classic_mode'
+gopath_folder = '/opt/gopath'
 http_proxy = ENV['HTTP_PROXY'] || ENV['http_proxy'] || ''
 https_proxy = ENV['HTTPS_PROXY'] || ENV['https_proxy'] || ''
 build_version = ENV['BUILD_VERSION'] || ''
-
-cluster_ip_nodes = ""
+cluster_ip_nodes = ''
+v2plugin_name = ENV['CONTIV_V2PLUGIN_NAME'] || 'contiv/v2netplugin:0.1'
+cluster_store = ENV['CONTIV_CLUSTER_STORE'] || 'etcd://localhost:2379'
+use_release = ENV['USE_RELEASE'] || ''
+node_os = ENV['CONTIV_NODE_OS'] || 'centos'
+base_ip = ENV['CONTIV_IP_PREFIX'] || '192.168.2.'
+num_nodes = ENV['CONTIV_NODES'].to_i == 0 ? 3 : ENV['CONTIV_NODES'].to_i
 
 provision_common_once = <<SCRIPT
 ## setup the environment file. Export the env-vars passed as args to 'vagrant up'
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -33,15 +38,15 @@ echo 'export GOPATH=#{gopath_folder}' > /etc/profile.d/envvar.sh
...
-echo "export http_proxy='$3'" >> /etc/profile.d/envvar.sh
-echo "export https_proxy='$4'" >> /etc/profile.d/envvar.sh
-echo "export USE_RELEASE=$5" >> /etc/profile.d/envvar.sh
-echo "export no_proxy=$2,127.0.0.1,localhost,netmaster" >> /etc/profile.d/envvar.sh
-echo "export CLUSTER_NODE_IPS=$2" >> /etc/profile.d/envvar.sh
-echo "export CONTIV_CLUSTER_STORE=$6" >> /etc/profile.d/envvar.sh
-echo "export CONTIV_V2PLUGIN_NAME=$9" >> /etc/profile.d/envvar.sh
-echo "export CONTIV_DOCKER_SWARM=${10}" >> /etc/profile.d/envvar.sh
-echo "export BUILD_VERSION=${11}" >> /etc/profile.d/envvar.sh
+echo "export http_proxy='#{http_proxy}'" >> /etc/profile.d/envvar.sh
```
dseevr: While you're cleaning this up, do you want to do something like:

```sh
cat >>/etc/profile.d/envvar.sh <<EOS
export FOO=foo
export BAR=bar
EOS
```

It'll cut the character count in half and make it easier to read :smile:

unclejack: Sure, I'll make the change. It's a good idea.

unclejack: I've made the changes. PTAL

dseevr: Looks good 👍

unclejack:
This is green and ready to go. PTAL

Vagrantfile
```diff
@@ -15,33 +15,40 @@ BEGIN {
...
+docker_version = ENV['CONTIV_DOCKER_VERSION'] || '1.12.6'
+docker_swarm = ENV['CONTIV_DOCKER_SWARM'] || 'classic_mode'
+gopath_folder = '/opt/gopath'
 http_proxy = ENV['HTTP_PROXY'] || ENV['http_proxy'] || ''
 https_proxy = ENV['HTTPS_PROXY'] || ENV['https_proxy'] || ''
 build_version = ENV['BUILD_VERSION'] || ''
-
-cluster_ip_nodes = ""
+cluster_ip_nodes = ''
+v2plugin_name = ENV['CONTIV_V2PLUGIN_NAME'] || 'contiv/v2netplugin:0.1'
```
rhim: pls check that it is the latest version available

Vagrantfile
```diff
@@ -207,7 +253,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     if Vagrant.has_plugin?("vagrant-vbguest")
         config.vbguest.auto_update = false
     end 
-    if ENV['CONTIV_NODE_OS'] && ENV['CONTIV_NODE_OS'] == "ubuntu" then
+    if node_os == "ubuntu" then
```
rhim: above, syntax used is #{node_os}. Is there a reason for a different syntax here?

Vagrantfile
unclejack: @rhim: This piece was exactly like this before. The v2plugin isn't released and there's no available version right now.

Vagrantfile
unclejack: `#{var}` is used to replace this with the value of the global variable with that same name. 

`%{var}` is used for the ruby template to do the same thing with local variables. The template is used to avoid passing arguments to the shell script (e.g.: s.args $foo, $bar, $baz become $1, $2, $3 in the shell script and we have to keep track of which variable was what when we modify the script). Instead of the numbered variables in the shell script, we have `%{foo}`, `%{bar}` `%{baz}` and we can pass a ruby hash to replace those with certain values. This is more straightforward and makes it easier to follow.

It's node_os here because we're in ruby code. The other occurrences were in strings which become templates. I've written this to explain why I've made the change.

rhim: Thanks for the clarification.

dseevr: `String#%` taking a hash argument is really uncommon in Ruby code, and you can think of it being similar to Python's `"{foo} {bar}".format(foo="foo", bar="bar")` construct

unclejack:
@gkvijay Can you take a look at this when you have the time? Please comment if there's something you'd like to discuss.

PR #838, opened 2017-04-17
Fix a timing issue in network tests

jojimt:


test/systemtests/network_test.go
```diff
@@ -338,6 +338,7 @@ func (s *systemtestSuite) testNetworkAddDeleteTenant(c *C, encap, fwdmode string
 			}
 		}
 
+		time.Sleep(3 * time.Second)
```
gaurav-dalvi: what is this timing issue and why it will go away if we remove this 3 seconds wait ?

jojimt: The timing issue is the delay from issuing the command from the system test code to the command taking effect. Because it is vagrant, sometimes it is slow to propagate.

unclejack: Could we check if the networks have been created instead? I'm not saying I have anything against this change, quite the opposite. It's just that we're using sleep in a lot of places and I'm not sure it's always right.

jojimt: Yes, ideally we should be doing that. Our test infra does not have a way to do this for net-partition scenarios today.

PR #839, opened 2017-04-17
Added RemoteEndpoint/PolicyRule notifications in netplugin for VPP driver

gkvijay:


drivers/ovsdriver.go
```diff
@@ -556,6 +556,19 @@ func (d *OvsDriver) DeleteEndpoint(id string) error {
 	return nil
 }
 
+// CreateRemoteEndpoint creates a remote endpoint by named identifier
+func (d *OvsDriver) CreateRemoteEndpoint(id string) error {
+
+	log.Info("OVS driver ignoring remote EP create as it uses its own EP sync")
```
unclejack: Shouldn't this just return nil and not log anything?

gkvijay: This was added for debugging. I will change it to log.Debug .

netplugin/agent/state_event.go
```diff
@@ -446,6 +481,41 @@ func processSvcProviderUpdEvent(netPlugin *plugin.NetPlugin, svcProvider *master
 	return nil
 }
 
+// processPolicyRuleState updates policy rule state
+func processPolicyRuleState(netPlugin *plugin.NetPlugin, opts core.InstanceInfo, ruleID string, isDelete bool) error {
+	netPlugin.Lock()
+	defer func() { netPlugin.Unlock() }()
```
dseevr: This is usually written as just

```go
defer netPlugin.Unlock()
```
No need for an anonymous function

netmaster/mastercfg/policyState.go
```diff
@@ -293,6 +293,13 @@ func (gp *EpgPolicy) createOfnetRule(rule *contivModel.Rule, dir string) (*ofnet
 		return nil, err
 	}
 
+	// Send AddRule to netplugin agents
+	err = addPolicyRuleState(ofnetRule)
```
dseevr: When return values are not used outside of the `if` statement immediately following a function call, it's convention to write it like this to avoid putting the variables into the function scope at all:

```go
if err := some_func(); err != nil {
   // err exists here but not in the scope of the function this code is in
}
```

This also means you don't have to worry about using `:=` or `=` in the assignment.  You can always use `:=`

`processStateEvent()` in `state_event.go` has some examples of doing it with multiple return values

netmaster/mastercfg/policyRuleState.go
```diff
@@ -0,0 +1,84 @@
+/***
+Copyright 2014 Cisco Systems Inc. All rights reserved.
```
dseevr: 2017 :smile:

netplugin/agent/state_event.go
```diff
@@ -250,6 +252,39 @@ func processEpState(netPlugin *plugin.NetPlugin, opts core.InstanceInfo, epID st
 	return err
 }
 
+// processRemoteEpState updates endpoint state
+func processRemoteEpState(netPlugin *plugin.NetPlugin, opts core.InstanceInfo, epCfg *mastercfg.CfgEndpointState, isDelete bool) error {
+	// take a lock to ensure we are programming one event at a time.
+	netPlugin.Lock()
+	defer func() { netPlugin.Unlock() }()
```
dseevr: here too

netplugin/agent/state_event.go
```diff
@@ -250,6 +252,39 @@ func processEpState(netPlugin *plugin.NetPlugin, opts core.InstanceInfo, epID st
...
+
+	if !checkRemoteHost(epCfg.VtepIP, epCfg.HomingHost, opts.HostLabel) {
+		// Skip local endpoint update, as they are handled directly in dockplugin
+		return nil
+	}
+
+	var err error
```
dseevr: this var can be completely removed if you switch the err checks to be inline in the `if` statements

netplugin/agent/state_event.go
gkvijay: made this change

netplugin/agent/state_event.go
gkvijay: var removed and used ':' in each block

netplugin/agent/state_event.go
gkvijay: Done.

netmaster/mastercfg/policyRuleState.go
gkvijay: Updated to 2017

netmaster/mastercfg/policyState.go
gkvijay: In this particular case, remoteEpgID has a function scope and using ':' for err will cause remoteEpgID also to be block scoped and hence I used the var. I agree with your suggestion for boolean checks (esp for map access and the processStateEvent example), but for explicit err value checks we should not enforce this rule. For code readability I feel its better to split it into multiple lines.

dseevr: Up to you whether to follow it or not, but inline error/ok checks are convention in Go code (and in the Go compiler + stdlib's code).  Our projects seem to follow it about half of the time 😄

PR #840, opened 2017-04-18
Fix issue 422

mhrobinson:
Fix tab not working for all the commands

Signed-off-by: Mike Robinson <mikerob@cisco.com>

### DIFF
--- a/scripts/contrib/completion/bash/netctl
+++ b/scripts/contrib/completion/bash/netctl
@@ -1,11 +1,19 @@
-#! /bin/bash
+#!/bin/bash
 
 GLOBAL_COMMANDS="\
-    group\
-    network\
-    tenant\
-    policy\
+    aci-gw\
+    app-profile\
+    bgp\
+    endpoint\
+    external-contracts\
     global\
+    group\
+    netprofile\
+    network\
+    policy\
+    service\
+    tenant\
+    version\
     help"
 
 GLOBAL_OPTIONS="\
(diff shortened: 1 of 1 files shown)


mhrobinson:
Currently extending the fix to include auto completion for all the command options as well.

scripts/contrib/completion/bash/netctl
```diff
@@ -1,11 +1,19 @@
...
     network\
     tenant\
     policy\
+    external-contracts\
     global\
+    aci-gw\
+    bgp\
+    app-profile\
+    service\
     help"
```
rchirakk: Can you keep this list sorted ?

dseevr: This is actually the current order of the command output from `netctl --help`

We need to reorder the `Commands` array in `netctl/commands.go`, too.

mhrobinson: Should we re-order bash/netctl and netctl/commands.go in a separate fix after this to keep this diff readable?

dseevr: That sounds good.  You weren't included in that "we" FYI.  That's our problem  😄

mhrobinson:
Updated the diff to include completion for all the command options as well.

scripts/contrib/completion/bash/netctl
rchirakk: ```>>re-order netctl/commands.go ```
done, https://github.com/contiv/netplugin/pull/848

mhrobinson: The completion script automatically sorts the completions when it displays them, the same as your above fix for the help.  So unless we re-order commands.go, it seems to make sense to leave this script in the original matching order.  All the bash functions in the script are also laid out the original order, so ideally we would then need to alphabetize the entire file.

```text
[vagrant@netplugin-node1 netplugin]$ netctl 
aci-gw              external-contracts  netprofile          tenant
app-profile         global              network             version
bgp                 group               policy              
endpoint            help                service             
```

rchirakk: A sorted list would be easier to compare with netctl --help to check missing  commands

mhrobinson: That is a valid point. :)  I've re-ordered the command lists in the completion script to match the help output.

mhrobinson:
Updated the diff to re-order the completion script command lists to match the help output.

PR #847, opened 2017-04-21
install/v2plugin/Dockerfile: remove maintainer

unclejack:
The `MAINTAINER` instruction is being deprecated in Docker. This means we should also remove it from our Dockerfiles.


gaurav-dalvi:
LGTM

PR #848, opened 2017-04-21
sort netctl command names in help

rchirakk:


netctl/netctl/main.go
```diff
@@ -2,16 +2,25 @@ package main
...
+
+func (a byName) Len() int           { return len(a) }
+func (a byName) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a byName) Less(i, j int) bool { return a[i].Name < a[j].Name }
+
 func main() {
 	app := cli.NewApp()
 	app.Flags = netctl.NetmasterFlags
 	app.Version = "\n" + version.String()
+	// use sort.Slice() in go1.8
```
gaurav-dalvi: wouldn't it break backward compatibility with earlier golang version ?

rchirakk: we can bump up the min. go version requirement to 1.8

unclejack: @rchirakk: We're not going to upgrade soon.

dseevr: Go 1.8.0 was a disastrous release with many bugs so I'd vote for skipping it 😄

rchirakk: @unclejack @dseevr how is 1.8.1.
The  comment in the code was to point that this PR can be 
changed to a single line in > go-1.8

dseevr: @rchirakk Yeah, that makes sense.  I might even make it a `TODO` so it's easier to find.

Based on the [bugs being fixed for 1.8.2](https://github.com/golang/go/milestone/54), it seems like 1.8.1 was an okay release

Issue #859, opened 2017-04-28
netctl group inspect -j creates invalid json


riffraff169:
Output includes the line:
Inspeting endpointGroup: epgA tenant: TestTenant
which (ignoring spelling error), makes the resulting json output invalid.  Removing this line makes it json parsing work.

PR #860, opened 2017-04-28
Removing non-json text because it creates invalid json output

riffraff169:
For ticket #859

### DIFF
--- a/netctl/netctl.go
+++ b/netctl/netctl.go
@@ -65,8 +65,6 @@ func inspectPolicy(ctx *cli.Context) {
 	tenant := ctx.String("tenant")
 	policy := ctx.Args()[0]
 
-	fmt.Printf("Inspeting policy: %s tenant: %s\n", policy, tenant)
-
 	pol, err := getClient(ctx).PolicyInspect(tenant, policy)
 	errCheck(ctx, err)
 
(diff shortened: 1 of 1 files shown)


netctl/netctl.go
```diff
@@ -65,8 +65,6 @@ func inspectPolicy(ctx *cli.Context) {
 	tenant := ctx.String("tenant")
 	policy := ctx.Args()[0]
 
-	fmt.Printf("Inspeting policy: %s tenant: %s\n", policy, tenant)
```
dseevr: I would consider changing these to `log.Debugf` so they're available if you `DEBUG=1` but won't show up in normal usage

dseevr:
If you change the description to

```
Fixes #859
```

the linked issue will be auto-closed when the PR is merged

netctl/netctl.go
```diff
@@ -844,8 +833,6 @@ func inspectBgp(ctx *cli.Context) {
 
 	hostname := ctx.Args()[0]
 
-	fmt.Printf("netctl. Inspecting bgp: %s\n", hostname)
```
gaurav-dalvi: How does exactly removing these printfs making o/p valid json

dseevr:
```
blah blah
{"foo":"bar"}
```

isn't JSON output that be parsed without manually tidying it up first.  If you request JSON with `-j`, the output should be parseable as JSON

PR #861, opened 2017-04-28
Removing non-json text because it creates invalid json output

### LINKED ISSUE #859: netctl group inspect -j creates invalid json

riffraff169:
Fixes #859

### DIFF
--- a/netctl/netctl.go
+++ b/netctl/netctl.go
@@ -163,8 +163,6 @@ func inspectPolicy(ctx *cli.Context) {
 	tenant := ctx.String("tenant")
 	policy := ctx.Args()[0]
 
-	fmt.Printf("Inspecting policy: %s tenant: %s\n", policy, tenant)
-
 	pol, err := getClient(ctx).PolicyInspect(tenant, policy)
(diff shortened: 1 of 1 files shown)


netctl/netctl.go
```diff
@@ -65,7 +65,7 @@ func inspectPolicy(ctx *cli.Context) {
 	tenant := ctx.String("tenant")
 	policy := ctx.Args()[0]
 
-	fmt.Printf("Inspeting policy: %s tenant: %s\n", policy, tenant)
```
rchirakk: import log package

dseevr: LOL.  Nice catch.

dseevr:
Actually, why don't we just delete these lines?

If I type `netctl policy inspect ...`, printing "inspecting policy" doesn't really add any value, and as we've seen it breaks the expectation that the output is JSON

PR #867, opened 2017-05-03
Remove json option from inspect commands that don't need it: tenant, …

riffraff169:
…policy, group

Signed-off-by: Lance Dillon <landillo@cisco.com>


dseevr:
@riffraff169 the changes look fine, but can you squash these two commits into one?  (using `git rebase -i`)

Issue #844, opened 2017-04-20
Avoid spurious endpointUpdates to master


jojimt:
Current implementation of Docker events handling can result in spurious updates being sent to netmaster. Appropriate filtering must be applied before sending the update. Additionally, it may be better to avoid listening to docker events directly when it is not necessary.

PR #868, opened 2017-05-04
process docker events only when lb services exist

rchirakk:
lb services are only interested in docker events
so process it only when lb services exist

Signed-off-by: rchirakk <rchirakk@users.noreply.github.com>

### DIFF
--- a/netplugin/agent/docker_event.go
+++ b/netplugin/agent/docker_event.go
@@ -23,6 +23,7 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/contiv/netplugin/mgmtfn/dockplugin"
 	"github.com/contiv/netplugin/netmaster/master"
+	"github.com/contiv/netplugin/netmaster/mastercfg"
 	"github.com/contiv/netplugin/netplugin/cluster"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/events"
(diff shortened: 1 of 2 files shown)


rchirakk:
fixes https://github.com/contiv/netplugin/issues/844

DivyaVavili:
LGTM

netplugin/agent/docker_event.go
```diff
@@ -47,6 +59,10 @@ func handleDockerEvents(events <-chan events.Message, errs <-chan error) {
 			}
 		case event := <-events:
 			log.Debugf("Received Docker event: {%#v}\n", event)
+			// process events only when LB services exist.
+			if ag.lbServiceExist() != true {
```
dseevr: This can just be `if !ag.lbServiceExist() {`

netplugin/agent/docker_event.go
```diff
@@ -31,9 +32,20 @@ import (
...
 
+func (ag *Agent) lbServiceExist() bool {
+	lbCfg := &mastercfg.CfgServiceLBState{}
+	plugin := ag.Plugin()
+	lbCfg.StateDriver = plugin.StateDriver
+
+	if lbList, err := lbCfg.ReadAll(); err == nil {
+		return len(lbList) > 0
+	}
+	return false
```
dseevr: In the case of etcd/consul being temporarily unavailable or some other error besides `key not found`, this function will return false... is that okay?

rchirakk: its is ok here.
```readAll()``` takes care of retry/cluster unavailable 
and there is no point in proceeding in case of error or key not found

PR #877, opened 2017-05-12
k8s watch optimization

DivyaVavili:
Stop processing service and ep updates for kube-system namespace. Processing these updates is causing a log of churn and filling up the log files as well.

### DIFF
--- a/mgmtfn/k8splugin/kubeClient.go
+++ b/mgmtfn/k8splugin/kubeClient.go
@@ -284,9 +284,11 @@ func (c *APIClient) WatchServices(respCh chan SvcWatchResp) {
 				continue
 			}
 
-			//if wss.Object.ObjectMeta.Namespace != "default" {
-			//	continue
-			//}
+			if wss.Object.ObjectMeta.Namespace == "kube-system" && (wss.Object.ObjectMeta.Name == "kube-scheduler" || wss.Object.ObjectMeta.Name == "kube-controller-manager") {
+				// Ignoring these frequent updates
+				continue
+			}
+
 			resp := SvcWatchResp{opcode: wss.Type}
 			resp.svcName = wss.Object.ObjectMeta.Name
 			sSpec := core.ServiceSpec{}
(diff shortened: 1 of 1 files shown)


mgmtfn/k8splugin/kubeClient.go
```diff
@@ -284,9 +284,10 @@ func (c *APIClient) WatchServices(respCh chan SvcWatchResp) {
 				continue
 			}
 
-			//if wss.Object.ObjectMeta.Namespace != "default" {
-			//	continue
-			//}
+			if strings.Compare(wss.Object.ObjectMeta.Namespace, "kube-system") == 0 {
```
dseevr: If you're checking for the result to be 0, there's no difference vs just using `==` here

From the docs: `The result will be 0 if a==b`

mgmtfn/k8splugin/kubeClient.go
```diff
@@ -358,9 +359,10 @@ func (c *APIClient) WatchSvcEps(respCh chan EpWatchResp) {
 				respCh <- EpWatchResp{opcode: "WARN", errStr: fmt.Sprintf("unmarshal %v", err)}
 				continue
 			}
-			//if weps.Object.ObjectMeta.Namespace != "default" {
-			//	continue
-			//}
+
+			if strings.Compare(weps.Object.ObjectMeta.Namespace, "kube-system") == 0 {
```
dseevr: Same here

unclejack:
This PR appears to fix the issue only when using kubeadm based setups.

mgmtfn/k8splugin/kubeClient.go
rchirakk: wasn't this already covered  by ```ObjectMeta.Namespace != "default"``` ?

DivyaVavili: That was commented out

DivyaVavili: Also, it might be better to ignore the kube-system namespace, instead of just processing the default namespace

mgmtfn/k8splugin/kubeClient.go
```diff
@@ -284,9 +284,10 @@ func (c *APIClient) WatchServices(respCh chan SvcWatchResp) {
 				continue
 			}
 
-			//if wss.Object.ObjectMeta.Namespace != "default" {
-			//	continue
-			//}
+			if wss.Object.ObjectMeta.Namespace == "kube-system" {
+				continue
+			}
+
```
rchirakk: this could break access to  kube-dns service

mgmtfn/k8splugin/kubeClient.go
```diff
@@ -284,9 +284,10 @@ func (c *APIClient) WatchServices(respCh chan SvcWatchResp) {
 				continue
 			}
 
-			//if wss.Object.ObjectMeta.Namespace != "default" {
-			//	continue
-			//}
```
rchirakk: do yo you need to add namespace check here ? 
```if namespace == "kube-system" ```

DivyaVavili: If the intention was to just ignore the updates from kube-scheduler and kube-controller-manager, this fix should work and I've verified it. I can add the extra check if you absolutely feel the necessary to do so; but considering it's a temporary fix until we move to the new client-go or figure something out, we can leave it as it is.

DivyaVavili:
@dseevr: PTAL

dseevr:
LGTM

Issue #878, opened 2017-05-15
kube-dns - Liveness probe failed


truncj:
Hi guys, running into some issues with kube-dns pod. 

After my previous issue, I was advised to recreate the contivh1 infra network, which I have since recreated. Now I'm noticing in the kube-dns container of my pod is showing the following "liveness probe" error:

`Liveness probe failed: Get http://20.1.1.1:8080/healthz-dnsmasq: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)`

<img width="682" alt="screen shot 2017-05-15 at 11 45 13 am" src="https://cloud.githubusercontent.com/assets/1036796/26077482/2c66da40-398a-11e7-9617-959a78dfde87.png">

<img width="236" alt="screen shot 2017-05-15 at 2 19 00 pm" src="https://cloud.githubusercontent.com/assets/1036796/26077400/e5ceb8aa-3989-11e7-9ac8-9ba112cbcadc.png">

<img width="624" alt="screen shot 2017-05-15 at 4 24 42 pm" src="https://cloud.githubusercontent.com/assets/1036796/26077710/092e41ca-398b-11e7-859f-b0540b5850fc.png">

I've also noticed that there are no endpoints listed for kube-dns. Normally using calico, we see endpoints for kube-dns that reflect the IP of the kube-dns pod. 

<img width="299" alt="screen shot 2017-05-15 at 4 19 45 pm" src="https://cloud.githubusercontent.com/assets/1036796/26077525/569d7602-398a-11e7-91e5-d0ba3fe7fd2e.png">

Any ideas appreciated!

rchirakk:
@truncj you need to specify a gateway for default-net  20.1.1.0/24 (netctl net create )

truncj:
@rchirakk I recreated the cluster and created the default-net with a gateway (20.1.1.1)... The error is still the same.

`kubelet k8worker01}	spec.containers{kube-dns}	Warning	Unhealthy	Readiness probe failed: Get http://20.1.1.2:8081/readiness: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
`

rchirakk:
Can you try to access http://20.1.1.2:8081/readiness from host ? can we have a look at your setup ? webex ?

truncj:
[...]

Issue #864, opened 2017-05-01
Exposed Service not accessible using cluster IP


overip:
Hello, I am running into an issue. I installed k86 1.5.6 and Contiv 1.0.1. Then I created the network and assigned a gateway with the following command $ netctl net create -t default --subnet=20.1.1.0/24 -g 20.1.1.1 default-net
Then  I created a nginx container, exposed the service, then tried to wget the index.html using nginx (container name) instead of using the IP from both another container and the host itself. From a container I was able to successfully access the nginx container using its name instead of the IP, and downloaded the index.html, so the DNS in that case worked wonderfully. When I try to grab the index.html of the nginx container from the host, it does not work. Do you happen to know why?

[1:39]  
In order to reproduce this last test, just follow the steps on the master of the cluster
$ kubectl run nginx --image=nginx --replicas=2
$ kubectl expose deployment nginx --port=80 
$ kubectl get svc,pod
NAME             CLUSTER-IP      EXTERNAL-IP   PORT(S)   AGE
svc/kubernetes   10.254.0.1      <none>        443/TCP   15m
svc/nginx        10.254.239.17   <none>        80/TCP    3m
$ wget 10.254.239.17
--2017-05-01 19:26:48--  http://10.254.239.17/
Connecting to 10.254.239.17:80... failed: Connection timed out.
Retrying.

Thank you.

neelimamukiri:
@rhim @rchirakk , this is vxlan on AWS

rchirakk:
this requires new routes in host.

chungewa:
k8s + contiv/vxlan mode:
[root@netmaster multi-tier]# kubectl get service
NAME         CLUSTER-IP     EXTERNAL-IP   PORT(S)        AGE
frontend     10.108.41.31   <nodes>       80:32108/TCP   4m

**[root@k8s-worker-1 log]# curl 10.108.41.31:80
curl: (7) Failed connect to 10.108.41.31:80; refused**

[root@k8s-worker-1 log]# journalctl -xe -u kube-proxy
-- No entries --
[root@k8s-worker-1 log]#

rchirakk:
It is because ```wget``` picks wrong interface/source-ip (multiple nics in host)
the solution is to force the source ip
```wget --bind-address=(ip address of contivh1)  <nginx-service-ip>```

PR #880, opened 2017-05-18
access k8s services using cluster-ip

rchirakk:
fixes kube-dns failures.
fixes https://github.com/contiv/netplugin/issues/878,
https://github.com/contiv/netplugin/issues/864

### DIFF
--- a/mgmtfn/k8splugin/cniserver.go
+++ b/mgmtfn/k8splugin/cniserver.go
@@ -36,6 +36,11 @@ var kubeAPIClient *APIClient
 var pluginHost string
 var contivK8Config k8sutils.ContivConfig
 
+// GetK8sClusterIPRange returns k8s cluster ip range
+func GetK8sClusterIPRange() string {
+	return contivK8Config.SvcSubnet
+}
+
 // setUpAPIClient sets up an instance of the k8s api server
 func setUpAPIClient() *APIClient {
 	// Read config
--- a/netplugin/agent/state_event.go
+++ b/netplugin/agent/state_event.go
@@ -23,6 +23,7 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/contiv/netplugin/core"
+	"github.com/contiv/netplugin/mgmtfn/k8splugin"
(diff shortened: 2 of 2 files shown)


DivyaVavili:
@rchirakk: The fix LGTM for vxlan networks. How would we handle reachability of cluster IP in vlan network?

srampal:
@DivyaVavili I discussed this topic with Joji before he left. He indicated this host access should be only for the vxlan case and for the vlan case we should do it via external path. I am not sure about that and would like to revisit this with you.  Meanwhile I think we should let this fix go in for the vxlan case and come back to the vlan case in a followup issue if we all agree it needs to address that as well.Currently the contivh1 path is anyway not used for the vlan case and we can revisit that. Meanwhile, this fix can go ahead and get merged for the vxlan case.

DivyaVavili:
Let's not merge this yet. This fix is not required for pod-to-pod communication over cluster-IP. The changes here allow the host to access the services via cluster IP from the cluster nodes. Is there a use-case which requires this?

srampal:
@DivyaVavili This is definitely needed asap.   The use case is for Openshift (for example when using the Openshift Docker registry)

DivyaVavili:
@srampal: Currently we don't see any use-case for this. We need to have a discussion to understand the use-case for Openshift. Let's follow it up on a separate channel.

srampal:
I think the confusion has been on how we define the meaning of "use case". There is a valid use case for the original issue (being able to access k8s services from the host). However the term "use case" in your comment referred to the *currently proposed fix* which is simply adding a static route towards cluster IPs and may not be the right fix. So the *use case* is certainly valid but the *current fix* may not be the right one.

rchirakk:
reopening the PR

srampal:
Double checking the static route solution proposed earlier. Looks like it is the right fix for this case because Linux source IP selection is based on which outgoing interface to use for the best next hop towards the destination and then use the primary IP address from that outgoing interface.  Will double check this with some folks in case yet another solution option is possible else we should merge the static route

srampal:
Didnt mean to close the PR ...

PR #882, opened 2017-05-23
Harden and simplify release script logic

dseevr:
- added "set -euo pipefail"
- fixes a bug where rereleasing a version resulted in a foo.tar.bz2.1 download being created and not actually used
- fixes a bug where extracting the netplugin release to "bin" failed silently if it existed and had files in it
- removed explicit call to `docker login`, now the cached login credentials will be used ("contiv" is an organization now)

### DIFF
--- a/scripts/netContain/release_image.sh
+++ b/scripts/netContain/release_image.sh
@@ -1,38 +1,31 @@
 #!/bin/bash
 
+set -euo pipefail
+
 contiv_version=""
-docker_user="contiv"
-docker_password=""
 image_name="contiv/netplugin"
 image_tag=""
 
 function usage() {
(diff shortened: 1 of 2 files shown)


scripts/netContain/release_image.sh
```diff
@@ -1,25 +1,26 @@
 #!/bin/bash
 
+set -euo pipefail
+
 contiv_version=""
 docker_user="contiv"
```
gkvijay: 'contiv' user should not be used as default, as it is not valid anymore. If it is not used, please remove

dseevr: Done, also removed the "password" option and removed username/password from getopt

PR #668, opened 2016-12-20
testing: migrate from gopkg.in/check.v1 to fork

unclejack:
This PR makes the required changes to migrate away from gopkg.in/check.v1. The fork includes useful features and fixes already:
- timeouts for each test which stop the whole suite
- flag to stop running the tests when a test fails
- a fix to avoid creating predictable temporary directories for testing

The first commit is already good to go. More commits will be added to start making use of the new features.

PR #886, opened 2017-05-26
Makefile: abort L3 tests early on failure

unclejack:
This PR makes one change which wasn't made by https://github.com/contiv/netplugin/pull/668. The L3 tests should also be interrupted when we run into the first failure. This change is identical to the change made to the system tests in #668.


unclejack:
@gkvijay: This is ready. PTAL

PR #888, opened 2017-06-01
Updated the v2plugin README.

gkvijay:


dseevr:
Can we figure out why this CI run failed?  Also, we are leaking a LOT of goroutines in our tests 😳

PR #893, opened 2017-06-07
k8s system test changes

DivyaVavili:
k8s system test changes
- create new make targets for legacy cluster "make k8s-legacy-cluster" and "make k8s-legacy-test"
- introduce kubeadm method of checking status on nodes
- create contiv_devtest.yml to mount local binaries for testing
- use contiv/netplugin:k8s_devtest for testing

NOTE: The test cluster is based off kubernetes 1.5.4 and kubeadm 1.6. Future improvements will move to the latest versions


DivyaVavili:
@unclejack: FYI

test/systemtests/network_test.go
```diff
@@ -12,14 +12,14 @@ import (
 )
 
 func (s *systemtestSuite) TestInfraNetworkAddDeleteVXLAN(c *C) {
-	if s.basicInfo.Scheduler == "k8" {
+	if s.basicInfo.Scheduler == kubeScheduler {
```
rhim: why are these tests being skipped for k8s?

test/systemtests/network_test.go
```diff
@@ -12,14 +12,14 @@ import (
...
 		return
 	}
 	s.testInfraNetworkAddDelete(c, "vxlan")
 }
 
 func (s *systemtestSuite) TestInfraNetworkAddDeleteVLAN(c *C) {
-	if s.basicInfo.Scheduler == "k8" {
+	if s.basicInfo.Scheduler == kubeScheduler {
```
rhim: why the skip for k8s?

Also, can you add a new line before line 203. I can't leave a comment there.

test/systemtests/network_test.go
```diff
@@ -380,6 +380,10 @@ func (s *systemtestSuite) TestNetworkAddDeleteTenantFwdModeChangeVXLAN(c *C) {
 	for i := 0; i < s.basicInfo.Iterations; i++ {
 		s.testNetworkAddDeleteTenant(c, "vxlan", fwdMode)
 		if fwdMode == "routing" {
+			if s.basicInfo.Scheduler == kubeScheduler {
```
rhim: see earlier comment about cleaning up this pattern

test/systemtests/network_test.go
```diff
@@ -389,9 +393,15 @@ func (s *systemtestSuite) TestNetworkAddDeleteTenantFwdModeChangeVXLAN(c *C) {
 				PvtSubnet:        "172.19.0.0/16",
 			}), IsNil)
 			time.Sleep(60 * time.Second)
+			if s.basicInfo.Scheduler == kubeScheduler {
```
rhim: same as above

test/systemtests/network_test.go
```diff
@@ -389,9 +393,15 @@ func (s *systemtestSuite) TestNetworkAddDeleteTenantFwdModeChangeVXLAN(c *C) {
...
+				c.Assert(s.SetupInfraNetwork(), IsNil)
+			}
 			s.testNetworkAddDeleteTenant(c, "vxlan", "bridge")
 			fwdMode = "bridge"
 		} else {
+			if s.basicInfo.Scheduler == kubeScheduler {
```
rhim: same as above.

test/systemtests/network_test.go
```diff
@@ -401,6 +411,9 @@ func (s *systemtestSuite) TestNetworkAddDeleteTenantFwdModeChangeVXLAN(c *C) {
 				PvtSubnet:        "172.19.0.0/16",
 			}), IsNil)
 			time.Sleep(60 * time.Second)
+			if s.basicInfo.Scheduler == kubeScheduler {
```
rhim: same as above.

Also, here the 60s sleep is outside of setupinfranetwork call. In other files (hostaccess_test.go) it is inside the if block.

test/systemtests/trigger_test.go
```diff
@@ -69,7 +69,7 @@ func (s *systemtestSuite) TestTriggerNetpluginUplinkUpgrade(c *C) {
 
 func (s *systemtestSuite) TestTriggerNetmasterSwitchover(c *C) {
 
-	if s.basicInfo.Scheduler == "k8" {
+	if s.basicInfo.Scheduler == kubeScheduler {
```
rhim: why skip?

test/systemtests/trigger_test.go
```diff
@@ -248,7 +248,7 @@ func (s *systemtestSuite) TestTriggerNetpluginDisconnect(c *C) {
 
 func (s *systemtestSuite) TestTriggerNodeReload(c *C) {
 	// can not run this test on docker 1.10 & k8s
-	if os.Getenv("CONTIV_DOCKER_VERSION") == "1.10.3" || s.basicInfo.Scheduler == "k8" {
+	if os.Getenv("CONTIV_DOCKER_VERSION") == "1.10.3" || s.basicInfo.Scheduler == kubeScheduler {
```
rhim: we should take anything related to 1.10.3 out now?

vagrant/k8s/Vagrantfile
```diff
@@ -114,133 +105,121 @@ Vagrant.configure(2) do |config|
...
-	   end
-	   c.vm.synced_folder "./../../bin","/opt/gopath/bin"
-	   c.vm.provision "shell" do |s|
-              s.inline = provision_k8
-               s.args = [ENV["http_proxy"] || "", ENV["https_proxy"] || ""]
-           end
-	else
-          if ENV['CONTIV_NODE_OS'] && ENV['CONTIV_NODE_OS'] == "rhel7" then
+           case vm_name
+	         when "k8master"
```
rhim: nit: indentation fix

vagrant/k8s/setup_cluster.sh
```diff
@@ -104,7 +104,11 @@ if [ "$legacyInstall" == 1 ]; then
    vagrant up
 else
    # Copy the contiv installation file to shared folder
-   cp -f ../../install/k8s/contiv/contiv.yaml ./export/.contiv.yaml
+   if [ "$k8s_devtest" == 1 ]; then
```
rhim: For this vagrant file, what's the scenario for devtest not to be set to 1?

scripts/netContain/scripts/contivNet.sh
```diff
@@ -114,34 +116,39 @@ if [ "$plugin" == "kubernetes" ]; then
    fi
 fi
 
+set +eo pipefail
```
dseevr: Worst case, turn set `+e` before the /contiv/bin/netmaster and /contiv/bin/netplugin calls and turn it on immediately afterward.  You could also leave `-e` enabled and just add `|| true` after the commands to ignore their exit codes when they crash.

There's nothing being piped, so there's no need to turn off `-o pipefail`

test/systemtests/kubeadm_test.go
```diff
@@ -0,0 +1,981 @@
...
+	}
+	logrus.Infof("Cleaning up slave on %s", k.node.Name())
+	podName, err := getPodName("netplugin", k.node.Name())
+	if err != nil {
+		logrus.Errorf("pod not found: %+v", err)
+		return
+	}
+
+	ovsCleanupCmd := `ovs-vsctl list-br | grep contiv | xargs -rt -I % ovs-vsctl del-br %`
+	_, err = k.podExec(podName, ovsCleanupCmd, "kube-system")
```
dseevr: If we're not checking `err` or printing any warning, there's no reason to assign it.  You could get rid of the `=` and everything to the left of it entirely

test/systemtests/kubeadm_test.go
```diff
@@ -0,0 +1,981 @@
...
+
+	"github.com/Sirupsen/logrus"
+	"github.com/contiv/contivmodel/client"
+	"os"
+	"time"
+)
+
+type kubePod struct {
+	node *node
+}
```
dseevr: Could you add a `func (k *KubePod) isMaster() bool` here and call that instead of repeating `name == "k8smaster"` many times in this file?

scripts/netContain/scripts/contivNet.sh
DivyaVavili: Yes, we do expect the script to continue when the process is killed by system-tests. The only commands beyond this point is to check the status of the netplugin/netmaster process and re-spawn them as necessay (dictated by presence of /tmp/restart* files)

scripts/netContain/scripts/contivNet.sh
```diff
@@ -19,6 +19,9 @@ vlan_if_param=""
 control_url=":9999"
 listen_url=":9999"
 
+touch /tmp/restart_netmaster
```
DivyaVavili: This is to give explicit control to the system-test framework to kill the netplugin and netmaster process and start them with custom arguments which is required to test different scenarios. Without this control, the script automatically spawns off the netplugin/netmaster process within 5 sec. Will insert a comment here.

install/k8s/contiv/contiv_devtest.yaml
```diff
@@ -0,0 +1,275 @@
...
+---
+
+# This manifest installs the Contiv etcd on the kubeadm master. 
+# If using an external etcd instance, this can be deleted. This uses a DaemonSet
+# to force it to run on the master even when the master isn't schedulable, and uses
+# nodeSelector to ensure it only runs on the master.
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: contiv-etcd
```
DivyaVavili: Having to install at test-cluster bringup time is only going to increase the testing time. I would rather have the compatible versions of dependencies installed and test the functionality of netplugin here. The end-to-end testing should ideally be the place to test the install and netplugin together. 

Yes, I do have a parallel effort to get this in sync with k8s1.6 which install has and this is just the first pass to get things going.

scripts/python/createcfg.py
```diff
@@ -21,8 +23,10 @@
 parser.add_argument("-iterations", default=3, help="Number of iterations for each test")
 parser.add_argument("-enableDNS", default=False, help="Enabling DNS")
 parser.add_argument("-contiv_cluster_store", default="etcd://localhost:2379", help="cluster info")
+parser.add_argument("-k8s_contiv_cluster_store", default="etcd://192.168.2.10:6666", help="cluster info")
```
DivyaVavili: Will explore this

test/systemtests/basic_test.go
```diff
@@ -226,18 +226,27 @@ func (s *systemtestSuite) testBasicSvcDiscovery(c *C, encap string) {
 }
 
 func (s *systemtestSuite) TestBasicNetmasterPortListen(c *C) {
+	var masterNodeIndex int32
+	var masterNode *node
+	masterDefaultPort := "9999"
+
 	for _, node := range s.nodes {
 		// Stop all netmaster instances
 		c.Assert(node.stopNetmaster(), IsNil)
```
DivyaVavili: Can you explain further?

test/systemtests/node_test.go
```diff
@@ -138,6 +138,7 @@ func (n *node) checkForNetpluginErrors() error {
 
 func (n *node) runCommandWithTimeOut(cmd string, tick, timeout time.Duration) error {
 	runCmd := func() (string, bool) {
+		logrus.Debugf("Running cmd: %s", cmd)
```
dseevr: Might as well log the timeout value here too

test/systemtests/cfg.json
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+{"scheduler": "k8s", "install_mode": "kubeadm", "swarm_variable": "", "platform": "vagrant", "product": "netplugin", "aci_mode": "off", "short": false, "containers": 3, "iterations": 3, "enableDNS": false, "contiv_l3": "", "contiv_cluster_store": "etcd://192.168.2.10:6666", "keyFile": "/home/admin/.ssh/id_rsa", "binpath": "contiv/bin", "hostips": "192.168.2.10,192.168.2.11,192.168.2.12", "hostusernames": "admin,admin,admin", "dataInterfaces": "eth2", "mgmtInterface": "eth1", "vlan": "1120-1150", "vxlan": "1-10000", "subnet": "10.1.1.0/24", "gateway": "10.1.1.254", "network": "TestNet", "tenant": "TestTenant", "encap": "vlan"}
```
DivyaVavili: the k8s_* vars are only in createcfg.py which generates the cfg.json (which in turn is consumed by system-tests). We should remove adding this cfg.json in our repository as it changes every time a system test is run. I've already added this in gitignoe, but I guess we have to remove from the git first for not tracking this.

test/systemtests/hostaccess_test.go
```diff
@@ -1,26 +1,44 @@
...
 	fm := global.FwdMode
 	global.FwdMode = "routing"
+	if s.basicInfo.Scheduler == kubeScheduler {
+		c.Assert(s.TearDownInfraNetwork(), IsNil)
+	}
+
 	c.Assert(s.cli.GlobalPost(global), IsNil)
+
+	if s.basicInfo.Scheduler == kubeScheduler {
+		time.Sleep(60 * time.Second)
```
DivyaVavili: GlobalPost restarts OVS on fwd-mode change and we need to wait for things to settle down before continuing

vagrant/k8s/Vagrantfile
dseevr: Also:

```rb
c.vm.box = "contiv/" + vm_name.gsub("-", "")
```

test/systemtests/hostaccess_test.go
```diff
@@ -1,26 +1,44 @@
...
 	}
 
-	time.Sleep(30*time.Second)
+	time.Sleep(30 * time.Second)
 	global, err := s.cli.GlobalGet("global")
 	c.Assert(err, IsNil)
 	// save the FwdMode
 	fm := global.FwdMode
 	global.FwdMode = "routing"
+	if s.basicInfo.Scheduler == kubeScheduler {
```
DivyaVavili: Agree. Was planning to clean it up in second pass.

test/systemtests/init_test.go
```diff
@@ -25,6 +23,7 @@ type systemtestSuite struct {
 }
 type BasicInfo struct {
 	Scheduler    string `json:"scheduler"`      //swarm, k8s or plain docker
+	InstallMode  string `json:"install_mode"`   //legacy or kubeadm
```
DivyaVavili: We might end up use it for swarm-mode as well when the system-tests for that are added. It's still relevant at this point.

vagrant/k8s/Vagrantfile
```diff
@@ -114,133 +105,121 @@ Vagrant.configure(2) do |config|
...
-	   if vm_name == "k8node-03" then
-             network_name = "contiv_orange"
-           end
-	   if vm_name == "k8node-01" then
+          s.inline = provision_k8
+          s.args = [ENV["http_proxy"] || "", ENV["https_proxy"] || ""]
+        end
+
+        network_name = "contiv_purple"
+        if ENV['CONTIV_L3'] then
```
dseevr: Lots of weird indentation here... Ruby uses 2 spaces for indentation (no tabs)

test/systemtests/kubeadm_test.go
```diff
@@ -0,0 +1,981 @@
...
+	})
+	if err != nil {
+		logrus.Errorf("default-net not created. Err: %+v", err)
+		return err
+	}
+	time.Sleep(time.Second)
+	return nil
+}
+
+func (k *kubePod) newContainer(node *node, containerID, name string, spec containerSpec) (*container, error) {
```
DivyaVavili: can't change the name that easily. It's tied up at the framework level. Will change if this function is just used locally within this file.

test/systemtests/kubeadm_test.go
```diff
@@ -0,0 +1,981 @@
...
+	cont, err := k.newContainer(n, podID, spec.name, spec)
+	if err != nil {
+		logrus.Info(err)
+		return nil, err
+	}
+
+	return cont, nil
+}
+
+func (k *kubePod) checkPingFailure(c *container, ipaddr string) error {
```
DivyaVavili: These are called from the actual system-tests. These functions are just an implementation of the abstracted functions that the systemtests call.

test/systemtests/kubeadm_test.go
```diff
@@ -0,0 +1,981 @@
...
+	//  - Backticks and quotes here ensure that the command
+	//    is properly wrapped to execute on pod rather than on the node
+	//  - nohup along with & ensures that the process continues to live
+	//    after the shell is terminated
+	// Since the command is to run in BG in the pod and not on the node,
+	// RunCommandBackground is not required here
+
+	podCmd := `kubectl -n ` + namespace + ` exec ` + podName + ` -- nohup sh -c '\''` + args + ` &'\''`
+	logrus.Debugf("Pod Exec BG: Running command -- %s", podCmd)
+	return k8sMaster.tbnode.RunCommand(podCmd)
```
DivyaVavili: Hmmm. Good point. Let me check it and update accordingly.

test/systemtests/kubeadm_test.go
```diff
@@ -0,0 +1,981 @@
...
+		out, _ := k8sMaster.tbnode.RunCommandWithOutput(fmt.Sprintf("kubectl get pod %s", c.containerID))
+		if strings.Contains(out, "not found") {
+			return nil
+		}
+		time.Sleep(5 * time.Second)
+	}
+	return fmt.Errorf("Error Termininating pod %s on node %s", c.name, c.node.Name())
+}
+
+func (k *kubePod) startListener(c *container, port int, protocol string) error {
```
DivyaVavili: Will continue to use this for now. If you feel a strong case to change this, will open another PR. Have to change it  in all modes and don't want to take the focus off this PR.

test/systemtests/kubeadm_test.go
```diff
@@ -0,0 +1,981 @@
...
+}
+
+func (k *kubePod) startIperfClient(c *container, ip, limit string, isErr bool) error {
+
+	var (
+		bwLimit int64
+		bwInt64 int64
+	)
+	bw, err := k.exec(c.containerID, fmt.Sprintf("iperf -c %s ", ip))
+	logrus.Infof("starting iperf client on : %v for server ip: %s", c, ip)
```
DivyaVavili: NOTE: Haven't changed these functions at all from legacy k8s mode. So, I have to look at these functions and simplify/refactor them as needed.

test/systemtests/kubeadm_test.go
```diff
@@ -0,0 +1,981 @@
...
+	if err != nil {
+		logrus.Errorf("Error starting the iperf client")
+	}
+	if strings.Contains(bw, "bits/sec") {
```
DivyaVavili: same as above

test/systemtests/kubeadm_test.go
```diff
@@ -0,0 +1,981 @@
...
+	outputInt, err := strconv.ParseInt(outputStr[0], 10, 64)
+	bwInt, err := BwConvertInt64(bw)
+	if err != nil {
+		return err
+	}
+	if bwInt == outputInt {
+		logrus.Infof("Applied bandwidth: %dkbits equals tc qdisc rate: %dkbits", bwInt, outputInt)
+	} else {
+		logrus.Errorf("Applied bandwidth: %dkbits does not match the tc rate: %d ", bwInt, outputInt)
+		return errors.New("Applied bandwidth doe sot match the tc qdisc rate")
```
DivyaVavili: same as above

test/systemtests/kubeadm_test.go
```diff
@@ -0,0 +1,981 @@
...
+	stopRestartCmd := `rm ` + netpluginRestartFile
+	k.exec(podName, stopRestartCmd, "kube-system")
+
+	logrus.Infof("Stopping netplugin on %s", k.node.Name())
+	killNetpluginCmd := `pkill netplugin`
+	_, err = k.exec(podName, killNetpluginCmd, "kube-system")
+	return err
+}
+
+func (k *kubePod) stopNetmaster() error {
```
DivyaVavili: Yes, pod will still remain. The process inside the container is killed and started with custom arguments as necessary by test cases

test/systemtests/kubeadm_test.go
```diff
@@ -0,0 +1,981 @@
...
+	logrus.Infof("Starting netmaster on %s", k.node.Name())
+	podName, err := getPodName("netmaster", k.node.Name())
+	if err != nil {
+		logrus.Errorf("pod not found: %+v", err)
+		return err
+	}
+
+	netmasterStartCmd := k.node.suite.basicInfo.BinPath + `/netmaster` + ` -cluster-store=` + k.node.suite.basicInfo.ClusterStore + ` -cluster-mode=kubernetes ` + args + ` > ` + netmasterLogLocation + ` 2>&1`
+
+	return k.podExecBG(podName, netmasterStartCmd, "kube-system")
```
DivyaVavili: yes

test/systemtests/kubeadm_test.go
```diff
@@ -0,0 +1,981 @@
...
+		return
+	}
+	clusterStoreInfo := strings.Split(k.node.suite.basicInfo.ClusterStore, "//")
+	etcdIP := clusterStoreInfo[len(clusterStoreInfo)-1]
+	logrus.Infof("Cleaning out etcd info on %s", etcdIP)
+
+	k.podExec(podName, `etcdctl -C `+etcdIP+` rm --recursive /contiv`, "kube-system")
+	k.podExec(podName, `etcdctl -C `+etcdIP+` rm --recursive /contiv.io`, "kube-system")
+	k.podExec(podName, `etcdctl -C `+etcdIP+` rm --recursive /docker`, "kube-system")
+	k.podExec(podName, `etcdctl -C `+etcdIP+` rm --recursive /skydns`, "kube-system")
```
DivyaVavili: Will do

test/systemtests/kubeadm_test.go
```diff
@@ -0,0 +1,981 @@
...
+			return str, errors.New(ep + " not found on " + k.node.Name())
+		}
+	}
+
+	return "", nil
+}
+
+//FIXME: This needs to be moved to node abstraction implmentation Once
+//that change is in.
+func (k *kubePod) reloadNode(n *node) error {
```
DivyaVavili: This is the actual vagrant box being reloaded. I have no history on why this was skipped for k8master. My guess is that - for legacy cases, the boxes were prebaked to be master and slaves and reloading master would cause the cluster to be broken. @neelimamukiri: What happens if the kubernetes master node is reloaded? Does any other node take over as master?

test/systemtests/network_test.go
DivyaVavili: Two different people. Two different coding styles :) Will make it consistent.

test/systemtests/trigger_test.go
DivyaVavili: same as above. Will track as separate issue

test/systemtests/trigger_test.go
DivyaVavili: I can remove the check for docker (will do it as a separate PR soon - want to keep this PR solely for kubeadm test cluster bringup).

vagrant/k8s/setup_cluster.sh
DivyaVavili: when "make k8s-legacy-test" target is run

vagrant/k8s/Vagrantfile
DivyaVavili: Yes, it looks weird here. I corrected the indentation on my local machine, but it doesn't look right here

vagrant/k8s/Vagrantfile
DivyaVavili: @dseevr: Does gsub remove '-' only if it exists?

test/systemtests/trigger_test.go
DivyaVavili: Create PR #896

vagrant/k8s/Vagrantfile
dseevr: @DivyaVavili yes, that's equivalent to `s/-//g`

dseevr: You can type `irb` and then `"foo-bar".gsub("-", "")` to test it locally 😄

test/systemtests/kubeadm_test.go
DivyaVavili: So,podExec which uses k.exec is also run on k8sMaster. Just abstracted one more level. These are kubectl commands and needs to run on k8sMaster.

test/systemtests/init_test.go
```diff
@@ -56,6 +55,13 @@ type GlobInfo struct {
 	Encap   string `json:"encap"`
 }
 
+const (
+	kubeScheduler  string = "k8s"
+	swarmScheduler string = "swarm"
+	legacyInstall  string = "legacy"
```
DivyaVavili: Might be relevant for v2plugin as well. Retaining for now

vagrant/k8s/Vagrantfile
```diff
@@ -114,133 +105,121 @@ Vagrant.configure(2) do |config|
...
+             when "k8node-03"
+               c.vm.box = "contiv/k8node03"
+             end
+	     else
+          if ENV['CONTIV_NODE_OS'] == "rhel7" then
             # Download rhel7.2 box from https://access.redhat.com/downloads/content/293/ver=2/rhel---7/2.0.0/x86_64/product-software
             # Add it as rhel7 vagrant box add rhel-cdk-kubernetes-7.2-29.x86_64.vagrant-virtualbox.box --name=rhel7
             c.vm.box = "rhel7"
           else
             c.vm.box = "contiv/k8s-centos"
```
unclejack: Do we really need to use this box or can we use another one? Is there something specific packaged in this Vagrant box?

unclejack:
The kubernetes tests pass with these changes. The contiv/k8s-centos box is using Go 1.6.3, not Go 1.7.6 as the main Vagrant box we're using right now. This is an issue because we're testing the code with Kubernetes using binaries compiled with a different version of Go.

We should try to use the contiv/centos73 Vagrant box, version 0.10.2. This requires some additional work to remove the bundled Docker and it might also require some other things to be run in the scripts which prepare the bootstrap of the nodes.

test/systemtests/network_test.go
DivyaVavili: We'll have to investigate later. Will open a separate PR.

test/systemtests/network_test.go
DivyaVavili: same as above. Will track as separate issue

DivyaVavili:
@unclejack: Good catch. the k8s-centos box is actually missing from contiv/. Looks like I was working off the image that was already on my setup. Moved it to contiv/centos73.

scripts/netContain/scripts/contivNet.sh
```diff
@@ -114,34 +117,39 @@ if [ "$plugin" == "kubernetes" ]; then
...
-		if [ "$cstore" != "" ]; then
-			cstore_param="-cluster-store"
-		fi
-		if [ "$vtep_ip" != "" ]; then
-			vtep_ip_param="-vtep-ip"
-		fi
-		if [ "$vlan_if" != "" ]; then
-			vlan_if_param="-vlan-if"
+		if [ -f /tmp/restart_netplugin ]; then
+			if [ "$cstore" != "" ]; then
```
gkvijay: Can you file an issue to clean this up and use the same startup script for both prod and test. Use docker cp to copy over the binary

test/systemtests/basic_test.go
rhim: yes, thanks for the clarification.

vagrant/k8s/setup_cluster.sh
rhim: Can you add a comment about what k8s_devtest means wherever it is defined?

unclejack:
I'm running into some failures with `make k8s-test`:
```
INFO[0136] Starting netmaster on k8master
ERRO[0143] default-net not created. Err: Cant change network parameters after its created [github.com/contiv/netplugin/netmaster/objApi.(*APIController).NetworkUpdate apiController.go 1146]


----------------------------------------------------------------------
FAIL: init_test.go:117: systemtestSuite.SetUpTest

init_test.go:124:
    s.SetUpTestVagrant(c)
util_test.go:1442:
    c.Assert(s.SetupDefaultNetwork(), IsNil)
... value *errors.errorString = &errors.errorString{s:"Cant change network parameters after its created [github.com/contiv/netplugin/netmaster/objApi.(*APIController).NetworkUpdate apiController.go 1146]\n"} ("Cant change network parameters after its created [github.com/contiv/netplugin/netmaster/objApi.(*APIController).NetworkUpdate apiController.go 1146]\n")

INFO[0145] ============================= systemtestSuite.TestBasicHostAccess completed ==========================

----------------------------------------------------------------------
PANIC: hostaccess_test.go:9: systemtestSuite.TestBasicHostAccess

... Panic: Fixture has panicked (see related PANIC)

----------------------------------------------------------------------
MISS: basic_test.go:223: systemtestSuite.TestBasicNetmasterPortListen

```

DivyaVavili:
@unclejack: Was this setup already up and was there a failure on the setup earlier? I've seen this issue in those cases.

unclejack:
@DivyaVavili: This setup was created a few minutes before the error was encountered. The source was the latest master (the merged PR from you) with this PR on top of it.

DivyaVavili:
Could you start off from a fresh setup and check if you are encountering the same issue?

unclejack:
@DivyaVavili That was a fresh setup I've brought up using the new source code. I've been testing kubernetes for a while and I don't keep clusters around between runs.

DivyaVavili:
Looks like a Makefile change didn't make it into the PR. I've repushed the PR including the change. This is the test log for reference: https://gist.github.com/DivyaVavili/2243f18b9421d6664ba658c83a8d2172

PR #895, opened 2017-06-09
update v2plugin README to include latest version of plugin

DivyaVavili:


install/v2plugin/README.md
```diff
@@ -32,12 +32,12 @@ dbg_flag      : To enable debug mode, set to '-debug'         : ""
 ### docker store
 Docker certified contiv plugin is avaliable on [Docker Store](https://store.docker.com/plugins/803eecee-0780-401a-a454-e9523ccf86b3?tab=description).
 ```
-docker plugin install store/contiv/v2plugin:1.0.0 iflist=eth1,eth2
+docker plugin install store/contiv/v2plugin:1.0.2 iflist=eth1,eth2
```
dseevr: This is intentionally 1.0.2, right?

DivyaVavili: Yes, I don't have the permission to docker store and I've applied to be a publisher. The latest version on the store is 1.0.2.

rhim: I see it is 1.0.3. The documentation should usually not contain a specific version like that (or just preface as example).

PR #896, opened 2017-06-10
Remove skipping tests for older docker versions

DivyaVavili:
## Description of the changes
Removing checks for older docker version as we no longer support it
#### Type of fix: 
Remove archaic checks

Cleaning up test framework to remove checks for older versions of docker

## NOTE
It is good to have these checks if we are going to support older versions of docker

### DIFF
--- a/test/systemtests/basic_test.go
+++ b/test/systemtests/basic_test.go
@@ -3,7 +3,6 @@ package systemtests
 import (
 	"fmt"
 	"math/rand"
-	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


unclejack:
@DivyaVavili: `goimports -w` has to be run on those two source files because of these errors:
```
# github.com/contiv/netplugin/test/systemtests
test/systemtests/basic_test.go:6: imported and not used: "os"
test/systemtests/trigger_test.go:6: imported and not used: "os"
```

PR #899, opened 2017-06-15
Godeps,vendor: bump ofnet to b44a11b

unclejack:
This PR bumps ofnet to switch to contiv/libOpenflow. This new libOpenflow library includes fixes for goroutines leak and race conditions. ofnet hasn't been modified in any other way. The only changes I've made to ofnet itself were the ones required to accommodate the new import paths.

The libOpenflow code has been properly tested and benchmarked. The contiv/libOpenflow PR which made these changes has also added some benchmarks and tests.


vendor/github.com/contiv/libOpenflow/util/stream.go
```diff
@@ -35,6 +37,8 @@ type MessageStream struct {
 	pool *BufferPool
 	// Message parser
 	parser Parser
+	// Channel to shut down the parser goroutine
+	parserShutdown chan bool
```
rchirakk: when do we send parser shutdown event ?

dseevr: It's sent later on in the diff of this file:

https://github.com/contiv/libOpenflow/commit/d9c950917b660d7a8258ffe5842f863e065f7d2f#diff-16ba25332f097a22ef695fe81b87c6e7R90

rchirakk: >> parser shutdown event 
how is it  triggered from netplugin & how to verify/test ?

dseevr: Looking at the original PR/discussion might give more context:  https://github.com/contiv/libOpenflow/pull/1

The tests are also in that PR

unclejack: @rchirakk The comments on that PR should show exactly where that happens. The outbound message processing goroutine has a select which watches the shutdown channel, in addition to watching for outbound messages to be sent out. We already had the means to shut down that outbound message processing queue. That PR has added a for loop which tells the individual goroutines which process inbound messages to shut down - just as was done before for the outbound messages goroutine. This means that the main shutdown procedure of this message sending & receiving code triggers the shutdown of the inbound message processing goroutines the same way the outbound message queue was shut down before.

This has no impact on performance or on how the code works in the end. If that code had been broken or its behavior had been altered in any way which caused issues, I would've discovered that.

unclejack:
I've added some more comments on the original PR. It seemed to be a good idea to explain those changes in order to make it clear they're isolated to cleanup and avoiding issues related to the Xid.

unclejack:
Can we merge this?

dseevr:
All of the changes have been approved here:  https://github.com/contiv/libOpenflow/pull/1

Merging per @rhim

PR #900, opened 2017-06-19
scripts/build.sh: strip unneeded debug symbols

unclejack:
This PR reduces the size of the binaries by removing the unneeded debugging information. This isn't needed for debugging crashes and other failures.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -30,5 +30,6 @@ echo $BUILD_VERSION >$VERSION_FILE
 GOGC=1500 go install \
 	-ldflags "-X $PKG_NAME.version=$BUILD_VERSION \
 	-X $PKG_NAME.buildTime=$BUILD_TIME \
-	-X $PKG_NAME.gitCommit=$GIT_COMMIT" \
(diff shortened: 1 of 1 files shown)


dseevr:
Nearly 50% savings on the binaries with these flags.  I ran a `make build` on HEAD and another one on this branch and copied the resulting binaries into folders:

```
➜  netplugin git:(master) ✗ du -sh netplugin_original 
138M    netplugin_original
➜  netplugin git:(master) ✗ du -sh netplugin_no_debug_info 
77M     netplugin_no_debug_info
```
And you still get the regular stack traces.  The only downside here is that you can no longer run these binaries through a debugger (gdb, etc.), but no one does that anyways so it's a free win.

rchirakk:
LGTM.
-s :  go tool nm is not going to work after strip @gkvijay  @DivyaVavili

DivyaVavili:
LGTM

gkvijay:
@rchirakk, debug tools requiring symbols will not work. We could use objcopy to save the symbols if needed.

PR #906, opened 2017-06-26
Makefile: use aci-gw:04-12-2017.2.2_1n

unclejack:
This PR updates the aci-gw image to a newer version which is smaller than the previous version.

TODO:
- [ ] figure out if we should test with multiple versions
- [ ] investigate potential issues

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -239,8 +239,8 @@ host-integ-test: host-cleanup start-aci-gw
 
 start-aci-gw:
 	@echo dev: starting aci gw...
-	docker pull contiv/aci-gw:11-28-2016.1.3_2i
-	docker run --net=host -itd -e "APIC_URL=SANITY" -e "APIC_USERNAME=IGNORE" -e "APIC_PASSWORD=IGNORE" --name=contiv-aci-gw contiv/aci-gw:11-28-2016.1.3_2i
+	docker pull contiv/aci-gw:04-12-2017.2.2_1n
+	docker run --net=host -itd -e "APIC_URL=SANITY" -e "APIC_USERNAME=IGNORE" -e "APIC_PASSWORD=IGNORE" --name=contiv-aci-gw contiv/aci-gw:04-12-2017.2.2_1n
 
 host-build-docker-image:
 	./scripts/netContain/build_image.sh


rchirakk:
What is new in aci-gw:04-12-2017.2.2_1n ?
Please confirm with @rhim if any changes are required in ACI GW.

unclejack:
@rchirakk: The new Docker image is about 3.5x smaller.

unclejack:
The CI is green now. PTAL

rchirakk:
How did you reduce the space, Is there any change to aci gateway code ?

unclejack:
@rchirakk: The Dockerfile was modified to use alpine as a base.

rchirakk:
@unclejack 
1. Can you run a few ACI tests to make sure it is not breaking anything. You can manually
 trigger or check with @gaurav-dalvi 
2. Is n't better to keep ```latest``` so that you don't have to change it ?

rchirakk:
I  see the  installer is using ```latest``` https://github.com/contiv/install/blob/master/scripts/build.sh
if aci-gw version is not passed to installer.
@neelimamukiri , do you specify ACI-GW version to install, or pick the latest ?
cc: @gkvijay @DivyaVavili

dseevr:
@rchirakk Every release from 1.0.1 to 1.1.0-rc.1 has used "latest" for `aci-gw`.  I wasn't involved in the first release so I'm not sure what went out in that one

unclejack:
@rchirakk @dseevr: The other releases have used the versions specified in the files. I didn't do anything to change that to another version. The releases were being made with the checked in files (e.g. no locally modified files or variables).

@rchirakk: We should specify a tag in here to make sure we have reproducible builds. If the image tagged `latest` has changed from one run of the CI to another, we no longer have reproducible tests. This isn't something we should allow.

However, we can make this a variable and make it possible to test using multiple aci-gw tags. Whould you prefer that we do this in addition to bumping the default?

This means we could also test using the latest tag and we'll be aware we're only testing the code works with latest, not that the particular build which uses latest is reproducible.

I've got to discuss with Gaurav to figure out the full testing process with ACI.

rchirakk:
@unclejack, since this is for CI,  current changes are good, no need to change it.
LGTM, make sure ACI sanity tests. 
@gkvijay @rhim to confirm about ACI version in installer 1.1

DivyaVavili:
LGTM. Please point us to ACI-sanity results.

rhim:
@unclejack do you have the aci-sanity results available?

dseevr:
@unclejack bumparino

rhim:
Go ahead and merge it. CI for ACI needs work, so ok to merge for now.

unclejack:
Thanks @rhim. I wasn't sure how this would be affected by the ACI related changes for testing.

neelimamukiri:
Latest is fine AFAIK. My understanding is the ACI SDK is backwards compatible, so any new version should work with any older version. The contiv/install repo has been using the smaller image for a while now, so this change should be fine.
@gaurav-dalvi usually tests and tags the images latest. For example, I pushed 2.3_1e recently, but the latest tag is still 04-12-2017.2.2_1n that Gaurav pushed.

unclejack:
@neelimamukiri: It's not that it wouldn't be fine. The main issue is that we need to have runs with identical parameters on the CI. I'll send a PR to add a variable to override the tag. The override can be used to test locally with latest.

PR #907, opened 2017-06-26
print error if policy doesn't exist 

rchirakk:
print error if policy doesn't exist instead of
listing empty rules.


DivyaVavili:
LGTM

PR #909, opened 2017-06-30
Makefile: build binaries before running k8s tests

unclejack:
This PR fixes an issue which breaks the k8s-test target. This issue was discussed with @DivyaVavili and she could've sent the PR as well. Since she's busy with other things right now, I've tested this change and sent the PR.

This is needed to enable the kubernetes tests on the CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,6 +143,8 @@ k8s-legacy-test:
 	CONTIV_K8S_LEGACY=1 CONTIV_NODES=3 go test -v -timeout 540m ./test/systemtests -check.v -check.f "00SSH|TestBasic|TestNetwork|ACID|TestPolicy|TestTrigger"
 	cd vagrant/k8s && vagrant destroy -f
 k8s-test: k8s-cluster
+	cd vagrant/k8s/ && \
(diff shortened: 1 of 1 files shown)


unclejack:
build PR

DivyaVavili:
Changes LGTM. CI failure is unrelated to this. Please collect info regarding this and file an issue. You can re-trigger the sanity.

rchirakk:
>>vagrant ssh k8master -c 'sudo -i

do you need ```sudo```  for builds ?

unclejack:
@rchirakk: https://github.com/contiv/netplugin/pull/909/files#diff-b67911656ef5d18c4ae36cb6741b7965L140 also uses sudo. It's the same thing.

unclejack:
The CI failure occurred with Docker 1.11.2. That's going to be replaced soon.

unclejack:
https://github.com/contiv/netplugin/issues/910 tracks the CI failure.

DivyaVavili:
@unclejack @rchirakk: We don't require sudo if we own the directories properly. I'm putting in a fix for this in 'provision_k8' to do this using 'sudo chown' instead of 'chown'. Based on which PR goes in first, we can make change to this appropriately.

PR #912, opened 2017-07-03
systemtests/hostaccess: remove redundant sleep

unclejack:
This PR removes one redundant sleep at the beginning of the TestBasicHostAccess test. Based on how the tests come up, this shouldn't be needed.

Let's see what happens on the CI.


DivyaVavili:
LGTM

PR #918, opened 2017-07-13
tests/systemtests/node_test.go: inlined the sole function we are using from an otherwise unused repo

dseevr:

The goal of this change is to allow us to delete the `systemtests-utils` repo as part of our cleanup effort


rchirakk:
LGTM

PR #919, opened 2017-07-14
netmaster/netplugin logs to k8s runtime

rchirakk:

### DIFF
--- a/scripts/netContain/scripts/contivNet.sh
+++ b/scripts/netContain/scripts/contivNet.sh
@@ -123,9 +123,9 @@ if [ $netmaster == true ]; then
 	while true; do
 		if [ -f /tmp/restart_netmaster ]; then
 			if [ "$cstore" != "" ]; then
-				/contiv/bin/netmaster $debug -cluster-mode $plugin -cluster-store $cstore -listen-url $listen_url -control-url $control_url &>/var/contiv/log/netmaster.log || true
+				/contiv/bin/netmaster $debug -cluster-mode $plugin -cluster-store $cstore -listen-url $listen_url -control-url $control_url || true
 			else
-				/contiv/bin/netmaster $debug -cluster-mode $plugin -listen-url $listen_url -control-url $control_url &>/var/contiv/log/netmaster.log || true
+				/contiv/bin/netmaster $debug -cluster-mode $plugin -listen-url $listen_url -control-url $control_url || true
 			fi
 			echo "CRITICAL : Netmaster has exited. Trying to respawn in 5s"
 		fi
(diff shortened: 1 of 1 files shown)


neelimamukiri:
Why don't we just remove the redirect and let the container runtime to handle the logs? Reason I am saying  is usually when we have a crash, we end up with a crash loop and so lastrun is likely to get overwritten soon.

unclejack:
This appears to be the responsibility of a different tool, as @neelimamukiri has also stated.

rchirakk:
We can use container runtime to handle  contivNet.sh output.
To keep consistency with other installations, we should use file based logging.

DivyaVavili:
@neelimamukiri: Is there a way for kubernetes to provide the log of the previous run? Like a history? If not, I think this will be good to have to get some details in the case of a crash.

DivyaVavili:
LGTM

neelimamukiri:
@DivyaVavili, typically you use external log management, same as you would for any other containers including kube-api-server, scheduler etc. Status of multiple pod restarts can be obtained with kubectl describe and kubectl logs show running pod logs only.

rhim:
Agree with @neelimamukiri. We should just focus on logging. Log rotation and retention are orthogonal aspects that should be covered separately, not by log generator.

@rchirakk @DivyaVavili ok to abandon?

rchirakk:
@rhim I think we should keep this fix,
1. to keep consistent logging with other installations
2. to keep the logs per node
3. Current logs are cleared on restart.

gkvijay:
@rhim,@neelimamukri : Agree that log rotation should be handled by log manager, but saving one last instance of the log would help debugging issues when restart of netplugin recovers from it. Debugging of netplugin restart during installation of k8s would have been easier with these logs. This is not a full log rotation and saves only one last instance. This was useful in v2plugin as well.

neelimamukiri:
@rchirakk, if we are going to redirect to a file (thats the current status), what you are proposing is an improvement and I dont see any issues in the change itself. 

If you want to you can go ahead with the change while we discuss and figure out the longer term strategy. My preference would be to follow the platform specific logging mechanism (syslog for systemd services, docker logs/kubectl logs for the container form). Usually Contiv is run along with a bunch of other infra components (other docker plugins, kube services which run in containers), so allowing the user to use a standard method is good for longer term maintainability. For example, with file redirect, what would be the way for a sysadmin to get netplugin logs from one now - ssh to the node and get, log collection/aggregation across nodes becomes tedious. If we redirect to the platform specific mechanism, you can collect all the logs from one place - syslog aggregator, container log aggregator etc.

unclejack:
In the case that the container gets restarted a few times, the previous log is quickly overwritten. Any relevant previously logged errors will get wiped out.

I don't have a strong opinion against something like this, but I don't expect this to be a silver bullet either. It seems to be a temporary workaround.

rchirakk:
modified to send contiv logs to kubectl logs, PTAL @unclejack @neelimamukiri @DivyaVavili @gkvijay @rhim

rchirakk:
send contiv logs to console so that it is available in docker logs/kubectl logs

PR #922, opened 2017-07-15
handle ovs-vsctl failures.

rchirakk:
ovs-vsctl fails during installation when ovs is not started.


rchirakk:
@rhim
 yes, linked to the issue

PR #923, opened 2017-07-19
exit startup script on ovs failure.

rchirakk:

### DIFF
--- a/scripts/netContain/scripts/contivNet.sh
+++ b/scripts/netContain/scripts/contivNet.sh
@@ -19,6 +19,11 @@ vlan_if_param=""
 control_url=":9999"
 listen_url=":9999"
 
+function die() {
+  echo "$*"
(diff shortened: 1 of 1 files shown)


scripts/netContain/scripts/contivNet.sh
```diff
@@ -135,7 +140,7 @@ if [ $netmaster == true ]; then
 	done
 elif [ $netplugin == true ]; then
 	echo "Starting netplugin"
-	modprobe openvswitch
+	modprobe openvswitch || die "failed to load openvswitch kernel module"
```
dseevr: Would including `$?` in the output be helpful?  I'm not sure how many exit codes `modprobe` can return lol

rchirakk: return code is not much useful, we need to check logs on failure.

PR #926, opened 2017-07-20
Makefile: allow aci-gw image to be overridden

unclejack:
This PR makes the changes discussed in #906. The new environment variable makes it possible to use another aci-gw Docker image for the integration tests. This should make it easy to test with different images.

Please keep in mind the fact that we can't use the latest tag by default because we need to use a certain version on the CI.


unclejack:
@rhim Can you take a look at this, please?

rhim:
@contiv/maintainers-core-networking PTAL and merge.

PR #933, opened 2017-07-24
WIP system tests improvements

unclejack:
This PR will make some improvements to the system tests.

This PR includes the following changes right now:
- [ ] lower the time.Sleep to 60 seconds for the Vagrant setup

There's no need to have such a long sleep for a Vagrant based setup. The VMs created by Vagrant aren't emulated, they're virtual machines which use hardware acceleration.

Once these changes pass and the L3 tests are fine, I'll make more changes.


dseevr:
build pr

Makefile
```diff
@@ -27,6 +27,7 @@ GOVET_CMD := go tool vet
 CI_HOST_TARGETS ?= "host-unit-test host-integ-test host-build-docker-image"
 SYSTEM_TESTS_TO_RUN ?= "00SSH|Basic|Network|Policy|TestTrigger|ACIM|Netprofile"
 ACI_GW_IMAGE ?= "contiv/aci-gw:04-12-2017.2.2_1n"
+L3_SYSTEM_TESTS_TO_RUN ?= "00SSH|Basic|Bgp|Network|Policy|TestTrigger|ACIM|Netprofile"
```
gkvijay: @unclejack, I think we should not filter the L3 tests. Instead we should run all the tests. Last time L3 tests failed because there was a change made that affected the docker service feature. L2 system tests in CI did not catch it as there were filters. Please run all the tests in L3 mode and that can be run once a day (not necessarily for commits)

unclejack: @gkvijay: Thank you for taking a look at this. I'll make sure to check if the tests pass now without filters.

unclejack:
@gkvijay: The BgpTriggerLinkUpDown test is failing now. Do you happen to know something about that? We've got to sort out that issue before we can enable all the tests for the L3 test run.

PR #941, opened 2017-07-31
Added locks that were missing in remoteEndpoint APIs.

gkvijay:
Change 8382c204 moved locking to the methods, except these new ones.

### DIFF
--- a/netplugin/plugin/netplugin.go
+++ b/netplugin/plugin/netplugin.go
@@ -151,11 +151,15 @@ func (p *NetPlugin) DeleteEndpoint(id string) error {
 
 // CreateRemoteEndpoint creates an endpoint for a given ID.
 func (p *NetPlugin) CreateRemoteEndpoint(id string) error {
+	p.Lock()
+	defer p.Unlock()
 	return p.NetworkDriver.CreateRemoteEndpoint(id)
 }
 
 // DeleteRemoteEndpoint destroys an endpoint for an ID.
 func (p *NetPlugin) DeleteRemoteEndpoint(id string) error {
+	p.Lock()
+	defer p.Unlock()
 	return p.NetworkDriver.DeleteRemoteEndpoint(id)
 }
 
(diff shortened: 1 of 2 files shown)


netplugin/plugin/netplugin.go
```diff
@@ -151,11 +151,15 @@ func (p *NetPlugin) DeleteEndpoint(id string) error {
 
 // CreateRemoteEndpoint creates an endpoint for a given ID.
 func (p *NetPlugin) CreateRemoteEndpoint(id string) error {
+	p.Lock()
```
gaurav-dalvi: Looks good to me.

dseevr:
@gaurav-dalvi Good catch on `AddPolicyRule`/`DelPolicyRule`.  `FetchNetwork` just returns an error (intentionally), so nothing needed there, and I think `Init` only gets called once.

One improvement that could be made to the file is to have the `Get*` and `Inspect*` functions use `RLock()` and `RUnlock()`, but that is a microoptimization

unclejack:
The use of RLock and RUnlock would be a good idea. This would allow us to improve performance a bit.

rchirakk:
LGTM

netplugin/plugin/netplugin.go
```diff
@@ -151,11 +151,15 @@ func (p *NetPlugin) DeleteEndpoint(id string) error {
...
+	defer p.Unlock()
 	return p.NetworkDriver.CreateRemoteEndpoint(id)
 }
 
 // DeleteRemoteEndpoint destroys an endpoint for an ID.
 func (p *NetPlugin) DeleteRemoteEndpoint(id string) error {
+	p.Lock()
```
abhinandanpb: Isnt the lock acquired here also ?  https://github.com/contiv/netplugin/blob/b6719ea52b1ee96cd13016a868ada555bfe545eb/netplugin/agent/state_event.go#L279

netplugin/plugin/netplugin.go
```diff
@@ -338,10 +342,14 @@ func (p *NetPlugin) DelSvcSpec(svcName string, spec *core.ServiceSpec) {
 
 // AddPolicyRule creates a policy rule
 func (p *NetPlugin) AddPolicyRule(id string) error {
+	p.Lock()
```
abhinandanpb: same here ? 
https://github.com/contiv/netplugin/blob/b6719ea52b1ee96cd13016a868ada555bfe545eb/netplugin/agent/state_event.go#L506

gkvijay: @abhinandanpb PTAL

PR #949, opened 2017-08-11
Added script to configure BGP for l3bgp-demo target.

gkvijay:
Makefile target for L3 mode demo setup without BGP.
make l3bgp-demo : sets up routing mode, configures BGP
make l3-demo : routing mode without BGP (no quagga vms)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -233,10 +233,16 @@ l3-test:
 	cd $(GOPATH)/src/github.com/contiv/netplugin/scripts/python && PYTHONIOENCODING=utf-8 ./createcfg.py -contiv_l3 2
 	CONTIV_L3=2 CONTIV_NODES=3 go test -v -timeout 900m ./test/systemtests -check.v -check.abort
 	CONTIV_L3=2 CONTIV_NODES=3 make stop
-l3-demo:
+
+#l3-demo setup for docker/swarm
+l3-demo: demo
+	vagrant ssh netplugin-node1 -c 'netctl global set --fwd-mode routing'
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -233,10 +233,15 @@ l3-test:
 	cd $(GOPATH)/src/github.com/contiv/netplugin/scripts/python && PYTHONIOENCODING=utf-8 ./createcfg.py -contiv_l3 2
 	CONTIV_L3=2 CONTIV_NODES=3 go test -v -timeout 900m ./test/systemtests -check.v -check.abort
 	CONTIV_L3=2 CONTIV_NODES=3 make stop
-l3-demo:
+
+l3-demo: demo
```
rchirakk: Could you add a line that it is for  docker/swarm  ?

scripts/l3bgp_demo.sh
```diff
@@ -0,0 +1,52 @@
+
```
rchirakk: it is better to keep this under docs dir.
And ```netctl``` can be moved to Makefile itself

Makefile
gkvijay: done

scripts/l3bgp_demo.sh
gkvijay: This is just a description of the l3bgp-demo setup. BGP documentation is already present in contiv doc/tutorial.

rchirakk:
LGTM

g1rana:
LGTM

PR #950, opened 2017-08-12
Dual stack support k8s + Fix systemtests to run IPv6 tests

yuva29:
#### Type of fix: New Feature + Bug

This PR contains the following:
- Enabled support for IPv6 in CNI plugin
- Identified and fixed system tests (IPv6 tests weren't running so far)

### DIFF
--- a/mgmtfn/k8splugin/driver.go
+++ b/mgmtfn/k8splugin/driver.go
@@ -48,9 +48,11 @@ type epSpec struct {
 
 // epAttr contains the assigned attributes of the created ep
 type epAttr struct {
-	IPAddress string
-	PortName  string
-	Gateway   string
+	IPAddress   string
+	PortName    string
+	Gateway     string
+	IPv6Address string
+	IPv6Gateway string
 }
 
 // netdGetEndpoint is a utility that reads the EP oper state
@@ -152,6 +154,7 @@ func createEP(req *epSpec) (*epAttr, error) {
 		return nil, err
 	}
 
+	// this response should contain IPv6 if the underlying network is configured with IPv6
 	log.Infof("Got endpoint create resp from master: %+v", mresp)
 
 	// Ask netplugin to create the endpoint
@@ -180,6 +183,11 @@ func createEP(req *epSpec) (*epAttr, error) {
 	epResponse.PortName = ep.PortName
 	epResponse.IPAddress = ep.IPAddress + "/" + strconv.Itoa(int(nw.SubnetLen))
 	epResponse.Gateway = nw.Gateway
+
+	if ep.IPv6Address != "" {
+		epResponse.IPv6Address = ep.IPv6Address + "/" + strconv.Itoa(int(nw.IPv6SubnetLen))
+		epResponse.IPv6Gateway = nw.IPv6Gateway
+	}
 
 	return &epResponse, nil
 }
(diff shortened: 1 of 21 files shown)


mgmtfn/dockplugin/netDriver.go
```diff
@@ -240,12 +241,24 @@ func createEndpoint(hostname string) func(http.ResponseWriter, *http.Request) {
...
+
 		epResponse := api.CreateEndpointResponse{
 			Interface: &api.EndpointInterface{
 				MacAddress: mresp.EndpointConfig.MacAddress,
 			},
 		}
 
+		// setup IPv6 address
+		if ep.IPv6Address != "" {
+			epResponse.Interface.AddressIPv6 = ep.IPv6Address + "/" + strconv.Itoa(int(nw.IPv6SubnetLen))
```
rchirakk: why do you need to set ipv6 address in response ? its already there in the request

mgmtfn/k8splugin/driver.go
```diff
@@ -448,7 +476,7 @@ func addPod(r *http.Request) (interface{}, error) {
 	}
 
 	// Set default gateway
-	err = setDefGw(pid, gw, gwIntf)
+	err = setDefGw(pid, gw, ep.IPv6Gateway, gwIntf)
```
rchirakk: you need to include ipv6 address/gwy in CNI response.

mgmtfn/dockplugin/netDriver.go
yuva29: It is not there in the request. Request only has IPv4

mgmtfn/k8splugin/driver.go
yuva29: I believe this is the response format expected from CNI https://github.com/contiv/netplugin/blob/master/mgmtfn/k8splugin/cniapi/api.go#L42-L48. Right now, it doesn't capture IPv6. Can the CNI response object be modified to accommodate more fields (IPv6 here)? Is it supported?

mgmtfn/dockplugin/netDriver.go
yuva29: @rchirakk https://github.com/docker/libnetwork/blob/master/docs/remote.md#create-endpoint

gkvijay: @yuva29. Thanks for taking care of the IPv6 changes. 
For Docker, can you update the docker network creation API with EnableIPv6 flag instead of returning the IPv6 address in the endpoint create response.
https://github.com/contiv/netplugin/blob/master/netmaster/docknet/docknet.go#L180

rchirakk: 👍

mgmtfn/k8splugin/driver.go
rchirakk: include ipv6 address/gateway in CNI response  https://github.com/contiv/netplugin/tree/master/mgmtfn/k8splugin/contivk8s

yuva29:
Updated the docker network creation and K8s response. PTAL @gkvijay @rchirakk

rchirakk:
@yuva29 
Changes looks good, couple more suggestions,
1. Can you use  cni.error/cni.Result to send status from netplugin to cni plugin 
2. Can you move CNI changes to another PR ?

yuva29:
@rchirakk Separated the docker changes into https://github.com/contiv/netplugin/pull/965. We need to pass the `cniVersion` which was not there in 0.1.0. Bumping the version will change the entire response format and other things. I feel that can go in a separate PR. Let me know WDYT?

rchirakk:
LGTM
Please create a new issue to track.

PR #956, opened 2017-08-22
remove  unsupported test cases from service  feature

g1rana:
…oviders
## Description of the changes
  remove  serviceSequenceAddDeleteProvider test  since it's not valid with current design.
#### Type of fix: 
    Bug Fix
#### Fixes #
    Fixes 908

- As per current design, service instances object should be created before adding service provider object. removing those test cases in which provider object is  created before service objects.

 - run L3-CI which is passed after those changes earlier it was failing all the time.
-  run make l3-test which is passed after this changes
  
TODO
 -documentation update under process


rchirakk:
LGTM

PR #957, opened 2017-08-23
Deferred cleanup of resources when a node crashes

### LINKED ISSUE #845: Garbage collect endpoints
In some cases, it is possible for a container to disappear without a notification to the network plugin. We need a way to garbage collect those endpoints. Preferably, there should be a manual way as well as a (conservative) automatic way to do this.

DivyaVavili:
#### Fixes #845 
Changes in the PR include:
- DeleteEndpoints API which takes in a hostname of the node crashed and cleans up the etcd endpoint cfg and oper state and calls ofnet API to clean up the flows of these endpoints on all other hosts in the cluster
- Vendoring in ofnet changes of ClearNode which clears up flow state of all endpoints in the host

#### TODO
- [ ] Cleanup ovs-driver oper state of the node (required exposing new APIs to netmaster. ovs-drive is currently accessed only by netplugin on the host)
- [ ] Test cases
- [ ] Have an option to configure the timer for starting cleanup of resources

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -635,23 +635,23 @@
 		},
 		{
 			"ImportPath": "github.com/contiv/ofnet",
-			"Rev": "b44a11b0bcffa08b3a7546e76fa2709457396867"
+			"Rev": "369938dc3d203fc4d35760bc96556441c467a7cc"
 		},
 		{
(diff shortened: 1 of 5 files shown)


rchirakk:
please create a new issue to address multiple GC timers being triggered on nodes.
LGTM

PR #977, opened 2017-09-18
refactor v2plugin Dockerfile & the main netplugin image Dockerfile

unclejack:
1. Migrates the main netplugin container from the Ubuntu base to alpine Linux. This should provide a more vendor independent base to make it more compatible with Ubuntu, CentOS and other distributions.

Fixes:
This addresses issues with running the container on CentOS 7.4 and later. The ubuntu image can't load kernel modules via insmod/modprobe when the kernel modules are compressed. This change was made in CentOS 7.4. This is issue #978

Advantages:
The size of the resulting image is ~100 MB. The previous image had a size of 300 MB. This decreases the requires storage and speeds up the download.

2. Refactors the v2plugin Dockerfile to make the image more like the main netplugin image. This removes the hacks which required glibc.

Fixes:
The third party untrusted binary package isn't needed any more.

Advantages:
The image is smaller, less likely to break in the future.

3. Refactors the code and the build.sh script to be able to produce static binaries. This enables us to not rely on the host's glibc any more. This allows us to remove the dependency on glibc.

Helps fix https://github.com/contiv/install/issues/248

### DIFF
--- a/install/v2plugin/Dockerfile
+++ b/install/v2plugin/Dockerfile
@@ -1,14 +1,26 @@
+##
+#Copyright 2017 Cisco Systems Inc. All rights reserved.
+#
+#Licensed under the Apache License, Version 2.0 (the "License");
+#you may not use this file except in compliance with the License.
+#You may obtain a copy of the License at
+#http://www.apache.org/licenses/LICENSE-2.0
+#
+#Unless required by applicable law or agreed to in writing, software
+#distributed under the License is distributed on an "AS IS" BASIS,
+#WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#See the License for the specific language governing permissions and
+#limitations under the License.
+##
+
 # Docker v2plugin container with OVS / netplugin / netmaster
 
-FROM alpine:3.5
+FROM alpine:3.6
 LABEL maintainer "Cisco Contiv (https://contiv.github.io)"
 
-RUN mkdir -p /run/docker/plugins /etc/openvswitch /var/run/contiv/log \
+RUN mkdir -p /run/docker/plugins /etc/openvswitch /var/log/contiv \
  && echo 'http://dl-cdn.alpinelinux.org/alpine/v3.4/main' >> /etc/apk/repositories \
- && apk --no-cache add openvswitch=2.5.0-r0 iptables ca-certificates openssl curl \
- && wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://raw.githubusercontent.com/andyshinn/alpine-pkg-glibc/master/sgerrand.rsa.pub \
- && wget https://github.com/andyshinn/alpine-pkg-glibc/releases/download/2.23-r1/glibc-2.23-r1.apk \
- && apk --no-cache add glibc-2.23-r1.apk
+ && apk --no-cache add openvswitch=2.5.0-r0 iptables ca-certificates openssl curl bash
 
 COPY netplugin netmaster netctl startcontiv.sh /
 
--- a/netplugin/netd.go
+++ b/netplugin/netd.go
@@ -19,8 +19,8 @@ import (
 	"log/syslog"
 	"net/url"
 	"os"
-	"os/user"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/contiv/netplugin/core"
@@ -181,8 +181,8 @@ func main() {
 	}
 
 	// Make sure we are running as root
-	usr, err := user.Current()
-	if (err != nil) || (usr.Username != "root") {
+	uid := syscall.Getuid()
+	if uid != 0 {
 		log.Fatalf("This process can only be run as root")
 	}
 
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -26,10 +26,10 @@ else
 fi
 
 echo $BUILD_VERSION >$VERSION_FILE
-
-GOGC=1500 go install \
+GOGC=1500 CGO_ENABLED=0 go install \
+	-a -installsuffix cgo \
 	-ldflags "-X $PKG_NAME.version=$BUILD_VERSION \
 	-X $PKG_NAME.buildTime=$BUILD_TIME \
 	-X $PKG_NAME.gitCommit=$GIT_COMMIT \
-	-s -w" \
+	-s -w -d" -pkgdir /tmp/foo-cgo \
 	-v $TO_BUILD
--- a/scripts/netContain/Dockerfile
+++ b/scripts/netContain/Dockerfile
@@ -13,19 +13,18 @@
 #limitations under the License.
 ##
 
-# One Container for OVS / netplugin / netmaster 
+# One Container for OVS / netplugin / netmaster
 
-FROM ubuntu:16.04
+FROM alpine:3.6
+LABEL maintainer "Cisco Contiv (https://contiv.github.io)"
 
-# Make sure to Modify the Proxy Server values if required 
+# Make sure to Modify the Proxy Server values if required
 # ENV export http_proxy=http://proxy.localhost.com:8080
 # ENV export https_proxy=https://proxy.localhost.com:8080
 
-RUN apt-get update \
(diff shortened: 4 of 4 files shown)


scripts/netContain/Dockerfile
```diff
@@ -13,19 +13,18 @@
 #limitations under the License.
 ##
 
-# One Container for OVS / netplugin / netmaster 
+# One Container for OVS / netplugin / netmaster
 
-FROM ubuntu:16.04
+FROM alpine:3.5
```
neelimamukiri: Any reason for not using the latest 3.6 version?

unclejack: @neelimamukiri: I was testing with 3.5 as with v2plugin. I need to check that it works properly.

rchirakk:
Can you verify k8s, v2plugin and any other deployments using this image & include it in this PR  ?
- [ ] k8s
- [ ] docker v2  plugin

unclejack:
I've updated this PR and the original description of the PR.

I've already tested the main netplugin image. I'll test the v2plugin image soon and comment here.

install/v2plugin/Dockerfile
```diff
@@ -1,14 +1,26 @@
...
-FROM alpine:3.5
+FROM alpine:3.6
 LABEL maintainer "Cisco Contiv (https://contiv.github.io)"
 
 RUN mkdir -p /run/docker/plugins /etc/openvswitch /var/run/contiv/log \
  && echo 'http://dl-cdn.alpinelinux.org/alpine/v3.4/main' >> /etc/apk/repositories \
- && apk --no-cache add openvswitch=2.5.0-r0 iptables ca-certificates openssl curl \
- && wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://raw.githubusercontent.com/andyshinn/alpine-pkg-glibc/master/sgerrand.rsa.pub \
- && wget https://github.com/andyshinn/alpine-pkg-glibc/releases/download/2.23-r1/glibc-2.23-r1.apk \
- && apk --no-cache add glibc-2.23-r1.apk
```
rhim: Just to make sure, glibc is not needed? If so, please ref. an issue in contiv/netplugin (create if it doesn't exist).

netplugin/netd.go
```diff
@@ -181,8 +181,8 @@ func main() {
 	}
 
 	// Make sure we are running as root
-	usr, err := user.Current()
-	if (err != nil) || (usr.Username != "root") {
+	uid := syscall.Getuid()
```
rhim: when netplugin runs as container, the UID = 0 may not mean the container has root privileges on the system. Is that a problem?

scripts/netContain/Dockerfile
```diff
@@ -13,19 +13,18 @@
 #limitations under the License.
 ##
 
-# One Container for OVS / netplugin / netmaster 
+# One Container for OVS / netplugin / netmaster
 
-FROM ubuntu:16.04
+FROM alpine:3.6
```
rhim: Please update associated issue with this requirement, and reference that issue in this PR.

install/v2plugin/Dockerfile
unclejack: v2plugin was installing glibc from a third party package before. The whole point of alpine is that it doesn't use, nor require glibc, for all the packages it provides.

The netplugin binaries were being built on ubuntu or on another host. Go does require a libc for some functions of the net packages. One way to have libc independent binaries is to compile those as static binaries (which don't include glibc).

This PR makes the following changes related to glibc and static binaries:
- makes the build.sh script build static binaries
- removes the glibc from the v2plugin Dockerfile and doesn't use it in the updated Dockerfile of the main Docker image

This last point is also mentioned here: https://github.com/contiv/netplugin/pull/977#issue-258457997

scripts/netContain/Dockerfile
unclejack: I'm not sure what issue I should point to here.

unclejack:
@rhim: Please let me know if you have other comments.

I'll test v2plugin and post another update.

unclejack:
I've looked into testing v2plugin and it seems this requires a bit more work. I've got to look into this a bit more to make sure v2plugin is ok as well.

unclejack:
I have tested v2plugin by hand with contiv/install and swarm. It seems healthy. All the daemons are running and netctl is creating networks properly. This is all exactly the same as the previous version (no errors, no failed tasks).

@rhim: Can we move this forward?

rhim:
This change is ready to move forward. It must be tested in nightly testing against a test build before the build is publicly released.

@contiv/committers-core-networking please review and merge.

scripts/netContain/Dockerfile
```diff
@@ -13,19 +13,18 @@
...
+# Make sure to Modify the Proxy Server values if required
 # ENV export http_proxy=http://proxy.localhost.com:8080
 # ENV export https_proxy=https://proxy.localhost.com:8080
 
-RUN apt-get update \
- && apt-get install -y openvswitch-switch=2.5.2-0ubuntu0.16.04.1 \
-        net-tools \
-        iptables \
- && rm -rf /var/lib/apt/lists/*
+RUN mkdir -p /run/docker/plugins /etc/openvswitch /var/run/contiv/log \
```
gkvijay: This line seems to be from v2plugin Dockerfile. You don't need the /run/docker/plugins/ directory or /var/run/contiv/log directory

install/v2plugin/Dockerfile
```diff
@@ -1,14 +1,26 @@
...
+#limitations under the License.
+##
+
 # Docker v2plugin container with OVS / netplugin / netmaster
 
-FROM alpine:3.5
+FROM alpine:3.6
 LABEL maintainer "Cisco Contiv (https://contiv.github.io)"
 
 RUN mkdir -p /run/docker/plugins /etc/openvswitch /var/run/contiv/log \
```
gkvijay: please change the /var/run/contiv/log to /var/log/contiv

unclejack:
@gkvijay: Thank you for the review. I've made the changes.

The jobs are running on the CI now. I've tested the v2plugin with swarm. I've also tested the netplugin container again with Kubernetes on contiv/install.

unclejack:
This needs to be merged.

rhim:
@unclejack can you trigger CI run for this PR please? As you know, the ngrok tunnel is disabled right now, so we need to trigger CI and update manually. Thanks.

scripts/build.sh
```diff
@@ -26,10 +26,10 @@ else
 fi
 
 echo $BUILD_VERSION >$VERSION_FILE
-
-GOGC=1500 go install \
+GOGC=1500 CGO_ENABLED=0 go install \
+	-a -installsuffix cgo \
 	-ldflags "-X $PKG_NAME.version=$BUILD_VERSION \
 	-X $PKG_NAME.buildTime=$BUILD_TIME \
 	-X $PKG_NAME.gitCommit=$GIT_COMMIT \
```
rchirakk: Why do you need to disable CGO ?

install/v2plugin/Dockerfile
```diff
@@ -1,14 +1,26 @@
...
+##
+
 # Docker v2plugin container with OVS / netplugin / netmaster
 
-FROM alpine:3.5
+FROM alpine:3.6
 LABEL maintainer "Cisco Contiv (https://contiv.github.io)"
 
-RUN mkdir -p /run/docker/plugins /etc/openvswitch /var/run/contiv/log \
+RUN mkdir -p /run/docker/plugins /etc/openvswitch /var/log/contiv \
```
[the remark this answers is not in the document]
unclejack: @rchirakk These were @gkvijay's comments related to my PR. I've made the changes and everything worked fine.

scripts/build.sh
unclejack: @rchirakk: This is needed to build static binaries which don't rely on glibc at all. We need static binaries which don't use cgo and any kind of features which require libc functions. This is something others have done as well.

dseevr: We do this on auth_proxy as well

unclejack:
build PR

Issue #962, opened 2017-08-29
Removal of Contiv network is possible even though there is a Docker Network attached to it


rkharya:
<img width="1490" alt="screen shot 2017-08-29 at 1 23 23 pm" src="https://user-images.githubusercontent.com/12057795/29810399-442ea55a-8cbd-11e7-81bb-f5a6242a9b0f.png">
## Description
v2Plugin on DEE 17.06 in native swarm mode, allows removal of Contiv network in L2/vLAN mode while its already attached with docker network by contiv-tag. 

#### Expected Behavior
It should throw an error while removing contiv network via `netctl network rm`. Removal of Contiv backend network leave Docker network with contiv driver in orphaned state and `docker service create` results in service deployment without any containers running. docker service throws not errors and it gets deployed. 

#### Observed Behavior
1).
RKHARYA-M-H07C:ucp-17.06-C rkharya$ docker network ls
NETWORK ID          NAME                                     DRIVER                  SCOPE
5a65d30ffbde        DEE-Ctrl-C-1.cisco.com/bridge            bridge                  local
fff4a7da2a01        DEE-Ctrl-C-1.cisco.com/docker_gwbridge   bridge                  local
8b98948f9e47        DEE-Ctrl-C-1.cisco.com/host              host                    local
573580b589d2        DEE-Ctrl-C-1.cisco.com/none              null                    local
75479d4151b8        DEE-Ctrl-C-2.cisco.com/bridge            bridge                  local
f7cc5018159e        DEE-Ctrl-C-2.cisco.com/docker_gwbridge   bridge                  local
762ae79b52eb        DEE-Ctrl-C-2.cisco.com/host              host                    local
ddf69843692a        DEE-Ctrl-C-2.cisco.com/none              null                    local
adbabf139804        DEE-Ctrl-C-3.cisco.com/bridge            bridge                  local
b8dbec7bb473        DEE-Ctrl-C-3.cisco.com/docker_gwbridge   bridge                  local
cc101f9652b0        DEE-Ctrl-C-3.cisco.com/host              host                    local
23405d0b69cb        DEE-Ctrl-C-3.cisco.com/none              null                    local
92ec4722a915        DEE-Wrk-C-1.cisco.com/bridge             bridge                  local
5200d9218cd0        DEE-Wrk-C-1.cisco.com/docker_gwbridge    bridge                  local
189906f6b69c        DEE-Wrk-C-1.cisco.com/host               host                    local
[...]

PR #986, opened 2017-09-26
v2plugin: add checks for delete network/epg

gkvijay:
In swarm-mode, add check for mapped docker network
when deleting netctl network or epg.

Fixes https://github.com/contiv/netplugin/issues/962

### DIFF
--- a/netmaster/docknet/docknet.go
+++ b/netmaster/docknet/docknet.go
@@ -287,6 +287,30 @@ func DeleteDockNetState(tenantName, networkName, serviceName string) error {
 	return dnetOper.Clear()
 }
 
+// GetDocknetState gets the docknet entry from state store
+func GetDocknetState(tenantName, networkName, serviceName string) (*DnetOperState, error) {
+	log.Infof("GetDocknetState for %s.%s.%s", tenantName, networkName, serviceName)
+
+	// Get the state driver
+	stateDriver, err := utils.GetStateDriver()
+	if err != nil {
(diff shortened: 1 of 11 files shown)


netmaster/docknet/docknet.go
```diff
@@ -287,6 +287,30 @@ func DeleteDockNetState(tenantName, networkName, serviceName string) error {
...
+		return nil, err
+	}
+
+	// save docknet oper state
+	dnetOper := DnetOperState{}
+	dnetOper.ID = fmt.Sprintf("%s.%s.%s", tenantName, networkName, serviceName)
+	dnetOper.StateDriver = stateDriver
+
+	// Read the dnet oper state
+	err = dnetOper.Read(dnetOper.ID)
```
rchirakk: Is this the correct way or walk thru dnet for matchin tenant/network/group ?
Also it is better to check "key not found" error here

gkvijay: Since we know tenant, network and group we can form the key and fetch that particular entry, instead of fetching all the entries and matching them.
The caller of the function is checking for 'key not found' in https://github.com/gkvijay/netplugin/blob/b6a3eb335de81f2b9e94f310735ee15f8d716963/netmaster/objApi/apiController.go#L995

unclejack:
2recheck

unclejack:
build PR

dseevr:
build PR

unclejack:
build PR

kahou82:
build PR

PR #995, opened 2017-10-06
netContain/Dockerfile: symlink /var/run to /run

unclejack:
This PR converts /var/run to symlink to /run. /var/run is a symlink to /run on the ubuntu image. The recent migration of this netplugin Docker image to the alpine base image has broken this assumption. The side effect of having /run and /var/run as independent directories is that contivk8s can't find its UNIX socket.

This PR solves the issue.

### DIFF
--- a/scripts/netContain/Dockerfile
+++ b/scripts/netContain/Dockerfile
@@ -24,7 +24,10 @@ LABEL maintainer "Cisco Contiv (https://contiv.github.io)"
 
 RUN mkdir -p /etc/openvswitch \
  && echo 'http://dl-cdn.alpinelinux.org/alpine/v3.4/main' >> /etc/apk/repositories \
- && apk --no-cache add openvswitch=2.5.0-r0 iptables ca-certificates openssl curl bash
+ && apk --no-cache add openvswitch=2.5.0-r0 iptables ca-certificates openssl curl bash \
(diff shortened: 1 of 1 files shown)


unclejack:
build PR

scripts/netContain/Dockerfile
```diff
@@ -24,7 +24,10 @@ LABEL maintainer "Cisco Contiv (https://contiv.github.io)"
 
 RUN mkdir -p /etc/openvswitch \
  && echo 'http://dl-cdn.alpinelinux.org/alpine/v3.4/main' >> /etc/apk/repositories \
- && apk --no-cache add openvswitch=2.5.0-r0 iptables ca-certificates openssl curl bash
```
rchirakk: I think a better fix will be to mount /run/contiv from installer ?

unclejack: @rchirakk: I've just reverted it to the way it was before the change to fix the issue with compressed kernel modules. We can fix it in various ways, but it'd be nice not having to update this in all the places (contiv/install, kubespray PR and perhaps other places).

This is a very simple change which lets us maintain backwards compatibility without new requirements.

unclejack:
The changes made to the Dockerfile were reverted. I'll close this now to send a different PR later.

PR #999, opened 2017-10-06
update deps for go-winio and remotessh

chrisplo:
go-winio has a bugfix to avoid building windows specific code on a
non-windows machine

remotessh: (only the last has any impact to this repo)
* migrate to vendor & update dependencies
* fix logrus import path to be uppercase
* Vagrantfile: bump Vagrant box 0.10.2
* check for empty env var for ssh command


dseevr:
Hmm, I don't see any changes for `MIcrosoft/go-winio` here...

vendor/github.com/contiv/remotessh/sshnode.go
```diff
@@ -93,7 +93,7 @@ func newCmdStrWithSource(cmd string, env []string) string {
 
 		if len(kv) == 2 {
 			envstr += fmt.Sprintf("%s=%q ", kv[0], kv[1])
-		} else if len(kv) == 1 {
+		} else if len(kv) == 1  && len(envvar) > 0 {
```
rchirakk: how can ```env``` variable contain whitespace ? 
can it be  trimmed while parsing & saving  ?

dseevr: That change was over here:  https://github.com/contiv/remotessh/pull/26

chrisplo: for first q: I had a single entry for environment, I think was empty string instead of space, but SplitN documents the behavior that I saw that if it can't split it, it just returns the original string, which is empty then you end up running something like `= my_command` which bash complains that `=` is a command not found

chrisplo: for the second, seemed easier to filter here than find all the callers (backtraced to vagrant.go Setup but stopped there) . . . are you asking if we should consider passing an empty line an error and not process?

chrisplo:
@dseevr reran make build, looks clear

unclejack:
build PR

chrisplo:
build env problem:

chrisplo:
build PR

vendor/github.com/contiv/remotessh/sshnode.go
rchirakk: not error out, but skip saving empty lines during init.
And this check can go to the beginning of the loop

```
for _, envvar := range env {
              if  len(envvar) > 0 {
                  .....
                   ...
              }
}
```

chrisplo: sorry, I don't know what you mean by saving lines, or what init you are referring to?  this code is skipping an empty input line, from your earlier comment, trimming would not change the string, it was already empty

it could go earlier, though probably better as if len(envar) == 0 {continue} to avoid the additional block though I prefer not to add extra block indents when possible, which is important with PEP8 formatting requirements, either way not sure it's worth repeating the approvals and testing?

rchirakk: >> mean by saving lines

env variables (multiple strings)  are saved in
```
type SSHNode struct {
	Name      string
	env       []string <------
	primaryIP net.IP
	sshAddr   string
	sshPort   string
	config    *ssh.ClientConfig
}
```
it is done from ``` NewSSHNode()```
I was suggesting if we could add code to skip```""``` in that function.
Anyway ```remotessh``` is used only in tests.
I think it is ok to merge

PR #1000, opened 2017-10-09
remove USE_RELEASE & remove release timestamp

unclejack:
This PR changes the USE_RELEASE logic to make it NIGHTLY_RELEASE. NIGHTLY_RELEASE can be used to make a release which has a timestamp appended to the version. Regular releases must not have this appended to the version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,7 +87,7 @@ checks: go-version gofmt-src golint-src govet-src misspell-src
 
 run-build: deps checks clean
 	cd $(GOPATH)/src/github.com/contiv/netplugin && \
-	USE_RELEASE=${USE_RELEASE} BUILD_VERSION=${BUILD_VERSION} \
+	NIGHTLY_RELEASE=${NIGHTLY_RELEASE} BUILD_VERSION=${BUILD_VERSION} \
 	TO_BUILD="${TO_BUILD}" VERSION_FILE=${VERSION_FILE} \
 	scripts/build.sh
 
@@ -335,5 +335,5 @@ clean-tar:
 release: tar
 	TAR_FILENAME=$(TAR_FILENAME) TAR_FILE=$(TAR_FILE) \
 	OLD_VERSION=${OLD_VERSION} BUILD_VERSION=${BUILD_VERSION} \
-	USE_RELEASE=${USE_RELEASE} scripts/release.sh
+	NIGHTLY_RELEASE=${NIGHTLY_RELEASE} scripts/release.sh
 	@make clean-tar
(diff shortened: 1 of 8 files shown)


unclejack:
This works properly. I'll trigger the build.

unclejack:
build PR

Makefile
```diff
@@ -87,7 +87,7 @@ checks: go-version gofmt-src golint-src govet-src misspell-src
 
 run-build: deps checks clean
 	cd $(GOPATH)/src/github.com/contiv/netplugin && \
-	USE_RELEASE=${USE_RELEASE} BUILD_VERSION=${BUILD_VERSION} \
+	NITHLY_RELEASE=${NIGHTLY_RELEASE} BUILD_VERSION=${BUILD_VERSION} \
```
chrisplo: misspelled:  NITHLY_RELEASE

scripts/release.sh
```diff
@@ -37,10 +36,8 @@ if [ "$OLD_VERSION" != "none" ]; then
 	comparison="$OLD_VERSION..HEAD"
 fi
 
-if [ "$USE_RELEASE" != "1" ]; then
-	echo "Making a pre-release.."
-	pre_release="-p"
-fi
+echo "Making a pre-release.."
+pre_release="-p"
```
chrisplo: this is only used once and never written to other than declaration, IMO, no variable needed

unclejack: @chrisplo: I've decided to leave it that way because we might want to add a variable to control that later. Would you like me to remove it?

Makefile
unclejack: @chrisplo: It's fixed now.

scripts/release.sh
chrisplo: not a blocker, but would be cleaner, if the time came where it needed to be variable, it can always come back

unclejack: My concern with removing that is it's going to be lost in the flags.

unclejack:
build PR

scripts/release.sh
chrisplo: adding a comment would be a nice place to call out that it's a pre-release, or if there is a long form --pre-release flag that would be even better

unclejack:
build PR

scripts/build.sh
```diff
@@ -9,10 +9,10 @@ if [ -n "$BUILD_VERSION" ]; then
 	VERSION=$BUILD_VERSION
 fi
 
-if [ -z "$USE_RELEASE" ]; then
-	BUILD_VERSION="$VERSION-$BUILD_TIME"
-else
+if [ -z "$NIGHTLY_RELEASE" ]; then
```
dseevr: Doesn't need to be changed because this script isn't setting it, but this would fail under `set -u`... the

```sh
if [ -z "${NIGHTLY_RELEASE:-}" ]; then`
```

form would need to be used to avoid tripping the unbound variable checker.

PR #1002, opened 2017-10-10
Enhance Dockerfile for local building

chrisplo:
Added a caching layer in the docker build for compiling all of the
vendor dependencies, drops compilation time to about 15s

Added support in Dockerfile for passing BUILD_VERSION and
NIGHTLY_RELEASE to build.sh

Add compile-with-docker target

Split out how we get the git commit SHA so it's reusable for docker
image tag


chrisplo:
* added "${BUILD_VERSION:-devBuild}-" to the tag, otherwise detail output below is the same:
examples:

BUILD_VERSION=1.2.3 make compile-with-docke
Successfully tagged netplugin:1.2.3-ecb31d87-unsupported

make compile-with-docker
Successfully tagged netplugin:devBuild-ecb31d87-unsupported

Detail with version info:
```
make compile-with-docker

scripts/build.sh
Version: 1.0.0-beta-10-10-2017.17-00-47.UTC
GitCommit: ecb31d8-unsupported
BuildTime: 10-10-2017.17-00-47.UTC

Successfully tagged netplugin:ecb31d87-unsupported
```

```
USE_RELEASE=y BUILD_VERSION=1.2.3 make compile-with-docker

Version: 1.2.3
GitCommit: ecb31d8-unsupported
BuildTime: 10-10-2017.17-01-36.UTC

Successfully tagged netplugin:ecb31d87-unsupported
```

```
USE_RELEASE=y make compile-with-docker

Version: 1.0.0-beta
GitCommit: ecb31d8-unsupported
BuildTime: 10-10-2017.17-02-52.UTC

Successfully tagged netplugin:ecb31d87-unsupported
```

```
BUILD_VERSION=1.2.3 make compile-with-docker

Version: 1.2.3-10-10-2017.16-58-15.UTC
GitCommit: ecb31d8-unsupported
BuildTime: 10-10-2017.16-58-15.UTC

Successfully tagged netplugin:ecb31d87-unsupported
```

chrisplo:
`make build` was successfull as well

chrisplo:
I pulled a compile target last minute as thought wasn't needed, need to restore, will add comments as well

chrisplo:
rechecked make `compile-with-docker` and `make build` and
```

**Comment by chrisplo** [comment]
build PR

**Comment by dseevr** [comment]
Now that https://github.com/contiv/netplugin/pull/999 is merged, you can remove that `grep -v` and such

File: scripts/getGitCommit.sh [hunk 1]
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
```
dseevr: Can we get a

```sh
set -euo pipefail
```

here?  I think we should include it in every new script by default.

scripts/getGitCommit.sh
```diff
@@ -0,0 +1,13 @@
...
+if command -v git &>/dev/null && git rev-parse &>/dev/null; then
+	GIT_COMMIT=$(git rev-parse --short HEAD)
+	if [ -n "$(git status --porcelain --untracked-files=no)" ]; then
+		GIT_COMMIT="$GIT_COMMIT-unsupported"
+	fi
+    echo $GIT_COMMIT
```
dseevr: Inconsistent indentation on lines 8 and 9 vs. 4-7

Check out [shfmt](https://github.com/mvdan/sh) if you want an easy way to automatically tidy up scripts.  We use this on auth_proxy and install

Dockerfile
```diff
@@ -28,14 +28,30 @@ FROM golang:1.7.6
...
+COPY ./ /go/src/github.com/contiv/netplugin/
 
-RUN make build
+ARG BUILD_VERSION=""
+ARG USE_RELEASE=""
 
+RUN GOPATH=/go/ \
+    BUILD_VERSION="${BUILD_VERSION}" \
+    USE_RELEASE="${USE_RELEASE}" \
+    make compile \
```
gkvijay: If you are compiling netplugin as part of docker build then you should add atleast checks here. Otherwise, we will end up not catching errors. If the idea is to make a quick build knowing that there are no code changes (for releases), then Dockerfile should only compile vendor and build the container. User can decide to skip tests when running the container using docker exec

chrisplo:
I will also check for changes due to PR1000 merging

scripts/getGitCommit.sh
chrisplo: wth is are tabs doing in there :/

Dockerfile
chrisplo: I have checks for docker based build as a backlog item but this isn't replacing other ways to build so it's not required (yet), and this PR is long enough already.

I was planning to have a separate Dockerfile that does the checks after compile which I think matches development flow better and will be faster to iterate.  Also, will allow choice as to when to run the checks based on which target is being run in the Makefile.

scripts/getGitCommit.sh
chrisplo: good catch, thx

dseevr:
build pr

chrisplo:
do you want me to pull apart the getGitCommit as separate PR @unclejack @dseevr ?

dseevr:
It's required for this PR to be merged and function so it seems fine to me as it is

chrisplo:
don't squash merge this please until after https://github.com/contiv/netplugin/pull/1006 lands

Makefile
```diff
@@ -86,23 +87,26 @@ endif
...
 	scripts/build.sh
 
+# fully prepares code for pushing to branch, includes building binaries
+run-build: deps checks clean compile
+
+compile-with-docker:
+	docker build \
+		--build-arg NIGHTLY_RELEASE=${NIGHTLY_RELEASE} \
+		--build-arg BUILD_VERSION=${BUILD_VERSION} \
+		-t netplugin:$${BUILD_VERSION:-devBuild}-$$(./scripts/getGitCommit.sh) .
```
chrisplo: devbuild not devBuild

dseevr:
build pr

dseevr:
@unclejack PTAL and merge if it looks good

Dockerfile
gkvijay: Please add the development Dockerfile here (root directory) and move the one with no checks to a release directory. Please take care of it in a later PR

Dockerfile
```diff
@@ -28,14 +28,27 @@ FROM golang:1.7.6
...
+RUN GOGC=1500 go install -ldflags "-s -w" $(go list ./vendor/...)
+
+# build the netplugin binaries
+COPY ./ /go/src/github.com/contiv/netplugin/
 
-RUN make build
+ARG BUILD_VERSION=""
+ARG NIGHTLY_RELEASE=""
 
+RUN GOPATH=/go/ \
```
unclejack: This Dockerfile should call `make deps` at some point to add the tools used to check if the source is fine to the image.

The executed command should include the checks target to run it before the build. This should reject code which doesn't meet the requirements.

Makefile
```diff
@@ -86,23 +87,26 @@ endif
...
+compile:
 	cd $(GOPATH)/src/github.com/contiv/netplugin && \
 	NIGHTLY_RELEASE=${NIGHTLY_RELEASE} BUILD_VERSION=${BUILD_VERSION} \
 	TO_BUILD="${TO_BUILD}" VERSION_FILE=${VERSION_FILE} \
 	scripts/build.sh
 
+# fully prepares code for pushing to branch, includes building binaries
+run-build: deps checks clean compile
+
+compile-with-docker:
```
unclejack: Could you set up these tasks like this, please?
`host-compile-with-docker:` -what we have in this target
`compile-with-docker:` - a wrapper which builds this in the Vagrant VM (e.g.: `vagrant ssh netplugin-node1 -c 'bash -lc "source /etc/profile.d/envvar.sh && cd /opt/gopath/src/github.com/contiv/netplugin && make host-compile-with-docker"'`)

It should be possible to invoke these targets on the host and on the VMs.

unclejack:
@chrisplo: Can you get rid of the merge commit and integrate the changes into their corresponding commits, please?

unclejack:
The merge commit is fine in this case. This particular PR will get squashed on merge.

PR #1003, opened 2017-10-11
Add .dockerignore for faster docker build times

chrisplo:
On OSX, the entire directory structure, including artifacts, are copied
into the docker machine every docker build.

This eliminates copying some large artifacts, git binary data, etc. to
save a few seconds each build attempt


unclejack:
build PR

PR #1008, opened 2017-10-14
'make tar' to use docker compiled assets

chrisplo:
`make tar` is primarily used for release, it is much faster to use compile-with-docker locally than spawn a VM to run the build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,3 @@
-
 .PHONY: all all-CI build clean default unit-test release tar checks go-version gofmt-src \
 	golint-src govet-src run-build compile-with-docker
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -313,11 +314,20 @@ host-plugin-release:
...
-tar: clean-tar
-	CONTIV_NODES=1 ${MAKE} build
-	@tar -jcf $(TAR_FILE) -C $(GOPATH)/src/github.com/contiv/netplugin/bin netplugin netmaster netctl contivk8s netcontiv -C $(GOPATH)/src/github.com/contiv/netplugin/scripts contrib/completion/bash/netctl -C $(GOPATH)/src/github.com/contiv/netplugin/scripts get-contiv-diags
+##########################
+## Packaging and Releasing
+##########################
+
+# build tarball
+tar: compile-with-docker
+	docker run --rm netplugin-build:$(NETPLUGIN_CONTAINER_TAG) \
```
unclejack: Please use `docker cp` to copy build artifacts.

chrisplo: there is no container to target docker cp

chrisplo: switching to docker create + docker cp + docker rm

chrisplo:
build PR

chrisplo:
this (and compile-with-docker) works in VM (docker 1.12) as well as desktop (docker 17.09), though without a private registry to cache things like the golang layer, the startup cost of compiling in docker on a fresh VM is large, better to run this target locally.

PR #1010, opened 2017-10-14
Makefile: print the list of system tests to be run

unclejack:
This PR adds one echo statement to print the list of system tests which are about to be run. The absence of this line can give the impression that the tests aren't run. This should help make it easier to understand how the tests work.


unclejack:
build PR

PR #1012, opened 2017-10-16
Build checks in container

chrisplo:
Before the only option was to spin up a VM to run code checks, which is
expensive in terms of time.

Add Dockerfile-check to provide a container build for a runtime
environment that supports go tool vet, gofmt, and golint

Add script 'code_checks_in_docker.sh' intended to be run inside the
Dockerfile-check container that runs `make checks` which in turn runs a
variety of go code checks against packages passed in.

Checks are reordered so more severe issues (e.g. functional) are
resolved before less severe ones (e.g. formatting).

Some checks are upated to be simpler and faster because the command
accept multiple directories and recurse automatically.

Drive-by:
* removed some trailing whitespace
* found another binary to exclude in dockerignore
* variables for go check commands not needed, they are only refernced
  once and require a visual lookup to find what the command actually
  is

### DIFF
--- a/Dockerfile-check
+++ b/Dockerfile-check
@@ -1,0 +1,10 @@
+FROM golang:1.7.6
+
+ENV GOPATH=/go
+
+WORKDIR /go/src/github.com/contiv/netplugin/
+
+RUN go get github.com/golang/lint/golint \
+           github.com/client9/misspell/cmd/misspell
+
+CMD ["make", "checks"]
--- a/Makefile
+++ b/Makefile
@@ -23,9 +23,6 @@ TAR_FILE := $(TAR_LOC)/$(TAR_FILENAME)
 GO_MIN_VERSION := 1.7
 GO_MAX_VERSION := 1.8
 GO_VERSION := $(shell go version | cut -d' ' -f3 | sed 's/go//')
-GOLINT_CMD := golint -set_exit_status
-GOFMT_CMD := gofmt -s -l
-GOVET_CMD := go tool vet
 CI_HOST_TARGETS ?= "host-unit-test host-integ-test host-build-docker-image"
 SYSTEM_TESTS_TO_RUN ?= "00SSH|Basic|Network|Policy|TestTrigger|ACIM|Netprofile"
 K8S_SYSTEM_TESTS_TO_RUN ?= "00SSH|Basic|Network|Policy"
@@ -63,19 +60,20 @@ godep-restore:
 
 gofmt-src:
 	$(info +++ gofmt $(PKG_DIRS))
-	@for dir in $(PKG_DIRS); do $(GOFMT_CMD) $${dir} | grep "go"; [[ $$? -ne 0 ]] || exit 1; done
+	@[[ -z "$$(gofmt -s -d $(PKG_DIRS) | tee /dev/stderr)" ]] || exit 1
 
+# go lint does not automatically recurse
 golint-src:
 	$(info +++ golint $(PKG_DIRS))
-	@for dir in $(PKG_DIRS); do $(GOLINT_CMD) $${dir}/... || exit 1;done
+	@for dir in $(PKG_DIRS); do golint -set_exit_status $${dir}/...; done
 
 govet-src:
 	$(info +++ govet $(PKG_DIRS))
-	@for dir in $(PKG_DIRS); do $(GOVET_CMD) $${dir} || exit 1;done
+	@go tool vet $(PKG_DIRS)
 
 misspell-src:
 	$(info +++ check spelling $(PKG_DIRS))
-	misspell -locale US -error $(PKG_DIRS)
+	@misspell -locale US -error $(PKG_DIRS)
 
 go-version:
 	$(info +++ check go version)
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -313,11 +320,20 @@ host-plugin-release:
...
-	@tar -jcf $(TAR_FILE) -C $(GOPATH)/src/github.com/contiv/netplugin/bin netplugin netmaster netctl contivk8s netcontiv -C $(GOPATH)/src/github.com/contiv/netplugin/scripts contrib/completion/bash/netctl -C $(GOPATH)/src/github.com/contiv/netplugin/scripts get-contiv-diags
+##########################
+## Packaging and Releasing
+##########################
+
+# build tarball
+tar: compile-with-docker
+	docker run --rm netplugin-build:$(NETPLUGIN_CONTAINER_TAG) \
+		-C /go/src/github.com/contiv/netplugin netplugin-version \
+	| tar -x netplugin-version
```
unclejack: Please use docker cp here. Piping can break if there's any Docker regression.

Makefile
```diff
@@ -313,11 +320,20 @@ host-plugin-release:
...
+	docker run --rm netplugin-build:$(NETPLUGIN_CONTAINER_TAG) \
```
unclejack: This is the same as above - it should be done using `docker cp`.

scripts/code_checks.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
unclejack: Please keep using the Makefile checks for now. This script would get blown away soon.

Dockerfile-check
```diff
@@ -0,0 +1,15 @@
+ARG TAG=latest
+
+FROM golang:1.7.6
+
+ENV GOPATH=/go
+
+WORKDIR /go/src/github.com/contiv/netplugin/
+ENTRYPOINT ["/code_checks.sh"]
+
+RUN go get github.com/tools/godep github.com/aktau/github-release \
```
unclejack: Please use https://github.com/contiv/netplugin/blob/master/scripts/deps by invoking the right Makefile target.

Makefile
chrisplo: From the docs: `docker cp [OPTIONS] CONTAINER:SRC_PATH DEST_PATH|-`

At this point there has never been a container, only an image, so there is no way I could use docker cp.

unclejack: @chrisplo: You can use `docker create` to create a container without running. You can remove it with `docker rm`.

Relying on the stdout isn't recommended. There have been regressions in the past with relying on piping from stdout and piping via stdin.

chrisplo: cool, I'll update with create/cp

Dockerfile-check
```diff
@@ -0,0 +1,15 @@
+ARG TAG=latest
```
dseevr: General comments on this file:

We do something similar already in auth_proxy, take a look at [Dockerfile.checks](https://github.com/contiv/auth_proxy/blob/6340df9b07a1723e2419069ebb5c0d5be6850a2e/build/Dockerfile.checks)

In general, we want to `COPY` in the code to be checked as one of the last operations in the Dockerfile (for caching reasons).  Bindmounting + SELinux = nightmare land, and it also prevents the use of `docker-machine` in general

scripts/code_checks.sh
dseevr: Unrelated to whatever script we end up using, these tools are _slooooooow_ unless you exclude `vendor`.  PTAL at https://github.com/contiv/auth_proxy/blob/6340df9b07a1723e2419069ebb5c0d5be6850a2e/scripts/checks_in_container.sh#L10-L11

Dockerfile-check
chrisplo: COPY is at the end, though this ARG was leftover from an earlier incantation (for FROM) I'll remove

dseevr: Right, you're `COPY`ing in the script, that's fine...

What I'm saying is we need to `COPY` in all the code to be checked, e.g.,:

```Dockerfile
COPY ./ /go/src/github.com/contiv/netplugin
```

We shouldn't use a bindmount on line 95 in Makefile

scripts/code_checks.sh
chrisplo: agreed, I borrowed PKG_DIRS which already filtered out vendor
EXCLUDE_DIRS := bin docs Godeps scripts vagrant vendor install
PKG_DIRS := $(filter-out $(EXCLUDE_DIRS),$(subst /,,$(sort $(dir $(wildcard */)))))

Makefile
chrisplo: actually did this in #1008, and removed this section as we don't depend on it for this PR

Makefile
chrisplo: did this in #1008, and removed this section as we don't depend on it for this PR

Dockerfile-check
```diff
@@ -0,0 +1,12 @@
+FROM golang:1.7.6
+
+ENV GOPATH=/go
+
+WORKDIR /go/src/github.com/contiv/netplugin/
+
+RUN go get github.com/tools/godep github.com/aktau/github-release \
```
dseevr: Is `github.com/aktau/github-release` here intentionally? 😄

chrisplo: was copied, removed!

dseevr:
@unclejack PTAL and merge if it looks good to you

chrisplo:
build pr

rchirakk:
LGTM, thx for fixing vet/lint/fmt.
another option for quick build/check is "make run-build"

chrisplo:
build PR

chrisplo:
build PR

PR #1018, opened 2017-10-17
Adding CPU and memory settings for k8s Vagrantfile

vhosakot:
This PR adds:
- CPU and memory variables for the k8s Vagrantfile so that they can be passed at runtime.
- Documentation for the user about how to pass variables to Makefile when deploying Contiv.

Tested on Mac laptop.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Optionally, variables can be passed to Makefile if needed. For example, to
 use 1GB memory for the vagrant VMs, run:
 
 ```
-CONTIV_MEMORY=1024 make demo
+CONTIV_MEMORY=4096 make demo
 ```
 
 #### Step 2: Create a network
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ $ vagrant ssh netplugin-node1
 ```
 
 Optionally, variables can be passed to Makefile if needed. For example, to
-use 1GB memory for the vagrant VMs, run:
+use 4 GB memory for the vagrant VMs, run:
 
 ```
 CONTIV_MEMORY=4096 make demo
--- a/README.md
+++ b/README.md
@@ -39,11 +39,14 @@ $ vagrant ssh netplugin-node1
 ```
 
 Optionally, variables can be passed to Makefile if needed. For example, to
-use 4 GB memory for the vagrant VMs, run:
+use 4 GB memory and 2 CPUs for the vagrant VMs, run:
 
 ```
-CONTIV_MEMORY=4096 make demo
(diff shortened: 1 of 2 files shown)


vhosakot:
build PR

README.md
```diff
@@ -38,6 +38,13 @@ $ cd netplugin; make demo
 $ vagrant ssh netplugin-node1
 ```
 
+Optionally, variables can be passed to Makefile if needed. For example, to
+use 1GB memory for the vagrant VMs, run:
+
+```
+CONTIV_MEMORY=1024 make demo
```
unclejack: Can you change this to 3072 or 4096, please? Using VMs with 1 GB of RAM will most likely be something we won't support officially. The person who tries this would get to keep all the broken pieces if the VMs with 1 GB of RAM break the environment.

vhosakot: Done, see latest commit. Do we need a note in the docs to let the user know that 1 GB of RAM must be avoided for the vagrant VMs as it breaks Contiv?

vhosakot:
build PR

vhosakot:
build PR

README.md
```diff
@@ -38,6 +38,13 @@ $ cd netplugin; make demo
 $ vagrant ssh netplugin-node1
 ```
 
+Optionally, variables can be passed to Makefile if needed. For example, to
+use 4 GB memory for the vagrant VMs, run:
+
+```
+CONTIV_MEMORY=4096 make demo
+```
+
```
chrisplo: did you forget CONTIV_CPUS ?  would be nice to show the defaults as well in the docs

chrisplo: not a big deal, but CONTIV_MEMORY_MB would be clearer on the unit

dseevr: If we change this, we should also change all the other Vagrantfiles to keep them consistent

vhosakot: > did you forget CONTIV_CPUS ? would be nice to show the defaults as well in the docs

@chrisplo Added CONTIV_CPUS too in the latest commit.  Also added the default values for `CONTIV_CPUS` and `CONTIV_MEMORY` in `README.md`.

> not a big deal, but CONTIV_MEMORY_MB would be clearer on the unit

@chrisplo - As @dseevr said, `CONTIV_MEMORY ` matches and is consistent with the variable name for Docker swarm in https://github.com/contiv/netplugin/blob/master/Vagrantfile#L33. `CONTIV_MEMORY` for both Docker swarm and k8s can be renamed to `CONTIV_MEMORY_MB` in the future if needed.

dseevr: Good idea, we can hit all of them at once in another PR if desired

PR #1019, opened 2017-10-17
Check if GOPATH variable is empty in netplugin/Makefile

vhosakot:
When we **DO NOT** set `GOPATH`, `netplugin/Makefile` uses an empty `GOPATH` variable without setting it. Hence, when we run `make k8s-test`, the line https://github.com/contiv/netplugin/blob/master/Makefile#L164 fails with the following error:
```
cd /src/github.com/contiv/netplugin/scripts/python && PYTHONIOENCODING=utf-8 ./createcfg.py -scheduler 'k8s' -binpath contiv/bin -install_mode 'kubeadm'

/bin/bash: line 0: cd: /src/github.com/contiv/netplugin/scripts/python: No such file or directory
```

This PR checks if `GOPATH` variable is empty in `netplugin/Makefile` and exits with an error if `GOPATH` is unset.

Tested on Mac laptop.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,7 @@
+ifeq ($(GOPATH),)
+$(error GOPATH is not set. Set GOPATH by doing "export GOPATH=<your GOPATH>".)
+endif
+
 .PHONY: all all-CI build clean default unit-test release tar checks go-version gofmt-src \
 	golint-src govet-src run-build compile-with-docker
 


vhosakot:
build PR

unclejack:
This is wrong. All the targets have failed on the CI.

unclejack:
The GOPATH environment variable must be set before invoking the Makefile. This is related to how Go development is done and it's something which applies to other targets as well, not only to this target.

vhosakot:
build PR

vhosakot:
> if GOPATH is not in the environment, which we have no certainty is correct
@chrisplo When `GOPATH` is not set in the environment, I see `go env GOPATH` shows the right path:
```
$ unset $GOPATH
$ echo $GOPATH

$ go env GOPATH
/Users/vhosakot/go
$ go env | grep GOPATH
GOPATH="/Users/vhosakot/go"
```

chrisplo:
you just happened to make your GOPATH ~/go, that's not valid for my environment for example . . looked up an example of bailing on unset variable:
```

**Comment by vhosakot** [comment]
build PR

**Comment by vhosakot** [comment]
@chrisplo @unclejack @dseevr How does this look now? I've tested for all cases for `GOPATH` - set, unset, default, non-default, export, non-export(local).

**Comment by vhosakot** [comment]
build PR

**Comment by vhosakot** [comment]
build PR

File: Makefile [hunk 1]
```diff
@@ -1,3 +1,7 @@
+ifeq ($(GOPATH),)
+$(error GOPATH is not set. Set GOPATH by doing "export GOPATH=<your GOPATH>".)
```
rchirakk: you need to update $PATH to include $GOPATH/bin for go tools to work.

vhosakot: sure @rchirakk , I'll address it in a separate PR as it is not related to the issue fixed by this PR.

vhosakot: sure @rchirakk , I'll address it in a separate PR as it is not related to the issue fixed by this PR.

PR #1020, opened 2017-10-17
Adding debugs in netplugin/scripts/build.sh

vhosakot:
When we run `make k8s-test`, the `compile` target takes a long time to run `scripts/build.sh` (https://github.com/contiv/netplugin/blob/master/Makefile#L94) as it builds the entire netplugin code in `netplugin, netmaster, netctl, mgmtfn/k8splugin and mgmtfn/mesosplugin`. It looks as if scripts/build.sh **hangs** for a long time when we run `make k8s-test`.

This PR adds debugs in `netplugin/scripts/build.sh` so that the user sees the name of each go file and the package name compiled in the output when netplugin is being built and does not think that `scripts/build.sh ` hangs for a long time when building netplugin.

Tested on Mac laptop.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -euxo pipefail
+
 BUILD_TIME=$(date -u +%m-%d-%Y.%H-%M-%S.UTC)
 VERSION=$(cat version/CURRENT_VERSION | tr -d '\n')
 PKG_NAME=github.com/contiv/netplugin/version
@@ -19,7 +21,7 @@ GIT_COMMIT=$(./scripts/getGitCommit.sh)
 
 echo $BUILD_VERSION >$VERSION_FILE
 
-GOGC=1500 go install \
+GOGC=1500 go install -v \
 	-ldflags "-X $PKG_NAME.version=$BUILD_VERSION \
 	-X $PKG_NAME.buildTime=$BUILD_TIME \
 	-X $PKG_NAME.gitCommit=$GIT_COMMIT \


vhosakot:
build PR.

vhosakot:
build PR

scripts/build.sh
```diff
@@ -19,7 +21,7 @@ GIT_COMMIT=$(./scripts/getGitCommit.sh)
 
 echo $BUILD_VERSION >$VERSION_FILE
 
-GOGC=1500 go install \
+GOGC=1500 go install -x -v \
```
unclejack: Please revert this change. `-v` was recently removed.

vhosakot: I see both `-x` and `-v` as go's build options in the latest go (both in the docs as well as in go's code in master):

Docs for `-x` and `-v`: https://golang.org/cmd/go/#hdr-Compile_packages_and_dependencies

`-x` in Go's code:
https://github.com/golang/go/blob/master/src/cmd/go/internal/work/build.go#L91-L92

`-v` in Go's code:
https://github.com/golang/go/blob/master/src/cmd/go/internal/work/build.go#L86-L87
```
$ go version
go version go1.9.1 darwin/amd64

$ go help build | grep -A 1 '\-x\|\-v'
	-v
		print the names of packages as they are compiled.
	-x
		print the commands.
```

chrisplo: I think you can get a better sense of time spent compiling vendor packages and netplugin code with `make compile-with-docker`, the time compiling isn't really that long, maybe there are other things happening during k8s-test?  Not sure we want to spam so much build time info for everyone, if there is a compile problem, there will be output.

unclejack: @vhosakot: The fact that Go supports both flags isn't relevant. `-x` would be far too verbose and that's not needed.

I agree only with bringing back `-v` and nothing else - no other build options.

rchirakk: Can you change it to do verbose build based on the command line option & 
keep the non-verbose  build by default ?

unclejack: @rchirakk: `-v` is the only acceptable option here and it should be the default. `-x` is excessive.

rchirakk: It looks like  -v was removed recently.
yes, please  add ```-v``` option & trigger the build.

vhosakot:
build PR

vhosakot:
@chrisplo @unclejack @rchirakk I removed `-x` and just kept `-v`. `-v` is not excessive and will not spam the user when building netplugin.

Yes, I see https://github.com/contiv/netplugin/commit/5732e5366c268dfe0beb20c02359de69e78105b7#diff-d99b26f864526f6f8f79861d49deb922L35 removed `-v` and this PR adds `-v` back so that the user sees what is built without spam in the output.

I've squashed all the commits into a single commit and re-trigerred the build.

dseevr:
@rchirakk PTAL again

chrisplo:
the set -x is chatty, but right now I think oversharing is OK till we get stability

PR #1025, opened 2017-10-19
image tag is just sha, rm 'devbuild' default

chrisplo:
'devbuild' moniker is not needed, that is a contiv/install default
convention

Simplify the tag name to just the SHA, which ties the image tag to the
actual contents, and not be variable based on a build time environment variable

* Added some comments
* compile-with-docker was not using the tag variable

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,8 @@
+# make BUILD_VERSION=1.2.3 [compile-with-docker] will set netplugin -version output
+# make BUILD_VERSION=1.2.3 tar will set the tar filename
+# default naming will otherwise be value of version/CURRENT_VERSION
+
+
 .PHONY: all all-CI build clean default unit-test release tar checks go-version gofmt-src \
 	golint-src govet-src run-build compile-with-docker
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -19,8 +23,7 @@ NAME := netplugin
 VERSION_FILE := $(NAME)-version
 VERSION := `cat $(VERSION_FILE)`
 TAR_EXT := tar.bz2
-# BUILD_VERSION=1.2 make --> 1.2-1097d2a7, otherwise devbuild-1097d2a7
-export NETPLUGIN_CONTAINER_TAG := $(or $(BUILD_VERSION),devbuild)-$(shell ./scripts/getGitCommit.sh)
+NETPLUGIN_CONTAINER_TAG := $(shell ./scripts/getGitCommit.sh)
```
tiewei: IMO instead of using the script to determine the version, we can simply uses `$(git describe --tags --always)` to get the tag, the format will be 
`<latest-tag>-<commits-after-tag>-g<sha>`, if it's right on tag, it will be `<tag>`. this contains commit sha so should be enough to identify the change

chrisplo: we should update the script (it's used in multiple places), that can be separate PR

dseevr:
build pr

PR #1026, opened 2017-10-20
Removed GOPATH check

vhosakot:
Not all Makefile targets need `GOPATH` to be set. Makefile targets like `demo` and `demo-v2plugin` do not need `GOPATH` to be set in order to sync the directory `src/github.com/contiv/netplugin` on the laptop with `/opt/gopath` inside the vagrant VM (https://github.com/contiv/netplugin/blob/master/Vagrantfile#L370). Targets like `tar` and `compile-with-docker` also do not need `GOPATH`.

Hence, this PR removes the check for `GOPATH` in the Makefile and adds documentation in `README.md` for the user to set `GOPATH` when using Makefile targets like `k8s-test`, `system-test`, `clean` and `compile` that **do need** `GOPATH` to be set (https://github.com/contiv/netplugin/blob/master/Makefile#L99).

Tested on Mac laptop.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,7 @@ that lives behind an OVS bridge and has its own unique interfaces.
 #### Step 1: Clone the project and bringup the VMs
 
 Note: if you have $GOPATH set, then please ensure either you unset GOPATH,
-or clone the tree in `$GOPATH/src/github.com/contiv/` location. If you are
-using a Makefile target that needs `GOPATH`, please set it by doing 
-`export GOPATH=<your GOPATH>`.
+or clone the tree in `$GOPATH/src/github.com/contiv/` location
 
 ```
 $ git clone https://github.com/contiv/netplugin
--- a/README.md
+++ b/README.md
@@ -30,7 +30,9 @@ that lives behind an OVS bridge and has its own unique interfaces.
 #### Step 1: Clone the project and bringup the VMs
 
 Note: if you have $GOPATH set, then please ensure either you unset GOPATH,
-or clone the tree in `$GOPATH/src/github.com/contiv/` location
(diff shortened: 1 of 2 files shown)


vhosakot:
build PR

delapsley:
Thanks for doing this @vhosakot. I did things slightly differently and it seemed to work:

- cd ~/repos
- git clone git@github.com:contiv/netgplugin
- cd netplugin
- export GOPATH=`pwd`
- make demo

That seemed to work.

README.md
```diff
@@ -30,7 +30,9 @@ that lives behind an OVS bridge and has its own unique interfaces.
 #### Step 1: Clone the project and bringup the VMs
 
 Note: if you have $GOPATH set, then please ensure either you unset GOPATH,
-or clone the tree in `$GOPATH/src/github.com/contiv/` location
+or clone the tree in `$GOPATH/src/github.com/contiv/` location. If you are
+using a Makefile target that needs `GOPATH`, please set it by doing
```
rchirakk: I don't think you need this change, it is covered by the first line.

vhosakot: @rchirakk The first line is not clear IMO. It does not clearly say "you **MUST** set `GOPATH` for certain targets". Hence, @delapsley and I both saw issues without `GOPATH` as there is no check for empty `GOPATH` in `Makefile`. Since there is no check for empty `GOPATH`, I added this clear note for the user.

vhosakot:
@rchirakk I removed the note in the documentation.  FYI, when `GOPATH` is not set, there is no check in the code now and no docs too about it and the build does *not* fail fast. I still think the documentation is unclear 😄 because of which @delapsley and I saw the same issue when `GOPATH` is not set.

vhosakot:
build PR

PR #1028, opened 2017-10-21
Ensure ovs logging directory created before start service

tiewei:
Without this patch, when a node first time start netplugin, and if
there wasn't netmaster started ever on the node, the ovs logging
directory will not exist and no logging will be able to be found.
This patch makes sure the ovs logging directory created before
service starts.

### DIFF
--- a/scripts/netContain/scripts/ovsInit.sh
+++ b/scripts/netContain/scripts/ovsInit.sh
@@ -6,31 +6,52 @@ set -euo pipefail
 modprobe openvswitch
 
 mkdir -p /var/run/openvswitch
+mkdir -p /var/contiv/log/
 
 sleep 2
 
 if [ -d "/etc/openvswitch" ]; then
-	if [ -f "/etc/openvswitch/conf.db" ]; then
-		echo "The Open vSwitch database exists"
-	else
-		echo "The Open VSwitch database doesn't exist"
-		echo "Creating the Open VSwitch database..."
-		ovsdb-tool create /etc/openvswitch/conf.db /usr/share/openvswitch/vswitch.ovsschema
-	fi
+    if [ -f "/etc/openvswitch/conf.db" ]; then
+        echo "INFO: The Open vSwitch database exists"
+    else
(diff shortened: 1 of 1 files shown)


scripts/netContain/scripts/ovsInit.sh
```diff
@@ -27,6 +28,8 @@ ovsdb-server --remote=punix:/var/run/openvswitch/db.sock --remote=db:Open_vSwitc
 echo "Starting ovs-vswitchd"
 ovs-vswitchd -v --pidfile --detach --log-file=/var/contiv/log/ovs-vswitchd.log -vconsole:err -vsyslog:info -vfile:info &
 
+sleep 2
```
chrisplo: wait for port open?

g1rana:
LGTM

scripts/netContain/scripts/ovsInit.sh
dseevr: Is there a simple `ovs-vsctl` command we can run here to poll instead of sleeping?

tiewei: done

scripts/netContain/scripts/ovsInit.sh
```diff
@@ -27,6 +28,11 @@ ovsdb-server --remote=punix:/var/run/openvswitch/db.sock --remote=db:Open_vSwitc
 echo "Starting ovs-vswitchd"
 ovs-vswitchd -v --pidfile --detach --log-file=/var/contiv/log/ovs-vswitchd.log -vconsole:err -vsyslog:info -vfile:info &
 
+while [[ $(ovsdb-client list-dbs | grep -c Open_vSwitch) -eq 0 ]] ; do
+    echo "Wait for ovsdb started"
```
dseevr: Can we change this to "Waiting for ovsdb to start"?

Also, should we fail after N attempts?  We do that in other places where we poll for readiness

tiewei: done

dseevr:
build pr

PR #1030, opened 2017-10-25
Subtree objdb repo into netplugin

dseevr:
This is step 1 of folding the objdb repo into netplugin as a Go package.  The objdb repo was subtree'd in and its Godeps/vendor stuff was removed.  (checks.sh had to be removed because one of netplugin's code quality tools was complaining about it)

Once this is in place, contivmodel can be updated to use this package.  Once that is done, netplugin itself can be updated to use this package.


dseevr:
build pr

dseevr:
build PR

PR #1033, opened 2017-10-27
Godeps/Godeps.json: update client-go version

dseevr:
There are no vendored changes with this commit.  This is necessary to resolve an issue with Godeps where it barfs on dependency detection because "client-go" had an empty Godeps.json


dseevr:
build pr

PR #1040, opened 2017-10-31
Build versioned v2plugin from versioned binaries

chrisplo:
This enables build and caching locally or in a build machine the
v2plugin before creating test environment VMs and failing much faster
on any PR issues, which is good for CI turnaround.

v2plugin_rootfs (called by host-pluginfs-create) builds a docker
container runtime with a versioned archive of netplugin binaries, then
extracts the entire filesystem from the container, which is now left as
a versioned tarball.

The versioned archive of netplugin binaries used for v2plugin can now be
created in two ways:
* run-build (dependency for demo-v2plugin) will compile then version an
  archive with built assets (runs only on the VM)
* tar target (locally or in VM)

host-pluginfs-create can be run in the VM or locally

Leaving the plugin rootfs as a tarball has a few advantages reasons:
* docker in a VM had internal issues when creating the plugin when the
  rootfs was unarchived on the host and exposed though a virtualbox
  mount, at least on OS X
* the rootfs is versioned

This adjusted host-plugin-release and demo-v2plugin to both use a new
target 'host-pluginfs-unpack'

Drive-by:
* Fix Dockerfile by adding objdb which is needed by vendored packages
* Fix config.template for the CONTIV_V2PLUGIN_NAME replacement


chrisplo:
checking on a minor issue during testing

chrisplo:
https://github.com/contiv/netplugin/pull/1042 helps to sort out version naming issues

Makefile
```diff
@@ -129,7 +134,7 @@ update:
 # setting CONTIV_NODES=<number> while calling 'make demo' can be used to bring
 # up a cluster of <number> nodes. By default <number> = 1
 start:
-	CONTIV_DOCKER_VERSION="$${CONTIV_DOCKER_VERSION:-$(DEFAULT_DOCKER_VERSION)}" CONTIV_NODE_OS=${CONTIV_NODE_OS} vagrant up
```
unclejack: Why are you making this change here? It seems unrelated to the purpose of the PR.

Makefile
```diff
@@ -6,7 +6,6 @@
 .PHONY: all all-CI build clean default unit-test release tar checks go-version gofmt-src \
 	golint-src govet-src run-build compile-with-docker
 
-DEFAULT_DOCKER_VERSION := 1.12.6
```
unclejack: Please leave this alone.

Makefile
```diff
@@ -213,7 +218,7 @@ ssh:
 	@vagrant ssh netplugin-node1 -c 'bash -lc "cd /opt/gopath/src/github.com/contiv/netplugin/ && bash"' || echo 'Please run "make demo"'
 
 ssh-build: start
-	vagrant ssh netplugin-node1 -c 'bash -lc "source /etc/profile.d/envvar.sh && cd /opt/gopath/src/github.com/contiv/netplugin && make run-build install-shell-completion"'
```
unclejack: This change doesn't seem related to the goal of this PR. Can you split this in a separate cleanup commit instead, please?

unclejack:
The permissions of some of the files have been changed. Are those changes intentional or accidental?

chrisplo:
the jenkins file was unintentional, the bash script was intentional

I'm holding working on this till https://github.com/contiv/netplugin/pull/1042 is resolved

Makefile
chrisplo: was reducing duplication (the build.sh defaults this to same value), but unrelated to PR, backing this out

Makefile
chrisplo: will back this change out

chrisplo:
build PR

unclejack:
build PR

unclejack:
build PR

chrisplo:
build PR

Makefile
```diff
@@ -284,55 +288,117 @@ host-restart:
...
+
+# assumes the v2plugin archive is available, installs the v2plugin and resets
+# everything on the vm to clean state
+v2plugin-install:
+	@echo Installing v2plugin
+	$(call make-on-node1, install-shell-completion host-pluginfs-unpack \
+		host-plugin-restart host-swarm-restart)
+
+# Just like demo-v2plugin except builds are done locally and cached
+demo-v2plugin-from-local: export CONTIV_DOCKER_VERSION ?= 1.13.1
```
dseevr: Can this var and the one for CONTIV_DOCKER_VERSION under `demo-v2plugin` below be consolidated into a variable at the top of the file so it only needs to be changed in one place?

install/v2plugin/Dockerfile
```diff
@@ -10,6 +10,12 @@ RUN mkdir -p /run/docker/plugins /etc/openvswitch /var/run/contiv/log \
  && wget https://github.com/andyshinn/alpine-pkg-glibc/releases/download/2.23-r1/glibc-2.23-r1.apk \
  && apk --no-cache add glibc-2.23-r1.apk
 
-COPY netplugin netmaster netctl startcontiv.sh /
+# copy in binaries and scripts
+ARG TAR_FILE
+ADD ${TAR_FILE} /
```
dseevr: If you're copying local files, use `COPY` instead of `ADD` ... generally you will not want to use `ADD` for anything, and all of Docker's official Dockerfiles have moved from `ADD` to `COPY` now

dseevr:
Do we need to worry about spaces in paths with any of this stuff?

chrisplo:
build PR

install/v2plugin/Dockerfile
chrisplo: the ADD exploded the tar file, I'd have to create another layer for a RUN to expand, there is nothing suggesting deprecation on the docs:
https://docs.docker.com/engine/reference/builder/#add

chrisplo:
only place I could see that isn't a subdirectory is here, which was a pre-existing issue:
https://github.com/contiv/netplugin/pull/1040/files#diff-b67911656ef5d18c4ae36cb6741b7965R320

chrisplo:
build PR

install/v2plugin/Dockerfile
dseevr: Oh, right. "TAR_FILE" lol (used to looking for .tar.gz and so on) 😂

.dockerignore
```diff
@@ -1,11 +1,15 @@
 # we want to get the git commit SHA but not all the binary repo data
 .git/objects/pack
 bin/
+**/*.pyc
```
dseevr: Forgot to mention this earlier... could you do a `git rm **/*.pyc` and see if anything is currently checked in which shouldn't be?  Could also be in a separate PR

.dockerignore
```diff
@@ -1,11 +1,15 @@
...
 .vagrant
 vagrant/
 docs/
 scripts/gobgp
+# this is the dockerfile for binary compilation, other dockerfiles
+# may want archive artifacts, but this one does not
 *.tar
+**/*.tar.gz
```
dseevr: Same on this guy

PR #1038, opened 2017-10-30
Remove test/systemtests/cfg.json so that it will be ignored

vhosakot:
To run the Contiv k8s system tests, before running `make k8s-test`, we must first run `scripts/python/createcfg.py` (https://github.com/contiv/netplugin/blob/master/Makefile#L172). `scripts/python/createcfg.py` creates `test/systemtests/cfg.json` in https://github.com/contiv/netplugin/blob/master/scripts/python/createcfg.py#L77-L80.

But, since `test/systemtests/cfg.json` already exists in https://github.com/contiv/netplugin/blob/master/test/systemtests/cfg.json, `. gitignore` does not ignore it in https://github.com/contiv/netplugin/blob/master/.gitignore#L63 as git does not ignore a file already present in the repo. Due to this issue, contributors have unnecessarily committed `test/systemtests/cfg.json` into the netplugin repo by mistake (https://github.com/contiv/netplugin/commits/master/test/systemtests/cfg.json).

Hence, removing `test/systemtests/cfg.json` so that it will be ignored by `. gitignore`.

Signed-off-by: Vikram Hosakote <vhosakot@cisco.com>

PR #1041, opened 2017-10-31
Adding test/systemtests/cfg.json.example

vhosakot:
This is a follow-up PR after https://github.com/contiv/netplugin/pull/1038. This PR adds `test/systemtests/cfg.json.example` as a sample for the user and updated the documentation about `cfg.json` in `test/systemtests/How-to-Run.md`.


vhosakot:
build PR

dseevr:
Since this is editing a markdown file and creating a new file which isn't even used, I think we can abort the test run here lol

PR #1042, opened 2017-10-31
version is based on git state or BUILD_VERSION

chrisplo:
Git version to be calculated by 'git describe --tags --always' and add
'-unsupported' when the local workspace has an uncommitted change.

the rules are:
* BUILD_VERSION will override any calculated version if set as env/make
  cli variable
* if current commit is tagged, use just the tag
* else `<tag>-<num commits since>-g<commit id>`

release target will use CURRENT_VERSION to make it simple to set in the
release branch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -340,10 +340,12 @@ release-built-version: tar
 	NIGHTLY_RELEASE=${NIGHTLY_RELEASE} scripts/release.sh
 	@make clean-tar
 
-# run make as a subshell but with BUILD_VERSION set to write the correct
-# version everywhere
+# The first "release" below is not a target, it is a "target-specific variable"
+#   and sets (and in this case exports) a variable to the target's environment
+# The second release runs make as a subshell but with BUILD_VERSION set
+# to write the correct version for assets everywhere
 #
-# GITHUB_USER and GITHUB_TOKEN are needed be set to run github-release
+# GITHUB_USER and GITHUB_TOKEN are needed be set (used by github-release)
 release: export BUILD_VERSION=$(shell cat version/CURRENT_VERSION)
 release:
 	@make release-built-version
--- a/Makefile
+++ b/Makefile
@@ -334,11 +334,16 @@ tar: compile-with-docker
 clean-tar:
 	@rm -f $(TAR_LOC)/*.$(TAR_EXT)
 
-# GITHUB_USER and GITHUB_TOKEN are needed be set to run github-release
+# do not run directly, use "release" target
 release-built-version: tar
 	TAR_FILENAME=$(TAR_FILENAME) TAR_FILE=$(TAR_FILE) OLD_VERSION=${OLD_VERSION} \
 	NIGHTLY_RELEASE=${NIGHTLY_RELEASE} scripts/release.sh
 	@make clean-tar
+
+# run make as a subshell but with BUILD_VERSION set to write the correct
+# version everywhere
+#
+# GITHUB_USER and GITHUB_TOKEN are needed be set to run github-release
 release: export BUILD_VERSION=$(shell cat version/CURRENT_VERSION)
 release:
-	@make release-built-version BUILD_VERSION=$(shell cat version/CURRENT_VERSION)
+	@make release-built-version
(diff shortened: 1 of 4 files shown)


chrisplo:
some testing results:
```
make release:
----
Version: 1.0.0-beta
GitCommit: 0e9fdd7-unsupported
BuildTime: 10-31-2017.23-14-26.UTC

netplugin-1.0.0-beta.tar.bz2


make tar
----
Version: 1.1.0-rc.1-207-g0e9fdd7c-unsupported
GitCommit: 0e9fdd7-unsupported
BuildTime: 10-31-2017.23-22-49.UTC

netplugin-1.1.0-rc.1-207-g0e9fdd7c-unsupported.tar.bz2


BUILD_VERSION=1.2.3 make tar
---------------------------
Version: 1.2.3
GitCommit: 0e9fdd7-unsupported
BuildTime: 10-31-2017.23-50-36.UTC

./netplugin-1.2.3.tar.bz2


make compile (on VM)
--------
[vagrant@netplugin-node1 netplugin]$ bin/netplugin --version
Version: 1.1.0-rc.1-207-g0e9fdd7-unsupported
GitCommit: 0e9fdd7-unsupported
BuildTime: 10-31-2017.23-41-11.UTC
```

unclejack:
build PR

unclejack:
@chrisplo: Can you remove the version file if it's not needed any more?

chrisplo:
@unclejack the change is such that the file is no longer generated (it was never checked in)

I rebased on master to pick up the objdb fix instead of having it as a commit in this PR

chrisplo:
build PR

tiewei:
you might need to `shfmt` because i can't tell if it's tab or space in the script

scripts/build.sh
```diff
@@ -2,28 +2,15 @@
...
-	BUILD_VERSION="$VERSION"
-else
-	BUILD_VERSION="$VERSION-$BUILD_TIME"
-fi
-
 GIT_COMMIT=$(./scripts/getGitCommit.sh)
 
-echo $BUILD_VERSION >$VERSION_FILE
+# TODO(chrisplo): remove when contiv/install no longer needs
+echo $BUILD_VERSION > netplugin-version
```
dseevr: What is this new filename for?  I don't see it referenced anywhere else

chrisplo: VERSION was $NAME-version, name was netplugin

chrisplo: let me double check, but I thought contiv/install uses it still

chrisplo:
rechecked tar,release,BUILD_VERSION

chrisplo:
build PR

Makefile
```diff
@@ -342,11 +333,12 @@ tar: compile-with-docker
 
 clean-tar:
 	@rm -f $(TAR_LOC)/*.$(TAR_EXT)
-	@rm -f ${VERSION_FILE}
 
 # GITHUB_USER and GITHUB_TOKEN are needed be set to run github-release
-release: tar
-	TAR_FILENAME=$(TAR_FILENAME) TAR_FILE=$(TAR_FILE) \
-	OLD_VERSION=${OLD_VERSION} BUILD_VERSION=${BUILD_VERSION} \
+release-built-version: tar
```
unclejack: Was this meant to be release-built-version or does it refer to the built version of netplugin?

Makefile
```diff
@@ -342,11 +333,12 @@ tar: compile-with-docker
...
+	TAR_FILENAME=$(TAR_FILENAME) TAR_FILE=$(TAR_FILE) OLD_VERSION=${OLD_VERSION} \
 	NIGHTLY_RELEASE=${NIGHTLY_RELEASE} scripts/release.sh
 	@make clean-tar
+release: export BUILD_VERSION=$(shell cat version/CURRENT_VERSION)
```
unclejack: This seems wrong.

chrisplo: this is not a target:
https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html

Makefile
chrisplo: this is a helper for release target, the release target sets BUILD_VERSION so that version is externally set instead of through scripts/getGitVersion.sh

Though I see I have BUILD_VERSION duplicated in the release target, should be able to remove one, will update

chrisplo:
latest run of release:
```

unclejack:
build PR

PR #1043, opened 2017-11-01
Init global settings from object store

tiewei:
Currently netplugin is trying to init from default settings if global
setting is not ready or wrong, and when the settings don't match it
will re-init netplugin. This is very easy to go wrong when there are
resources created before global settings are ready.
This commit makes netplugin wait until the global settings ready, and
initiate itself according to it. If the global settings are mis-match,
it will return error.

Signed-off-by: Wei Tie <wtie@cisco.com>

PR #1051, opened 2017-11-02
global fwd and subnet error msgs fix & test cases

chrisplo:
Fixed the error message to indicate correctly the state of vlans and
vxlans instead of printing an empty string instead of a number when 0.

The test cases are testing too many things in one test case, split out
the tests for forward mode and subnet as they are doing more validation.

### DIFF
--- a/netmaster/objApi/objapi_test.go
+++ b/netmaster/objApi/objapi_test.go
@@ -24,7 +24,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/contiv/contivmodel"
+	contivModel "github.com/contiv/contivmodel"
 	"github.com/contiv/contivmodel/client"
 	"github.com/contiv/netplugin/core"
 	"github.com/contiv/netplugin/netmaster/gstate"
@@ -297,7 +297,7 @@ func checkInspectGlobal(t *testing.T, expError bool, allocedVlans, allocedVxlans
 }
 
 // checkGlobalSet sets global state and verifies state
-func checkGlobalSet(t *testing.T, expError bool, fabMode, vlans, vxlans, fwdMode, arpMode, pvtSubnet string) {
+func checkGlobalSet(t *testing.T, expError bool, fabMode, vlans, vxlans, fwdMode, arpMode, pvtSubnet string) *error {
 	gl := client.Global{
 		Name:             "global",
 		NetworkInfraType: fabMode,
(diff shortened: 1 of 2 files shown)


unclejack:
build PR

chrisplo:
build PR

chrisplo:
build PR

netmaster/objApi/objapi_test.go
```diff
@@ -297,7 +297,7 @@ func checkInspectGlobal(t *testing.T, expError bool, allocedVlans, allocedVxlans
 }
 
 // checkGlobalSet sets global state and verifies state
-func checkGlobalSet(t *testing.T, expError bool, fabMode, vlans, vxlans, fwdMode, arpMode, pvtSubnet string) {
+func checkGlobalSet(t *testing.T, expError bool, fabMode, vlans, vxlans, fwdMode, arpMode, pvtSubnet string) *error {
```
dseevr: return `error` instead of `*error` here

netmaster/objApi/objapi_test.go
```diff
@@ -1318,7 +1320,79 @@ func TestDynamicGlobalVxlanRange(t *testing.T) {
 }
 
 // TestGlobalSetting tests global REST api
+func TestGlobalSettingFwdMode(t *testing.T) {
+	// set to default values (no-op)
+	checkGlobalSet(t, false, "default", "1-4094", "1-10000", "bridge", "proxy", "172.19.0.0/16")
```
dseevr: Since this returns an error, we should probably be asserting that it's nil or some expected value...

If there's a legit reason for skipping error checks on a call, maybe add a comment above the call

chrisplo: ah, i was just copy pasting that part, but I can make a helper checkGlobalSetNotNil(..) to avoid if block every call

netmaster/objApi/objapi_test.go
chrisplo: done

chrisplo:
build PR

dseevr:
Changes LGTM but let's get a @contiv/reviewers-core-networking review as well because I can't validate the network stuff is correct

unclejack:
#1043 needs to be merged first.

chrisplo:
build PR

PR #1053, opened 2017-11-07
Return pkttag with network list objects

amccormi:
Bug Fix: Return pkttag with network list objects

Manual testing: add network with VLAN id, list networks, verify VLAN id

Fixes-Jira: CNTV-175

### DIFF
--- a/vendor/github.com/contiv/contivmodel/contivModel.go
+++ b/vendor/github.com/contiv/contivmodel/contivModel.go
@@ -3513,12 +3513,19 @@ func GetOperNetwork(obj *NetworkInspect) error {
 
 // LIST REST call
 func httpListNetworks(w http.ResponseWriter, r *http.Request, vars map[string]string) (interface{}, error) {
+	var netobj NetworkInspect
 	log.Debugf("Received httpListNetworks: %+v", vars)
 
 	list := make([]*Network, 0)
 	collections.networkMutex.Lock()
 	defer collections.networkMutex.Unlock()
 	for _, obj := range collections.networks {
+		netobj.Config = *obj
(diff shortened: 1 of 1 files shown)


kahou82:
can you add unit test?

vendor/github.com/contiv/contivmodel/contivModel.go
```diff
@@ -3513,12 +3513,19 @@ func GetOperNetwork(obj *NetworkInspect) error {
...
 func httpListNetworks(w http.ResponseWriter, r *http.Request, vars map[string]string) (interface{}, error) {
+	var netobj NetworkInspect
 	log.Debugf("Received httpListNetworks: %+v", vars)
 
 	list := make([]*Network, 0)
 	collections.networkMutex.Lock()
 	defer collections.networkMutex.Unlock()
 	for _, obj := range collections.networks {
+		netobj.Config = *obj
+		if err := GetOperNetwork(&netobj); err != nil {
```
kahou82: what is the diff between this?
`if GetOperNetwork(&netobj) != nil { ... }`

dseevr:
@amccormi this change looks fine, but notice that the file you're editing is under the `vendor` directory... that means it's coming from another repo.

The file you really want to edit is here:  https://github.com/contiv/contivmodel/blob/21d55827df093ccf432760d0ff375dbec276cc9b/contivModel.go#L3515

However, that file is codegen'd from _another_ repo:  https://github.com/contiv/modelgen/blob/6e594250ebafdcc9b4cc932b91411ae6ceee4174/generators/templates.go#L938

Making this change is going to require a minimum of 3 PRs, and the final PR on netplugin will just be updating the "contivmodel" dependency in Godeps.

PR #1058, opened 2017-11-11
v2plugin set forward mode when netmaster up

chrisplo:
Docker expects the netplugin socket to be available within 10 seconds
before it fails enabling (or installing) the v2plugin.

Due to #1043, netplugin is blocking waiting for the forward mode to be
set, which is done by netctl calling netmaster, but netmaster is not
started until the plugin is activating.

Instead of backgrounding the plugin install/enabling then letting
ansible set the forward mode, do it in the plugin script to avoid
ansible's unpredictable round trip delays.

Adds a plugin setting for the forward mode

Drive-by: Logging to stdout and stderr for some commands were not
making it to the logfile, this change redirects all to the logfile

### DIFF
--- a/install/v2plugin/config.template
+++ b/install/v2plugin/config.template
@@ -110,7 +110,7 @@
           "Settable": [
              "value"
           ],
-          "Value": "bridge"
+          "Value": ""
        }
     ],
     "mounts": [
--- a/install/v2plugin/config.template
+++ b/install/v2plugin/config.template
@@ -103,6 +103,14 @@
           ],
           ## Do not change the default value, this will be replaced with $CONTIV_V2PLUGIN_NAME
           "Value": "__CONTIV_V2PLUGIN_NAME__"
+       },
+       {
+          "Description": "Forwarding mode for netplugin",
+          "Name": "fwd_mode",
+          "Settable": [
+             "value"
+          ],
+          "Value": "bridge"
        }
     ],
     "mounts": [
--- a/install/v2plugin/startcontiv.sh
+++ b/install/v2plugin/startcontiv.sh
@@ -74,6 +74,10 @@ while true ; do
 done &
 
 if [ $plugin_role == "master" ]; then
+    if [ -z "$fwd_mode" ]; then
+        echo "fwd_mode is not set, plugin cannot be enabled"
+        exit 1
+    fi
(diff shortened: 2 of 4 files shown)


install/v2plugin/startcontiv.sh
```diff
@@ -77,9 +83,24 @@ if [ $plugin_role == "master" ]; then
...
+        [ "$(curl -s -o /dev/null -w '%{http_code}' $control_url)" != "000" ] \
+           && break
+        sleep 1
+    done
+    if [ "$i" -ge "10" ]; then
+        echo "Failed to set forwarding mode, plugin will fail to enable"
+        exit 1
+    fi
+    sleep 1
+    echo "Netmaster ready for connections, setting forward mode to $fwd_mode"
```
tiewei: a nit, could we exit 1 when `$fwd_mode` not set ?

install/v2plugin/config.template
```diff
@@ -103,6 +103,14 @@
           ],
           ## Do not change the default value, this will be replaced with $CONTIV_V2PLUGIN_NAME
           "Value": "__CONTIV_V2PLUGIN_NAME__"
+       },
+       {
+          "Description": "Forwarding mode for netplugin",
+          "Name": "fwd_mode",
```
rchirakk: update https://github.com/contiv/netplugin/tree/master/install/v2plugin

rchirakk: check if you need to configure Fabric mode(ACI/default) also here

tiewei: it's fine for now about fabric mode

install/v2plugin/startcontiv.sh
```diff
@@ -77,9 +87,24 @@ if [ $plugin_role == "master" ]; then
...
+    /netctl --netmaster http://$control_url global set --fwd-mode "$fwd_mode"
```
tiewei: i think we need to check if fwd mode has been set or if it's allowed to update, or we can check the return code, if it's 5xx then retry, if it's 2xx or 4xx, it might be good to move on

tiewei: i did testing, the request is idempotent , as long as it's not changing the mode when there's network, we should be good

chrisplo:
build PR

PR #1064, opened 2017-11-14
Fix netmaster starting script in tesing codes

tiewei:
We need to add fwd_mode in netmaster starting code, also ensured
log direcotry exists when start contiv v2plugin container.


tiewei:
build PR

PR #1067, opened 2017-11-14
v2plugin logdir was created late

chrisplo:


chrisplo:
already fixed in master

PR #1069, opened 2017-11-15
Disallow invalid VLAN/VNI value during network creation

kahou82:
Currently, netctl parse the pkgTag input as an integer directly.
Therefore when user put a garbage in pkgTag, the conversion will
default it to 0 instead.

This patchset is to cast the input to string and perform further
validation logic before we create network

Signed-off-by: Kahou Lei <kalei@cisco.com>

### DIFF
--- a/netctl/netctl.go
+++ b/netctl/netctl.go
@@ -11,7 +11,6 @@ import (
 	"os"
 	"regexp"
 	"sort"
-	"strconv"
 	"strings"
 	"text/tabwriter"
 
@@ -518,8 +517,6 @@ func createNetwork(ctx *cli.Context) {
 	subnetv6 := ctx.String("subnetv6")
 	gatewayv6 := ctx.String("gatewayv6")
 
-	pktTagInString := ctx.String("pkt-tag")
-
 	if subnet == "" {
 		errExit(ctx, exitHelp, "Subnet is required", true)
 	}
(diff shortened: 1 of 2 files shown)


netctl/netctl.go
```diff
@@ -527,14 +530,22 @@ func createNetwork(ctx *cli.Context) {
...
 		if ok := net.ParseIP(gatewayv6); ok == nil {
-			errExit(ctx, exitHelp, "Invalid IPv6 gateway ", true)
+			errExit(ctx, exitHelp, "Invalid IPv6 gateway", true)
+		}
+	}
+
+	var pktTag int
+	var err error
+	if len(pktTagInString) > 0 {
+		pktTag, err = strconv.Atoi(pktTagInString)
```
chrisplo: nit, drop the declaration of err on 538 and use
pktTag, err := strconv.Atoi(pktTagInString)

kahou82: I can't as pktTag need to be used in line 560.

kahou82:
This is the issue refer to https://github.com/contiv/install/issues/182

netctl/netctl.go
chrisplo: just the err, pktTag still decalred on 537, not a blocker

chrisplo:
build PR

rchirakk:
isn't it more appropriate to change to cli.Intflag ?https://github.com/contiv/netplugin/blob/master/netctl/commands.go#L219

kahou82:
@rchirakk i thought pkt-tag can represent VLAN IDs and VNIs and therefore it was using String for the command input?

kahou82:
@chrisplo please review again

kahou82:
build PR

kahou82:
build PR

PR #1072, opened 2017-11-16
Makefile: fix tarball creation

unclejack:
The Docker versions we use for the netplugin project don't have support for the `-a` flag for `docker cp`. Based on my testing and the changes made by this PR, everything seems to work fine without that flag.

This PR also makes one small change to fix the detection of the tar binary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ HOST_GOBIN := `if [ -n "$$(go env GOBIN)" ]; then go env GOBIN; else dirname $$(
 HOST_GOROOT := `go env GOROOT`
 NAME := netplugin
 VERSION := $(shell scripts/getGitVersion.sh)
-TAR := $(shell command -v gtar || echo command -v tar || echo "Could not find tar")
+TAR := $(shell command -v gtar || command -v tar || echo "Could not find tar")
 TAR_EXT := tar.bz2
 export NETPLUGIN_CONTAINER_TAG := $(shell ./scripts/getGitVersion.sh)
 TAR_FILENAME := $(NAME)-$(VERSION).$(TAR_EXT)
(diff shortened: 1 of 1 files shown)


unclejack:
build PR

unclejack:
build PR

Makefile
```diff
@@ -15,7 +15,7 @@ HOST_GOBIN := `if [ -n "$$(go env GOBIN)" ]; then go env GOBIN; else dirname $$(
 HOST_GOROOT := `go env GOROOT`
 NAME := netplugin
 VERSION := $(shell scripts/getGitVersion.sh)
-TAR := $(shell command -v gtar || echo command -v tar || echo "Could not find tar")
+TAR := $(shell command -v gtar || command -v tar || echo "Could not find tar")
```
chrisplo: ouch

Makefile
```diff
@@ -381,7 +381,7 @@ binaries-from-container:
 	c_id=$$(docker create --name netplugin-build \
 		 netplugin-build:$(NETPLUGIN_CONTAINER_TAG)) && \
 	for f in netplugin netmaster netctl contivk8s netcontiv; do \
-		docker cp -a $${c_id}:/go/bin/$$f bin/$$f; done && \
+		docker cp $${c_id}:/go/bin/$$f bin/$$f; done && \
```
chrisplo: this will affect the ownership, so tarballs/plugin/etc. might end up with a weird ownership, have you checked the outcome?  we may need to add in a chown in some places

unclejack: They were fine in the resulting tarball when I've checked. I think it's fine. Please feel free to check again.

Makefile
unclejack: I've missed this as well during the review of some previous PR, so one more ouch.

Makefile
chrisplo: tested, binary tarball and v2plugin are both good

chrisplo:
build PR

unclejack:
This is green.
@dseevr: PTAL

tiewei:
LGTM

PR #1073, opened 2017-11-17
Makefile: build tarballs on CI

unclejack:
This PR adds the tar target to the list of targets which are called on the CI. This should make this less likely to break when changes are made.


unclejack:
build PR

unclejack:
This is WIP. I'm looking into adding some other targets to this list.

chrisplo:
build PR

unclejack:
This is green. @dseevr: PTAL

PR #1075, opened 2017-11-17
Update ovs to use info logging for docker swarm mode

kahou82:
This patchset is to update ovs log level to use info logging.

Signed-off-by: kahou lei <kalei@cisco.com>


dseevr:
build pr

PR #1077, opened 2017-11-19
[WIP] Support secure etcd connection with tls cert

BSWANG:
#### Fixes #777 
Please describe:
- support secure etcd connection with tls cert
- manual tested on secure etcd server

## TODO
- [ ] Tests
- [ ] Documentation

PR #1078, opened 2017-11-20
Refactor netplugin CLI

tiewei:
This commit refactors netplugin CLI, it uses "github.com/urfave/cli"
now in order to provide configuration from CLI and environment vars,
and provides easy way to load from config files too.
In details, this commit introduced:
1. New configs for netmode and fwdmode settings, which needs to
match what in etcd;
2. Improved validation for all CLI/ENV inputs before
starting netplugin;
3. Removed requirement of running as root user, as long as the running
user has privilege to run ovs and most networking related CLIs, is good
enough;
4. Option "-cluster-store" has been reworked, replaced by "--etcd-endpoints"
and "--consul-endpoints", so that it will be easy to enable https endpoints
and adding more than one endpoints for HA;
5. Option "-config" which reads configuration from stdin has been removed,
it will soon be replaced by reading configuration from a real file instead;
6. Option "-debug" has been replaced by "--log-level", this will allow user
to set loglevel to higher level if doesn't want to use INFO level;
7. Option "-syslog" has been replaced by "--use-syslog" and "--syslog-url",
this will make it more clear on syslog settings and endpoint.
8. Removed Option "-net-driver", because netplugin currently only supports
OVS, will add it back when there are second backend type.
9. Rework state driver and objdb client, make them more flexible to support
multi-endpoints model of etcd and consul, and http/https as URL schema.
10. Updated testing cases in react of all above changes.


tiewei:
build PR

tiewei:
build PR

netplugin/netd.go
```diff
@@ -14,248 +14,298 @@ limitations under the License.
...
+			Name:   "netmode, network-mode",
+			EnvVar: "CONTIV_NETPLUGIN_NET_MODE",
+			Usage:  "set netplugin network mode, options: [vlan, vxlan]",
+		},
+		cli.StringFlag{
+			Name:   "fwdmode, forward-mode",
+			EnvVar: "CONTIV_NETPLUGIN_NET_MODE",
+			Usage:  "set netplugin forwarding network mode, options: [bridge, routing]",
+		},
+		/*
```
dseevr: Can you mark this as WIP or leave a comment explaining why it's being left here commented out?  Commented code with no explanation is not a good thing to have in general

netplugin/netd.go
```diff
@@ -14,248 +14,298 @@ limitations under the License.
...
+			}
+		*/
+	}
+	sort.Sort(cli.FlagsByName(app.Flags))
+	app.Action = func(ctx *cli.Context) error {
+		configs, err := initNetPluginConifg(ctx)
+		if err != nil {
+			errmsg := err.Error()
+			logrus.Error(errmsg)
+			return cli.NewExitError(errmsg, (len(errmsg)%254 + 1))
```
dseevr: Could you explain what is happening here?

scripts/netContain/scripts/contivNet.sh
```diff
@@ -133,11 +136,20 @@ if [ $netmaster == true ]; then
 	done
 elif [ $netplugin == true ]; then
 	echo "Starting netplugin"
+    if [[ "$cluster_store" =~ ^etcd://.+ ]]; then
+        store_arg="--etcd-endpoints $(echo $cluster_store | sed s/etcd/http/)"
```
dseevr: We have [another PR to add TLS support to etcd](https://github.com/contiv/netplugin/pull/1077) and I'm working on upgrading the etcd code to v3 which will eventually support TLS as well...  need to think about how to handle this.  I'd like to see us move away from the `datastore://` format entirely

state/consulstatedriver.go
```diff
@@ -44,13 +46,22 @@ type ConsulStateDriver struct {
...
+	if instInfo == nil || instInfo.DbURL == "" {
+		return errors.New("no Consul config found")
+	}
+	endpoint, err = url.Parse(instInfo.DbURL)
+	if endpoint.Scheme == "consul" {
+		endpoint.Scheme = "http"
+	} else if endpoint.Scheme != "http" && endpoint.Scheme != "https" {
+		return fmt.Errorf("invalid Consul URL scheme %q", endpoint.Scheme)
+	}
+	if err != nil {
```
dseevr: Check error immediately after `url.Parse()` ... if that fails, nothing else matters

state/etcdstatedriver.go
```diff
@@ -53,19 +53,28 @@ type EtcdStateDriver struct {
...
+	endpoint, err = url.Parse(instInfo.DbURL)
+	if err != nil {
+		return err
+	}
+	if endpoint.Scheme == "etcd" {
+		endpoint.Scheme = "http"
+	} else if endpoint.Scheme != "http" && endpoint.Scheme != "https" {
+		return core.Errorf("invalid Etcd URL scheme %q", endpoint.Scheme)
+	}
+	// TODO: support multi-endpoints
```
dseevr: You can `strings.Split()` on a comma, iterate over the result, validate the URLs, and append to the array you set as `Endpoints`

test/systemtests/swarm_test.go
```diff
@@ -391,8 +391,18 @@ func (w *swarm) cleanupContainers() error {
...
+	if cStore := w.node.suite.basicInfo.ClusterStore; strings.HasPrefix(cStore, "etcd") {
+		storeArgs = " --etcd-endpoints " + strings.Replace(cStore, "etcd", "http", 1) + " "
+	} else {
+		storeArgs = " --consul-endpoints " + strings.Replace(cStore, "consul", "http", 1) + " "
+	}
+
 	logrus.Infof("Starting netplugin on %s", w.node.Name())
-	cmd := "sudo " + w.node.suite.basicInfo.BinPath + "/netplugin -plugin-mode docker -vlan-if " + w.node.suite.hostInfo.HostDataInterfaces + " --cluster-store " + w.node.suite.basicInfo.ClusterStore + " " + args + "&> /tmp/netplugin.log"
+
+	cmd := "sudo " + w.node.suite.basicInfo.BinPath + "/netplugin --netmode " + netMode + " --fwdmode " + fwdMode + " -plugin-mode docker -vlan-if " + w.node.suite.hostInfo.HostDataInterfaces + storeArgs + args + "&> /tmp/netplugin.log"
```
dseevr: interpolate?  looks kinda ugly 🤢

core/error.go
```diff
@@ -38,17 +37,12 @@ type Error struct {
 // Error() allows *core.Error to present the `error` interface.
 func (e *Error) Error() string {
 	var ret string
+	ret = e.desc + "\n"
```
dseevr: Do we need to change this file as part of this PR?  Weren't we planning to just get rid of this `core.Error` type entirely? :D

netplugin/netd.go
tiewei: the comment is `// only ovs is supported` and it also mentioned in commit message ... we only support ovs at the moment there's no option ti add driver option

netplugin/netd.go
tiewei: get error msg and print log, exit with error msg and error code.
There's trick that the return code is length of error msg mod 254 +1 to make sure it's an in 1-255 range

scripts/netContain/scripts/contivNet.sh
tiewei: we are, i have to do this step by step, the reason it's still here because netmaster doesn't support the new way yet, once netmaster is changed to use new way, these codes can be dropped

state/consulstatedriver.go
tiewei: hum, I recalled i did that ,maybe i missed this one

state/etcdstatedriver.go
tiewei: the reason it still only supports single endpoint is because netmaster also shares the same part of codes, i need to make netmaster using new way then tweak these codes, it's on my list

core/error.go
tiewei: we are but not in this PR, here is really to get rid of the ENV['CONTIV_DEBUG'] thing only

netplugin/netd.go
dseevr: Right, I understand what the code is doing, but why are we returning an error code based on the length of the string?  Doesn't seem very useful to me (there could be multiple error messages with the same length, etc.)

tiewei: because all errors returns the same code is a little bad, i wanted to refactor `core/error` to include cli & http return code but it's a little to much for this PR.

tiewei:
build PR

netplugin/netd.go
dseevr: We should subtree contivmodel and then we can fix the entire API in one PR 👍

netplugin/netd.go
```diff
@@ -14,248 +14,299 @@ limitations under the License.
...
-		log.Fatalf("Error getting local address. Err: %v", err)
+	// 4. validate forward mode
+	forwardMode := strings.ToLower(ctx.String("fwdmode"))
+	if (networkMode == "vlan" && forwardMode == "bridge") || forwardMode == "routing" {
+		logrus.Infof("Using netplugin forwarding mode: %v", forwardMode)
+	} else if forwardMode == "" {
+		return plugin.Config{}, fmt.Errorf("netplugin network mode is not set")
+	} else {
+		return plugin.Config{}, fmt.Errorf("invalid netplugin forwarding mode: %v", forwardMode)
+	}
```
chrisplo: I think here it's frustrating to end user to decipher why output would say
`invalid netplugin forwarding mode: bridge`
when of course bridge is valid, but not with vxlan .. vs something more precise like `bridge forwarding mode is invalid when configured with vxlan` and `invalid netplugin forwarding mode: forwarding_mode_that_does_not_exist` when the forwarding mode was `forwarding_mode_that_does_not_exist`, which is never valid.

netplugin/netd.go
```diff
@@ -14,248 +14,299 @@ limitations under the License.
...
+		},
+		cli.StringFlag{
+			Name:   "vlan-uplinks, vlan-if",
+			EnvVar: "CONTIV_NETPLUGIN_VLAN_UPLINKS",
+			Usage:  "a comma-delimited list of netplugin uplink interfaces",
+		},
+		cli.StringFlag{
+			Name:   "etcd-endpoints, etcd",
+			EnvVar: "CONTIV_NETPLUGIN_ETCD_ENDPOINTS",
+			Usage:  "a comma-delimited list of netplugin etcd endpoints",
```
chrisplo: specify that consul-endpoints is ignored if this is set

netplugin/netd.go
```diff
@@ -14,248 +14,299 @@ limitations under the License.
...
+		},
+		cli.StringFlag{
+			Name:   "consul-endpoints, consul",
+			EnvVar: "CONTIV_NETPLUGIN_CONSUL_ENDPOINTS",
+			Usage:  "a comma-delimited list of netplugin consul endpoints",
```
chrisplo: indicate that this is ignored if etcd-endpoints are set

netplugin/netd.go
tiewei: so the log will be 
```
Using netplugin netmode vxlan
invalid netplugin forwarding mode: bridge
```
I think thats clear enough to say it's not a valid option, plus docs will list down it as well

netplugin/netd.go
tiewei: etcd endpoint will always take priority when it's set

netplugin/netd.go
tiewei: done

netplugin/netd.go
tiewei: done

tiewei:
build PR

tiewei:
build PR

core/error_test.go
```diff
@@ -35,17 +33,6 @@ func TestErrorStackTrace(t *testing.T) {
...
-
-	expectedStr := fmt.Sprintf("%s [%s %s %d]", msg, funcName, fileName, lineNum)
-
-	if e.Error() != expectedStr {
-		t.Fatalf("Error message yielded an incorrect result with CONTIV_TRACE unset: %s", e.Error())
-	}
-
-	os.Setenv("CONTIV_TRACE", "1")
 	if e.Error() == "an error\n" {
 		t.Fatal("Error message did not yield stack trace with CONTIV_TRACE set")
```
vhosakot: `CONTIV_TRACE` is not used anymore, right?

netplugin/netd.go
vhosakot: Refactoring `core/error` is one idea. Alternatively, I think `initNetPluginConifg` can return a unique error code for every error (using Go's `hash`) like
```
func initNetPluginConifg(ctx *cli.Context) (plugin.Config, error, uniqueErrorCode) {
```
instead of
```
func initNetPluginConifg(ctx *cli.Context) (plugin.Config, error) {
```

objdb/client.go
```diff
@@ -24,6 +24,21 @@ import (
 
 var defaultDbURL = "etcd://127.0.0.1:2379"
 
+// InitClient aims to support multi endpoints
+func InitClient(storeName string, storeURLs []string) (API, error) {
+	plugin := GetPlugin(storeName)
+	if plugin == nil {
+		log.Errorf("Invalid DB type %s", storeName)
```
vhosakot: I'd use`datastore` instead of `DB` :)

scripts/netContain/scripts/contivNet.sh
```diff
@@ -133,19 +132,21 @@ if [ $netmaster == true ]; then
...
+    elif [[ "$cluster_store" =~ ^consul://.+ ]]; then
+        store_arg="--consul-endpoints $(echo $cluster_store | sed s/consul/http/)"
```
vhosakot: Can `--cluster-store` and `--listen-url` be retained for backward compatibility?

vhosakot:
Nice work! Had some comments. Also, did not review the 21 new files vendored in `vendor/github.com/urfave`.

scripts/netContain/scripts/contivNet.sh
tiewei: it will still work for the container script but not for binary, because `--cluster-store` has restriction on the url schema

core/error_test.go
tiewei: right

objdb/client.go
tiewei: The `NewClient` is using the DB name, here is just make them consistent. meanwhile the package named `objdb` so i think db is not a terrible name

core/error_test.go
vhosakot: Then, why does the error say `Error message did not yield stack trace with CONTIV_TRACE set`? Should the message be corrected?

vhosakot:
The error `Error message did not yield stack trace with CONTIV_TRACE set` in `core/error_test.go` need to refer `CONTIV_TRACE` as `CONTIV_TRACE` is not used anymore.  Can be corrected in a follow-up PR.

netplugin/netd.go
tiewei: the `core/error.go` needs to be refactored and will include a http return codes and cli return codes IMO, i'll put it later

PR #1081, opened 2017-11-22
build static binaries & remove glibc from v2plugin

unclejack:
- makes the compiled binaries static
- removes glibc from the v2plugin Dockerfile
- bumps the alpine image version used in v2plugin to alpine 3.6

### DIFF
--- a/netplugin/netd.go
+++ b/netplugin/netd.go
@@ -75,8 +75,7 @@ func startNetPlugin(pluginConfig *plugin.Config) {
 	}
 }
 
-func initNetPluginConifg(ctx *cli.Context) (plugin.Config, error) {
-
+func initNetPluginConfig(ctx *cli.Context) (plugin.Config, error) {
 	// 1. validate and set up log
 	if ctx.Bool("use-syslog") {
 		syslogURL := ctx.String("syslog-url")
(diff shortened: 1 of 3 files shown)


unclejack:
build PR

unclejack:
build PR

netplugin/netd.go
```diff
@@ -76,6 +77,11 @@ func startNetPlugin(pluginConfig *plugin.Config) {
 }
 
 func initNetPluginConifg(ctx *cli.Context) (plugin.Config, error) {
```
chrisplo: if you end up with another push, please fix the spelling here as a drive by?

netplugin/netd.go
```diff
@@ -76,6 +77,11 @@ func startNetPlugin(pluginConfig *plugin.Config) {
...
+	// Make sure we are running as root
+	uid := syscall.Getuid()
+	if uid != 0 {
+		log.Fatalf("This process can only be run as root")
+	}
```
chrisplo: is this required to talk to ovs?  though wei removed this because it was not required

unclejack: @chrisplo: I can remove it if it's not needed.

netplugin/netd.go
unclejack: Sure, I'll fix it in a separate commit.

unclejack:
@chrisplo: PTAL

unclejack:
build PR

PR #1085, opened 2017-11-28
Update contivmodel import paths

dseevr:
This PR updates all the Go import paths for `contivmodel` to point to the new directory under `netplugin` instead of the original separate repo.

It also updates Godeps and removes the contivmodel dependency from `vendor`.


dseevr:
build pr

PR #1092, opened 2017-12-01
netctl: provide fallback for user.Current

unclejack:
This PR fixes the static netctl binary. This makes it possible to run a statically built netctl binary without a panic.


netctl/config.go
```diff
@@ -50,8 +50,19 @@ func configExists(ctx *cli.Context) bool {
...
+	if err == nil {
+		homeDir = usr.HomeDir
+	}
+	// this should work where we don't have static binaries
 	if err != nil {
+		homeDir = os.Getenv("HOME")
+	}
+	// panic if we've failed to retrieve the home directory
+	if homeDir == "" {
+		err = fmt.Errorf("user.Current is unavailable & HOME isn't set")
```
tiewei: This error msg doesn't help end user understand whats the problem and how to resolve it, maybe just saying `Failed to determine home directory of current user, please set environment variable HOME`

dseevr:
Since I don't think we intend to ever go back to non-static binaries, can we just remove the `user.Current` call entirely?

unclejack:
build PR

unclejack:
@dseevr: In case we start building binaries for other platforms, I'd rather have it there.

I've modified the error message.

unclejack:
build PR

dseevr:
build pr

chrisplo:
build PR

unclejack:
build PR

unclejack:
This is green now. I'll merge.

PR #1095, opened 2017-12-04
add OVS Docker image 

unclejack:
This PR adds support for building a standalone openvswitch image and it adds a new target to the list of CI targets.

The PR also modifies scripts/netContain/build_image.sh. That change makes it load the openvswitch kernel module only if it's not loaded.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ export V2PLUGIN_TAR_FILENAME := v2plugin-$(VERSION).tar.gz
 GO_MIN_VERSION := 1.7
 GO_MAX_VERSION := 1.8
 GO_VERSION := $(shell go version | cut -d' ' -f3 | sed 's/go//')
-CI_HOST_TARGETS ?= "host-unit-test host-integ-test host-build-docker-image tar host-pluginfs-create clean-tar"
+CI_HOST_TARGETS ?= "host-unit-test host-integ-test host-build-docker-image host-build-ovs-image tar host-pluginfs-create clean-tar"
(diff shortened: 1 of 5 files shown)


unclejack:
build PR

tiewei:
Since this is pretty standalone, can we create a repo for ovs container (or fork a upstream one) instead ? this way could keep thie repo only for netplugin/netmaster codes

unclejack:
We can't use any upstream image. This is tightly coupled with our netplugin setup. It puts logs in the contiv directory. We also need to use a version which has been tested with netplugin.

tiewei:
Actually other than logging dir this is a pretty common ovs configuration (and we can update entrypoint if we wanted), we can customized base on an upstream repo to avoid maintain generic containers, also we can move it out of release scope, let upstream (active) to do security upgrading and all we need to do is to make sure it works for contiv.
It's fine to put this fully under control if you think it's really needed, but could you please make it in different repo ? It's really a generic component doesn't have to be in netplugin repo

unclejack:
I've worked on this because it was part of some work which was needed to be finished soon. I can close this PR and attempt to create a repository for this.

amccormi:
I agree with Wei. If there is a way for us to use a generic OVS container, and customize for our environment where needed, that would be better than us maintaining our own builds/repo/etc. Also, I'm wondering why the OVS logs have to be in the contiv dir?

dseevr:
@amccormi that directory is bindmounted from the host... otherwise we would lose the logs if the container crashed/restarted

unclejack:
I'll close this now. The image will live in a separate repository.

PR #1096, opened 2017-12-05
Add extra validation check for v2plugin

kahou82:
Currently, demo-v2plugin doesn't check if contiv is installed and
enabled.

This patchset is to check if v2plugin is installed and enabled at the
end of the installation

Signed-off-by: <kalei@cisco.com>

### DIFF
--- a/scripts/python/startSwarm.py
+++ b/scripts/python/startSwarm.py
@@ -2,10 +2,10 @@
 
 # Start netplugin and netmaster
 import api.tnode
-import time
-import sys
-import os
 import argparse
+import os
+import re
+import time
 
 # Parse command line args
 # Create the parser and sub parser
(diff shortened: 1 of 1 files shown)


scripts/python/startSwarm.py
```diff
@@ -45,6 +45,21 @@
         node.runCmdThread(command)
 
     time.sleep(15)
+
+    print "Check netplugin is installed and enabled"
+    out, x, y = nodes[0].runCmd("docker plugin ls")
```
chrisplo: if you don't care about the err and exit value, use '_', though I think you could have docker installed but docker plugin ls fail when dockerd is not running, so you might want to check exit status

kahou82: 👍 done.

PR #1098, opened 2017-12-06
fix ovsd driver failing to init

chrisplo:
The logic was reversed for checking if driver info was already in db


chrisplo:
build PR

chrisplo:
the weird logic for the err check:
https://github.com/contiv/netplugin/blob/master/core/error.go#L81

dseevr:
build pr

chrisplo:
actually think the usage is correct . . .despite the PR working for me

PR #1104, opened 2017-12-19
Move environment init to netplugin-init container

tiewei:
Move envrionment init codes to netplugin-init container.
Make get k8s configs from "/var/contiv/config/contiv.json" to avoid
copying configs around

### DIFF
--- a/install/k8s/contiv/contiv.yaml
+++ b/install/k8s/contiv/contiv.yaml
@@ -1,4 +1,99 @@
 ---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: contiv-netplugin
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: contiv-netplugin
+subjects:
+- kind: ServiceAccount
+  name: contiv-netplugin
+  namespace: kube-system
+
+---
+
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: contiv-netplugin
+  namespace: kube-system
+rules:
+  - apiGroups:
+    - ""
+    - extensions
+    resources:
+      - endpoints
+      - nodes
+      - namespaces
+      - networkpolicies
+      - pods
+      - services
+    verbs:
+      - watch
+      - list
+      - update
+      - get
+
+---
+
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: contiv-netplugin
+  namespace: kube-system
+  labels:
+    kubernetes.io/cluster-service: "true"
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: contiv-netmaster
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: contiv-netmaster
+subjects:
+- kind: ServiceAccount
+  name: contiv-netmaster
+  namespace: kube-system
+
+---
+
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: contiv-netmaster
+  namespace: kube-system
+rules:
+  - apiGroups:
+    - ""
+    - extensions
+    resources:
+      - pods
+      - nodes
+      - namespaces
+      - networkpolicies
+    verbs:
+      - watch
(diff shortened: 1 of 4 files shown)


tiewei:
build PR

install/k8s/contiv/contiv.yaml
```diff
@@ -1,4 +1,99 @@
 ---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
```
vhosakot: Contiv works with the default k8s ClusterRoleBindings.  Curious why this is needed? for RBAC?

install/k8s/contiv/contiv.yaml
```diff
@@ -1,4 +1,99 @@
...
+    verbs:
+      - watch
+      - list
+      - update
+      - get
+
+---
+
+apiVersion: v1
+kind: ServiceAccount
```
vhosakot: Same question here too. Contiv works with the default `ServiceAccount` cert in (`/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`). Why is this needed?

https://github.com/contiv/netplugin/blob/master/Makefile#L18

tiewei: That's CA, not token, I think token is the one indicates k8s role and accounts

install/k8s/contiv/contiv.yaml
tiewei: RBAC

install/k8s/contiv/contiv.yaml
vhosakot: Every `ServiceAccount` translates to a `ca.crt` file on disk:

https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/

install/k8s/contiv/contiv.yaml
```diff
@@ -90,20 +143,63 @@ spec:
...
       hostNetwork: true
       hostPID: true
+      tolerations:
+      - key: node-role.kubernetes.io/master
+        effect: NoSchedule
+      serviceAccountName: contiv-netplugin
+      initContainers:
+        - name: contiv-netplugin-init
+          image: contiv/netplugin-etcd-init:latest
+          imagePullPolicy: Always
```
dseevr: Could this result in an incompatible init container being pulled down and run with an older release of Contiv?

install/k8s/contiv/contiv.yaml
```diff
@@ -219,21 +289,51 @@ spec:
...
+      - key: node-role.kubernetes.io/master
+        effect: NoSchedule
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
+      serviceAccountName: contiv-netmaster
       containers:
+      initContainers:
+        - name: contiv-netplugin-init
+          image: contiv/netplugin-etcd-init:latest
+          imagePullPolicy: Always
```
dseevr: here, too

install/k8s/contiv/contiv.yaml
```diff
@@ -219,21 +289,51 @@ spec:
...
+                configMapKeyRef:
+                  name: contiv-config
+                  key: contiv_k8s_config
+            - name: CONTIV_CNI_CONFIG
+              valueFrom:
+                configMapKeyRef:
+                  name: contiv-config
+                  key: contiv_cni_config
+          volumeMounts:
+            - mountPath: /var/contiv
```
dseevr: Is this for backwards compatibility with the old location?

install/k8s/contiv/contiv.yaml
```diff
@@ -262,52 +357,142 @@ spec:
...
+        - name: contiv-api-proxy
+          image: contiv/auth_proxy:1.2.0
+          args:
+            - --tls-key-file=/var/contiv/auth_proxy_key.pem
+            - --tls-certificate=/var/contiv/auth_proxy_cert.pem
+            - --data-store-address=$(STORE_URL)
+            - --data-store-driver=$(STORE_DRIVER)
+            - --netmaster-address=localhost:9999
+          env:
+            - name: NO_NETMASTER_STARTUP_CHECK
```
dseevr: We shouldn't have this env var set by default... this allows auth_proxy to send requests to an incompatible netmaster

install/k8s/contiv/contiv.yaml
tiewei: it could be, but we have to do it before we starting tagging `netplugin-etcd-init`

tiewei:
close pull request  need to rework based on #1106

PR #1105, opened 2017-12-20
Create a single Dockerfile that builds K8S image

tiewei:
We need a single Dockerfile in order to create docker image by
dockerhub autobuild, and also make it easy for nightly build and
dev build.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,3 @@
-# we want to get the git commit SHA but not all the binary repo data
-.git/objects/pack
 bin/
 **/*.pyc
 .vagrant
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,55 +1,38 @@
-##
-#Copyright 2014 Cisco Systems Inc. All rights reserved.
-#
-#Licensed under the Apache License, Version 2.0 (the "License");
-#you may not use this file except in compliance with the License.
-#You may obtain a copy of the License at
-#http://www.apache.org/licenses/LICENSE-2.0
-#
-#Unless required by applicable law or agreed to in writing, software
-#distributed under the License is distributed on an "AS IS" BASIS,
-#WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-#See the License for the specific language governing permissions and
-#limitations under the License.
-##
+# builder is where netplugin got complied
+FROM golang:1.7.6 as builder
 
-##
-# Container image for compiled netplugin binaries
-#
-# Run netplugin:
-# docker run --net=host <image> -host-label=<label>
-##
+ENV GOPATH /go
 
-FROM golang:1.7.6
+COPY . $GOPATH/src/github.com/contiv/netplugin/
 
-# Insert your proxy server settings if this build is running behind
-# a proxy.
-#ENV http_proxy ""
-#ENV https_proxy ""
-ARG http_proxy
-ARG https_proxy
+WORKDIR $GOPATH/src/github.com/contiv/netplugin/
 
-WORKDIR /go/src/github.com/contiv/netplugin/
+RUN VERSION_SUFFIX="$(if git diff-index --quiet HEAD --; then echo '-unsupported'; fi)" && \
+    GIT_COMMIT=$(git rev-parse --short HEAD)$VERSION_SUFFIX && \
+    BUILD_VERSION=$(git describe --tags --always)$VERSION_SUFFIX  && \
+    PKG_NAME=github.com/contiv/netplugin/version && \
+    BUILD_TIME=$(date -u +%m-%d-%Y.%H-%M-%S.UTC) && \
+    GOGC=1500 CGO_ENABLED=0 go install -v \
+    -a -installsuffix cgo \
+    -ldflags "-X $PKG_NAME.version=$BUILD_VERSION \
+    -X $PKG_NAME.buildTime=$BUILD_TIME \
+    -X $PKG_NAME.gitCommit=$GIT_COMMIT \
+    -s -w -d" -pkgdir /tmp/foo-cgo \
+    ./netplugin/ ./netmaster/ ./netctl/netctl/ ./mgmtfn/k8splugin/contivk8s/ && \
+    mkdir -p /contiv/bin && \
+    for bin in netplugin netmaster netctl contivk8s; do cp /go/bin/$bin /contiv/bin/ ; done && \
+    /contiv/bin/netplugin --version && /contiv/bin/netmaster --version
 
-ENTRYPOINT ["netplugin"]
-CMD ["--help"]
+# The container where netplugin will be run
+FROM ubuntu:16.04
 
-# by far, most of the compilation time is building vendor packages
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -1,55 +1,40 @@
-##
-#Copyright 2014 Cisco Systems Inc. All rights reserved.
```
vhosakot: Any reason copyright is removed?

Dockerfile
```diff
@@ -1,55 +1,40 @@
...
+    -X $PKG_NAME.buildTime=$BUILD_TIME \
+    -X $PKG_NAME.gitCommit=$GIT_COMMIT \
+    -s -w -d" -pkgdir /tmp/foo-cgo \
+    ./netplugin/ ./netmaster/ ./netctl/netctl/ ./mgmtfn/k8splugin/contivk8s/ && \
+    mkdir -p /contiv/bin && \
+    for bin in netplugin netmaster netctl contivk8s; do cp /go/bin/$bin /contiv/bin/ ; done && \
+    /contiv/bin/netplugin --version && /contiv/bin/netmaster --version
+
+# The container where netplugin will be run
+FROM ubuntu:16.04
```
vhosakot: Will there be a version for CentOS too?

Dockerfile
```diff
@@ -1,55 +1,40 @@
...
+
+COPY ./vendor/ $GOPATH/src/
+
+COPY . $GOPATH/src/github.com/contiv/netplugin/
+
+WORKDIR $GOPATH/src/github.com/contiv/netplugin/
+
+RUN VERSION_SUFFIX="$(if git diff-index --quiet HEAD --; then echo '-unsupported'; fi)" && \
+    GIT_COMMIT=$(git rev-parse --short HEAD)$VERSION_SUFFIX && \
+    BUILD_VERSION=$(git describe --tags --always)$VERSION_SUFFIX  && \
```
vhosakot: Why not use `scripts/getGitVersion.sh` to get `BUILD_VERSION`?

https://github.com/contiv/netplugin/blob/master/scripts/getGitVersion.sh
https://github.com/contiv/netplugin/blob/master/Makefile#L18

tiewei: Because that file can be deleted after this PR

Dockerfile
tiewei: No, we actually want to move to alpine

Dockerfile
tiewei: Because i think it's not useful given there's `LICENSE` file under this repo already

Dockerfile
```diff
@@ -1,55 +1,40 @@
...
-ARG NIGHTLY_RELEASE=""
-
-RUN GOPATH=/go/ \
-    BUILD_VERSION="${BUILD_VERSION}" \
-    NIGHTLY_RELEASE="${NIGHTLY_RELEASE}" \
-    make compile \
-      && cp scripts/contrib/completion/bash/netctl /etc/bash_completion.d/netctl \
-      && netplugin -version
+# builder is where netplugin got complied
+FROM golang:1.7.6 as builder
```
dseevr: Multi-stage builds are only supported in 17.05 or >

Will this change affect anything relying on this var?  https://github.com/contiv/netplugin/blob/0308c0f45158412968e0e899430c3736481338e8/Makefile#L8

tiewei: This is used for dockerhub auto build(and they have newer enough docker), in our local build though, i think we should upgrade if its not supported (and/or provide a fallback if it's a customer ask)

chrisplo: if you run centos and maybe ubuntu, you get the older version, you have to do extra work to get the newer stuff, I'm not personally opposed to requiring newer docker considering we're not supporting older docker runtime deployments . .

.dockerignore
```diff
@@ -1,5 +1,3 @@
-# we want to get the git commit SHA but not all the binary repo data
-.git/objects/pack
```
chrisplo: this will make the container build really long if this container is the primary one used for compiling, the history for netplugin includes binaries, etc. and is copied over

Dockerfile
chrisplo: the reason it was done as a script outside was because the copy was huge and done twice (at least on mac), once to context, then to the image

Dockerfile
```diff
@@ -1,55 +1,40 @@
...
+    BUILD_TIME=$(date -u +%m-%d-%Y.%H-%M-%S.UTC) && \
+    GOGC=1500 CGO_ENABLED=0 go install -v \
+    -a -installsuffix cgo \
+    -ldflags "-X $PKG_NAME.version=$BUILD_VERSION \
+    -X $PKG_NAME.buildTime=$BUILD_TIME \
+    -X $PKG_NAME.gitCommit=$GIT_COMMIT \
+    -s -w -d" -pkgdir /tmp/foo-cgo \
+    ./netplugin/ ./netmaster/ ./netctl/netctl/ ./mgmtfn/k8splugin/contivk8s/ && \
+    mkdir -p /contiv/bin && \
+    for bin in netplugin netmaster netctl contivk8s; do cp /go/bin/$bin /contiv/bin/ ; done && \
```
chrisplo: can we hard link instead of copy

Dockerfile
tiewei: it doesn't impact where running the container, it only impacts where build the container

.dockerignore
tiewei: yes, that's intended, without objects file git couldn't figure out which version it's on, thus will put the version info wrong

Dockerfile
tiewei: it's really fast do a copy, i don't think this impact anything, it's just to make it easier to copy a dir to the runner container

Dockerfile
```diff
@@ -1,55 +1,40 @@
...
-      && cp scripts/contrib/completion/bash/netctl /etc/bash_completion.d/netctl \
-      && netplugin -version
+# builder is where netplugin got complied
+FROM golang:1.7.6 as builder
+
+ENV GOPATH /go
+
+COPY ./vendor/ $GOPATH/src/
+
+COPY . $GOPATH/src/github.com/contiv/netplugin/
```
chrisplo: can we enumerate the source dirs and other artifacts here?  why two copies of vendor and one in /go/src?

Dockerfile
tiewei: it doesn't really slow a lot things down and plus dockerhub doesn't allow you to run make file, so we have to keep git info inside container anyway

Dockerfile
dseevr: when building a package, $PACKAGE_BASE/vendor is checked before $GOPATH/src, so I think the COPY on line 6 is unnecessary

tiewei: 👍 i'm remove it

tiewei:
build PR

Dockerfile
dseevr: `cp` will just make the build image slightly larger than if you used `ln`

vhosakot: `cp` duplicates files in the container which hardlink doesn't.

PR #1110, opened 2018-01-17
Don't kill existing traffic on netplugin restart

nbartos:
Blindly deleting and recreating ports cause pod network outages
until the pod is either recreated, or the node is rebooted
(because that causes a CNI DEL & ADD).

There is still a problem with restarting OVS causing similar network
outages, but this at least fixes half of the problem.

### DIFF
--- a/drivers/ovsd/ovsSwitch.go
+++ b/drivers/ovsd/ovsSwitch.go
@@ -299,7 +299,7 @@ func (sw *OvsSwitch) CreatePort(intfName string, cfgEp *mastercfg.CfgEndpointSta
 	var ovsIntfType string
 	var err error
 	vethCreated := false
-	dbUpdated := false
+	portCreated := false
 
 	// Get OVS port name
 	ovsPortName := getOvsPortName(intfName, skipVethPair)
@@ -308,7 +308,7 @@ func (sw *OvsSwitch) CreatePort(intfName string, cfgEp *mastercfg.CfgEndpointSta
 			if vethCreated {
 				deleteVethPair(intfName, ovsPortName)
 			}
-			if dbUpdated {
+			if portCreated {
 				sw.ovsdbDriver.DeletePort(intfName)
 			}
 		}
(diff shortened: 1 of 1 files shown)


drivers/ovsd/ovsSwitch.go
```diff
@@ -337,22 +337,15 @@ func (sw *OvsSwitch) CreatePort(intfName string, cfgEp *mastercfg.CfgEndpointSta
...
-	// If the port already exists in OVS, remove it first
-	if sw.ovsdbDriver.IsPortNamePresent(ovsPortName) {
-		log.Debugf("Removing existing interface entry %s from OVS", ovsPortName)
-
-		// Delete it from ovsdb
-		err = sw.ovsdbDriver.DeletePort(ovsPortName)
+	// Recreating the OVS port will kill existing interface traffic, so only
+	// create if necessary.  Never delete + create.
+	if !sw.ovsdbDriver.IsPortNamePresent(ovsPortName) {
+		err = sw.ovsdbDriver.CreatePort(ovsPortName, ovsIntfType, cfgEp.ID, pktTag, burst, bandwidth)
```
chrisplo: can you call UpdatePort instead?

drivers/ovsd/ovsSwitch.go
```diff
@@ -337,22 +337,15 @@ func (sw *OvsSwitch) CreatePort(intfName string, cfgEp *mastercfg.CfgEndpointSta
...
 		if err != nil {
-			log.Errorf("Error deleting port %s from OVS. Err: %v", ovsPortName, err)
+			return err
```
chrisplo: the local context logging might be good to keep, but keep the return too

drivers/ovsd/ovsSwitch.go
```diff
@@ -839,24 +832,16 @@ func (sw *OvsSwitch) AddHostPort(intfName string, intfNum, network int, isHostNS
...
-	// If the port already exists in OVS, remove it first
-	if sw.ovsdbDriver.IsPortNamePresent(ovsPortName) {
-		log.Infof("Removing existing interface entry %s from OVS", ovsPortName)
-
-		// Delete it from ovsdb
-		err := sw.ovsdbDriver.DeletePort(ovsPortName)
+	// Recreating the OVS port will kill existing interface traffic, so only
+	// create if necessary.  Never delete + create.
+	if !sw.ovsdbDriver.IsPortNamePresent(ovsPortName) {
+		err = sw.ovsdbDriver.CreatePort(ovsPortName, ovsPortType, portID, hostVLAN, 0, 0)
```
chrisplo: update?

PR #1126, opened 2018-03-20
modified burst value

wangxf1987:
Bugfix: modified that check burst value when run the function ValidateNetprofile.
1. the burst is hard code
2. the burst can't set 10% of rate
3. if want to achieve the desired performance, it's best to equal the value of rate

#### Fixes #
Please describe:
1. No matter how much the value of rate is set, The value of burst cannot exceed 10486. 
2. If rate is 100mbps and the burst is 10485, Actually, the bandwidth  only 50mbps.
image
3. If set the rate is 1000mbps and the burst is 10485, the result is same to above.
image
## TODO
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contivmodel/contivModel.go
+++ b/contivmodel/contivModel.go
@@ -3457,7 +3457,7 @@ func ValidateNetprofile(obj *Netprofile) error {
 		return errors.New("bandwidth string invalid format")
 	}
 
-	if obj.Burst > 10486 {
+	if obj.Burst > obj.Burst {
 		return errors.New("burst Value Out of bound")
 	}
 


contivmodel/contivModel.go
```diff
@@ -3457,7 +3457,7 @@ func ValidateNetprofile(obj *Netprofile) error {
 		return errors.New("bandwidth string invalid format")
 	}
 
-	if obj.Burst > 10486 {
+	if obj.Burst > obj.Burst {
```
dseevr: How would this condition ever be true? `x` can't be greater than `x`

wangxf1987: It's ensure that some errors, I will be commit again. sorry.

dseevr: I reverted the change on `master`.  Feel free to open a new PR 👍

PR #1135, opened 2018-05-22
configure Vxlan  veth interface mtu based on configured minimum host …

g1rana:
…interface mtu

## Description of the changes:
Currently container mtu in contiv network set as static MTU value as 1450. With this PR, contiv/netplugin 
detect  minimum configured mtu of host and cofingured Veth pair container side MTU to read MTU - 50 if container is part of Vxlan Bridge in OVS  
#### Fixes #1132
Please describe:
-  Contiv/netplugin bringup 2 types of OVS switch in every node and all container interface are plugged into these OVS switch based on OVS bridge mode type( Vlan or Vxlan). Incase of Vxlan ,  Container should not send packets more configured host Interface , to fix this issue, netplugin  read lowest configured mtu in set of host interface and set container interface mtu  50 byte less if container interface is part OVS vxlan bridge

 Bring up  Contiv network with Vxlan encap as default switch. 
 Bringup container which is part of above contiv network .
check the configured interface MTU which should be 50 byte less than host minimum MTU

## TODO
- [ ] Tests
- [ ] Documentation


drivers/ovsd/ovsSwitch.go
```diff
@@ -100,6 +101,10 @@ func NewOvsSwitch(bridgeName, netType, localIP, fwdMode string,
 	sw.netType = netType
 	sw.uplinkDb = cmap.New()
 	sw.hostPvtNW = hostPvtNW
+	sw.vxlanEncapMtu, err = netutils.GetHostLowestLinkMtu()
+	if err != nil {
+		log.Fatalf("Failed to get Host Node MTU. Err:%v", err)
```
dseevr: nitpick: add a space between `:` and `%v`

drivers/ovsd/ovsSwitch.go
```diff
@@ -359,7 +364,12 @@ func (sw *OvsSwitch) CreatePort(intfName string, cfgEp *mastercfg.CfgEndpointSta
 
 	// Set the link mtu to 1450 to allow for 50 bytes vxlan encap
 	// (inner eth header(14) + outer IP(20) outer UDP(8) + vxlan header(8))
-	err = setLinkMtu(intfName, vxlanEndpointMtu)
+	if sw.netType == "vxlan" {
```
dseevr: Is there a constant you can use here instead of a string literal?

g1rana: There is no constant available ; same string literal has been used other part of code

drivers/ovsd/ovsSwitch.go
g1rana: ok;

drivers/ovsd/ovsSwitch.go
g1rana: see this section of code:
 sw.ovsdbDriver.ovsSwitch = sw

        if netType == "vxlan" {
                ofnetPort = vxlanOfnetPort
                ctrlrPort = vxlanCtrlerPort
                switch fwdMode {
                case "bridge":
                        datapath = "vxlan"
                case "routing":
                        datapath = "vrouter"
                default:

PR #1141, opened 2018-06-04
fix_the_etcdv2_1000ErrorCodeEventIndexCleared_bug

liucimin:
#### Fixes #
Please describe:
And the protection for the watcher.
When the watcher get the ErrorCodeEventIndexCleared from the etcd,we should create a new watcher
to catch Future events。

manual  test in my env.

## TODO
- [ ] Tests
1.Create a new network for the contiv.
   Result : make sure the pod can be create by the netplugin.

2.Use the contiv until the etcd index 1000 larger than step 1.
   Result :  use the curl to get the etcd's index.

3.Interrupt the network between netplugin and etcd.
   Result:  can see "Error client:etcd cluster is unavailable or misconfigured during watch" in the netplugin log.

4.Resume the etwork between netplugin and etcd.
    Result:  can see "Error 401: The event in requested index is outdated and cleared (*) during watch" in the netplugin log.

5.Create a new network for the contiv.
    Result:  create succeed.
6.Create new pods.
  Result : make sure the pod can be create by the new network.

- [ ] Documentation

### DIFF
--- a/state/etcdstatedriver.go
+++ b/state/etcdstatedriver.go
@@ -176,16 +176,26 @@ func (d *EtcdStateDriver) ReadAll(baseKey string) ([][]byte, error) {
 	return [][]byte{}, err
 }
 
-func (d *EtcdStateDriver) channelEtcdEvents(watcher client.Watcher, rsps chan [2][]byte) {
+func (d *EtcdStateDriver) channelEtcdEvents(watcher client.Watcher, rsps chan [2][]byte, baseKey string) {
 	for {
 		// block on change notifications
 		etcdRsp, err := watcher.Next(context.Background())
 		if err != nil {
-			log.Errorf("Error %v during watch", err)
-			time.Sleep(time.Second)
+			log.Infof("Watch %s next failed: %v %v", baseKey, reflect.TypeOf(err), err)
+			switch err.(type) {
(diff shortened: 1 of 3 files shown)


liucimin:
@dseevr

state/etcdstatedriver.go
```diff
@@ -176,16 +176,26 @@ func (d *EtcdStateDriver) ReadAll(baseKey string) ([][]byte, error) {
...
 		// block on change notifications
 		etcdRsp, err := watcher.Next(context.Background())
 		if err != nil {
-			log.Errorf("Error %v during watch", err)
-			time.Sleep(time.Second)
+			log.Infof("Watch %s next failed: %v %v", baseKey, reflect.TypeOf(err), err)
+			switch err.(type) {
+			case *client.ClusterError:
+				// retry and wait for etcd cluster to recover!
+				time.Sleep(time.Second * 5)
```
unclejack: How did you choose this value for the sleep?

liucimin: The value for the sleep can be set any value.But this decides how many times the watcher will send request to the etcd until the cluster being recover.
The etcd cluster may be recovered any time after it broken.So i choose the value from my experiences.

PR #1145, opened 2018-06-06
Modified that check burst value when run the function ValidateNetprofile

wangxf1987:
#### Type of fix: Bug Fix
Please describe:
- The burst is hard code, if want to achieve the desired performance, 
   it's best to equal the value of bandwidth.
- Add func ConvertBandwidth,that convert the bandwidth to bytes.
- Modified func ValidateNetprofile.
- Test pass on local env

## TODO
- [ ] Documentation

### DIFF
--- a/contivmodel/contivModel.go
+++ b/contivmodel/contivModel.go
@@ -9,7 +9,7 @@ import (
 	"errors"
 	log "github.com/Sirupsen/logrus"
 	"github.com/contiv/netplugin/objdb/modeldb"
-        "github.com/contiv/netplugin/utils/netutils"
+	"github.com/contiv/netplugin/utils/netutils"
 	"github.com/gorilla/mux"
 	"net/http"
 	"regexp"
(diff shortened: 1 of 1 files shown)


contivmodel/contivModel.go
```diff
@@ -3457,7 +3458,8 @@ func ValidateNetprofile(obj *Netprofile) error {
 		return errors.New("bandwidth string invalid format")
 	}
 
-	if obj.Burst > 10486 {
+        bwint64 := netutils.ConvertBandwidth(obj.Bandwidth)
```
dseevr: Where is this function?  It looks like it wasn't include in the PR, but I see it in the other PR

dseevr: Where is this function? It looks like it wasn't include in the PR, but I see it in the other PR

wangxf1987: The function from the netutils.ConvertBandwidth and we use it directly()。The function will be convert the Bandwidth value, and the value type is int64.

Issue #1054, opened 2017-11-07
contiv cni plugin is broken with cri-o runtime.


sbskas:
## Description
When running kubernetes with the cri-o runtime (v1.0.2), the contiv plugin is unable to add network.
When creating a new pod, the following errors are seen : 

`crio[1754]:` time="2017-11-07 23:19:48.420019761+01:00" level=error msg="Error adding network: Contiv:Error moving to netns; Err: invalid nw name space: `/var/run/netns/k8s_kube-dns-5c48f5cf98-lgth2_kube-system_1670dee9-c2ea-11e7-a60b-782bcb3fbaa1_0-bf1577dd"

`kubelet[2065]:` E1107 23:19:48.460313    2065 pod_workers.go:182] Error syncing pod 1670dee9-c2ea-11e7-a60b-782bcb3fbaa1 ("kube-dns-5c48f5cf98-lgth2_kube-system(1670dee9-c2ea-11e7-a60b-782bcb3fbaa1)"), skipping: failed to "CreatePodSandbox" for "kube-dns-5c48f5cf98-lgth2_kube-system(1670dee9-c2ea-11e7-a60b-782bcb3fbaa1)" with CreatePodSandboxError: "CreatePodSandbox for pod \"kube-dns-5c48f5cf98-lgth2_kube-system(1670dee9-c2ea-11e7-a60b-782bcb3fbaa1)\" failed: rpc error: code = Unknown desc = failed to create pod network sandbox k8s_kube-dns-5c48f5cf98-lgth2_kube-system_1670dee9-c2ea-11e7-a60b-782bcb3fbaa1_0(e57408547d77f78c0e99edd130b3d7a97754806976599a2553aa1bb4d3bfc466): Contiv:Error moving to netns; Err: invalid nw name space: /var/run/netns/k8s_kube-dns-5c48f5cf98-lgth2_kube-system_1670dee9-c2ea-11e7-a60b-782bcb3fbaa1_0-bf1577dd"

#### Expected Behavior
A new network should be created and then an interface and an address should be assigned to the deployed pod.

#### Observed Behavior
Pod fails to get an address an thus cannot communicate with the rest of the cluster.

## Steps to Reproduce (for bugs)
1- Instal cri-o v1.0.2 and remove docker
2- Create a k8s cluster with kubeadm (v1.8.2).
3- Install contiv 1.1.6: ./install/k8s/install.sh -n <kubemaster IP>
[...]

PR #1150, opened 2018-08-08
k8splugin support manage_network_ns_lifecycle=true from crio

n3wt0nSAN:
#### Type of fix: 
Bug fix
#### Fixes #
[1054](https://github.com/contiv/netplugin/issues/1054)
Please describe:
- changes made in the Pull request
The cni 0.3 support the network namespace path can be `/proc/[pid]/ns/net` or a `bind-mount/link(like /var/run/netns/cni-<something>)` to it.
So, when the flag `manage_network_ns_lifecycle=true` on the **cri-o runtime**, **contiv** return the bug #1054.
I made some logical modifications to contiv run over buth net path. Some problems of permission on container, I can't execute the operation on the mounted path and that force me send part of the code to k8s_cni code.
Modifications: 
> On the function `addPod()` of file `driver.go` of the k8splugin  package
> And put the same logical on the function `addPodToContiv()`  of file` k8s_cni.go` of cniapi  package
Manual tests on Kubernetes v1.10.5 with CRI-O v1.10.6 and Kata Container v1.1.0, and Kata needed the manage_network_ns_lifecycle=true, https://github.com/kata-containers/documentation/blob/master/architecture.md#cni. 

## TODO
- [x] Tests
- [ ] Documentation


mgmtfn/k8splugin/contivk8s/k8s_cni.go
```diff
@@ -95,15 +279,40 @@ func addPodToContiv(nc *clients.NWClient, pInfo *cniapi.CNIPodAttr) {
 		}
 		os.Exit(1)
 	}
+	log.Infof("EP created IP: %s\n", result.Attr.IPAddress)
```
liucimin: Why change this function?
Why not change the 
// addPod is the handler for pod additions
func addPod(w http.ResponseWriter, r *http.Request, vars map[string]string) (interface{}, error) 

in \netplugin\mgmtfn\k8splugin\driver.go

n3wt0nSAN: When, the runtime send the pid process the operation inside the container is permitted, but when is the path of network namespace I have problems to exec same operation I get the error:  `RTNETLINK answers: Invalid argument, exit status 2`

`ls -lha /var/run/netns/`
```
-rw-r--r--  1 root root    0 Aug  9 18:41 cni-d5e070e5-08c4-46f0-96ee-4bb5e6c6e21a
lrwxrwxrwx  1 root root   55 Aug  9 18:41 k8s_contiv-blue-c1_default_01bc671d-9a69-11e8-a90e-000af70485d0_1-81a93eff -> /var/run/netns/cni-d5e070e5-08c4-46f0-96ee-4bb5e6c6e21a
```

So, the k8s_cni is part of the contivk8s plugin execute on the host. 

I make some test to tray maintain the logical on `./netplugin\mgmtfn\k8splugin\driver.go` but no success.

utils/ns/ns_linux.go
```diff
@@ -0,0 +1,305 @@
...
+// WithNetNSPath executes the passed closure under the given network
+// namespace, restoring the original namespace afterwards.
+func WithNetNSPath(nspath string, toRun func(NetNS) error) error {
+	ns, err := GetNS(nspath)
+	if err != nil {
+		return err
+	}
+	defer ns.Close()
+	return ns.Do(toRun)
+}
```
liucimin: maybe these files can use vendor in the project?
https://github.com/containernetworking/plugins/tree/master/pkg/ns

n3wt0nSAN:
Yes. Is good practice. But the master version of this vendor don't have the
function I need. Only
https://github.com/containernetworking/plugins/tree/v0.7/pkg/ns
<https://github.com/containernetworking/plugins/tree/master/pkg/ns>.

You now how I get this specific branch with *go get*?

TIA

